PR #285, opened 2023-06-29
284 auditor code review general comments

### LINKED ISSUE #284: Auditor Code Review - General Comments
For the main Proposal code. This I think merits some discussion... Do we want the FundingManager to be a ERC20 itself? If it's the  way we are going to go we should enforce it on Interface level. I'd be generally against it, though :
Proposal-1
For the PaymentClient. We chose to remove this. Should we reconsider?
PaymentClient-1
For the Module. Also mentioned in #281:
Module-1
For the MilestoneManager
MilestoneManager-2
MilestoneManager-1

FHieser:
Solves #284


0xNuggan:
Everything looks good. 
The only thing for me is the complete removal of receiptToken(). I thought we were going to move it to IFundingManager? I think there is value in having a standard defined way for other modules to know which token the proposal is "giving out" on deposit. But we can make an issue and eventually do a separate PR for that

0xNuggan (approved):

FHieser:
> Everything looks good. The only thing for me is the complete removal of receiptToken(). I thought we were going to move it to IFundingManager? I think there is value in having a standard defined way for other modules to know which token the proposal is "giving out" on deposit. But we can make an issue and eventually do a separate PR for that

Yeah we can do a separate issue for it if necessary.
But I thought about it again and I still dont like it.
The receiptToken function for the current FundingManager would just return its own address, because it is its own receiptToken.
How would that look like in other fundingManagers? In what functionality combination would this function point to something else beside itself.
The only one I could come up with that was reasonable was when it didnt have a receiptToken Function at all and then the function would return a address(0) and that doesnt make sense for me

Issue #283, opened 2023-06-26
Auditor Code Review - Streaming Payment Processor


0xNuggan:
Tagging @saxenism 
![StreamingPM1](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/9631e16a-af2e-4d11-a81e-ee4d461e1807)
![StreamingPM-2](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/0da5bb5c-5436-4914-8071-4afa3070f7f4)
![StreamingPM-3](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/8de5553c-2213-4ad2-a674-d00831722a48)
![StreamingPM-4](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/d2b14d8f-e434-4fc2-ae53-2f65bd4285e1)
![StreamingPM-5](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/84ca928d-e54b-4c0d-a5cd-c47253b70256)
![StreamingPM-12](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/549b55bf-19b6-4e22-8026-4d5580cd3914)
![StreamingPM-6](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/b15b168e-6cfb-48a5-b71a-4450b93c54e5)
![StreamingPM-7](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/eeb7c943-00ee-4e5a-8d3f-abb5a4ec219e)
![StreamingPM-8](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/9eb15383-6fce-44bd-8808-0c84f52778dd)
![StreamingPM-9](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/7a53b897-0153-42ba-a180-a8194b512744)
![StreamingPM-10](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/08af7e96-8ee7-444d-aa27-6e5044a9fbf2)
![StreamingPM-11](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/3ad4ae56-2f29-4cf5-9ba6-ed440fa3f0db)
![StreamingPM-12](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/eaceb464-eac1-4c64-a0a8-e462991bde2a)

saxenism:
Acknowledged the tag, @0xNuggan .

[...]

PR #286, opened 2023-06-30
283 streaming payment processor audit review

saxenism:
This PR fixes the (relevant) issues raised in #283


FHieser:
I wanna be a real git wizard
I broke the original branch during the rebasing process for some reason so I created a new PR with the correct changes
See #287

PR #287, opened 2023-07-04
283 auditor code review streaming payment processor

FHieser:
This PR fixes the (relevant) issues raised in https://github.com/InverterNetwork/inverter-contracts/issues/283

Sorry Rahul,
I broke parts of the old branch and therefor I created a new branch and PR to solve that
Solves #283


src/modules/paymentProcessor/StreamingPaymentProcessor.sol
```diff
@@ -209,10 +200,10 @@ contract StreamingPaymentProcessor is Module, IStreamingPaymentProcessor {
     }
 
     /// @inheritdoc IStreamingPaymentProcessor
-    function removePayment(IPaymentClient client, address contributor)
-        external
-        onlyAuthorized
-    {
+    function removeAllPaymentForContributor(
```
SchneiderNejc: typo: use removeAllPaymentsForContributor or even better, removeAllContributorPayments instead.

SchneiderNejc (changes requested):
Checked both the auditor recommendations as well as code changes made by @saxenism. Added my opinion on each one, check the table bellow.

| Screenshot id* | Line | Should apply** | Fix applied | Description |
| ------ | ----------- | ----------- | ----------- | ----------- |
| 1 | 155 | NO | YES | Make sure funds are available at the time of payment creation to ensure they can be paid out. |
| 2 | 407 | YES | YES | Loops iterations amount has been decreased since we now only check the active contributor wallets and not all the wallets. |
| 3 | 212 | YES | NEEDS FIX | Applied but there is a typo in the function name. |
| 4 | 244, 460, 672 | YES | YES | Code has been cleaned, duplications are removed by using a single function for "complex" operations. |
| 5 | 42 | YES | PARTIALLY | `ContributorWallets` has been renamed to `StreamingWallets`, `activePayments` has been renamed to `activeContributors`, `vestings` has not been renamed. I recommend renaming everything to either to vesting (or streaming). So instead of `StreamingWallets` -> `VestingWallets`; instead of `activePayments` -> `activeVestings`. The less keywords for the same thing, the better. |
| 6 | 24 | YES | NO | `uint _streamingWalletID` is used nowhere else in the project and can be removed |
| 7 | 54 | ? | NO | Documentation and data structure has been updated, but issue has not been addressed. |
| 8 | 33 | YES | YES | `isActiveContributor` has been removed since activity is already tracked in `activePayments` (renamed to `activeContributors`) mapping. |
| 9 | 701 | YES | YES | `_vestingScheduleForSpecificWalletId` has been renamed to `_vestingAmountForSpecificWalletId`. |
| 10 | 699 | YES | PARTIALLY | Only the documentation has been updated, but I recommend removing `uint timestamp` parameter altogether and using `block.timestamp` instead. |
| 11 | 662 | NO | NO | We use low-level call for upgradable contracts in this case for transfering tokens from client to contributor. `SafeTransferFrom` will not work in this case. |
| 12 | 567 | NO | NO | This cant be enforced because of the recurring payment Manager that needs to be able to create payments that can immediatly be paid out because they happend in the past. |
| 13 | 24 | NA | NA | DUPLICATION, check number #6 |

*Screenshot id's refer to the pictures in first post.
**Should apply is my personal opinion weather fix should be applied, read more about it in description

TL;DR Id's 1, 3, 5, 6, 10 should be updated, check them out.

FHieser:
@SchneiderNejc
update validStart function to also check that _dueTo is greater than (or equal to in case we allow immidiate payouts) _start.
-> This cant be enforced because of the recurring payment Manager that needs to be able to create payments that can immediatly be paid out because they happend in the past

SchneiderNejc:
@FHieser makes sense, than we can leave that out. Updated my original post.

saxenism:
> 

I 10: If I change the timestamp to `block.timestamp`, that would mean we could only find the `vestingAmountForSpecificWalletId` for the current timestamp and not an arbitrary timestamp. Is that something we want to go ahead with?

I 6: Actually the streamingWalletId is being used and used quite extensively throughout the code. If you take a look at all the places where the `vestings` mapping is being used, there the streamingWalletId plays quite a considerable role in the identification of the specific wallet to perform the operation on. I am not very sure if we should remove it. Please reconsider.

I 5: Renamed streamingWallets to vestingWallets  and `activePayments` to `activeVestings`

I 3: Changed to `removeAllContributorPayments`.

I 1: Reverted the changes and re-introduced the balance check.

src/modules/paymentProcessor/StreamingPaymentProcessor.sol
saxenism: Fixed. Went ahead and renamed.

SchneiderNejc:
1, 3, 5: Looks good.
6: Fine, lets keep that than.
10. I guess that's up for debate, but I don't see a reason to lookup vesting at arbitrary timestamp?

SchneiderNejc (approved):

saxenism:
> 1, 3, 5: Looks good. 6: Fine, lets keep that than. 10. I guess that's up for debate, but I don't see a reason to lookup vesting at arbitrary timestamp?

@FHieser @marvinkruse @0xNuggan any comments regarding the timestamp thingy? I personally think it would be good if the contributors can actually take a sneak peek into what exactly they'll earn some time out in the future, right?

FHieser:
> > 1, 3, 5: Looks good. 6: Fine, lets keep that than. 10. I guess that's up for debate, but I don't see a reason to lookup vesting at arbitrary timestamp?
> 
> @FHieser @marvinkruse @0xNuggan any comments regarding the timestamp thingy? I personally think it would be good if the contributors can actually take a sneak peek into what exactly they'll earn some time out in the future, right?

Yeah I also think it was only a documentation problem here.
You could even go so far to remove the internal function entirely and just put its functioanlity into the vestedAmountForSpecificWalletId function altogether, because its not used elsewhere and the natspec is essentially the same, but that is more of a preference thing I guess

Issue #261, opened 2023-05-30
Move and Refactor Payment Client


FHieser:
- Rename PaymentClient to ERC20PaymentCient (At some point we might create another one that handles ERC721 tokens)
- let Payment Client inherit Module and make functions _ensureTokenBalance() _ensureTokenAllowance and _isAuthorizedPaymentProcessor not virtual anymore
- move file into LogicModule Folder (I think thats the place to put it. Any opinions?)

Issue #208, opened 2023-04-05
Check whether the contracts work with "weird" ERC20 tokens 


0xNuggan:
For example, USDC has only 6 decimals, we should check that they don't break anything.

For reference: https://github.com/d-xo/weird-erc20

FHieser:
Will tackle at a later time

PR #293, opened 2023-07-18
277 bountymanager logicmodule

### LINKED ISSUE #277: BountyManager LogicModule
Bounty module that receives bounty completion requests, manages submission/verification and submits payout upon verification. 

Bounty submission is:
NFT gated
Subject to verification by 3rd address (neither worker nor owner).
Payout happens on verification.

Note: The Bounty doesn’t get set up beforehand. The workers independently submit work done and request payment, and verifiers decide if it falls within scope/ requested payout is adequate.

Bounties set a minimum and maximum payout on each bounty
Verification can happen by 3rd address (neither worker nor owner)
Can be NFT/token gated to submit bounties
Payout happens after verification.

FHieser:
solves #277


src/modules/logicModule/IBountyManager.sol
```diff
@@ -0,0 +1,266 @@
...
+        uint maximumPayoutAmount;
+        /// @dev Arbitrary data to store Bounty details if necessary.
+        ///      CAN be empty.
+        bytes details;
+        /// @dev Id that claimed the bounty
+        ///      A Bounty is claimed if a Claim got acknowledged by a Verifier
+        uint claimedBy;
+    }
+
+    struct Claim {
```
saxenism, line 39: Perhaps, it will be a good idea to keep the `Contributor` struct above struct `Claim`. It simply reads better.

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+        }
+        _;
+    }
+
+    modifier onlyClaimContributor(uint claimId) {
+        address sender = _msgSender();
+        Contributor[] memory contribs = _claimRegistry[claimId].contributors;
+        uint length = contribs.length;
+        bool found;
+        for (uint i; i < length; i++) {
```
saxenism: unchecked pre-increments would be faster.

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+            if (contribs[i].addr == sender) {
```
saxenism, line 60: If we could ensure that the list of contributors is sorted... then we can use binary search instead of linear search, which will reduce the current O(n) complexity to O(log N).

This seems doable Solidity language wise, but let me know if this feasible based on the context of the module or not.

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+            revert Module__BountyManager__OnlyRole(roleId, address(this));
+        }
+        _;
+    }
+
+    modifier onlyClaimContributor(uint claimId) {
+        address sender = _msgSender();
+        Contributor[] memory contribs = _claimRegistry[claimId].contributors;
+        uint length = contribs.length;
+        bool found;
```
saxenism: You can save the cost of this extra boolean `found` by comparing the for loop iterator with the length of the array after it gets out of the loop. However in that case, do-while might be a more concrete choice.

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+        //@update to be in correct range
+        uint length = contributors.length;
+        //length cant be zero
+        if (length == 0) {
+            revert Module__BountyManager__InvalidContributorsLength();
+        }
+        uint totalAmount;
+        uint currentAmount;
+        address contrib;
+        for (uint i; i < length; i++) {
```
saxenism: unchecked pre-increment interations would be faster

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+            currentAmount = contributors[i].claimAmount;
+
+            //amount cant be zero
+            if (currentAmount == 0) {
```
saxenism, line 122: I don't really think having a zero payout for one of the contributors should be considered as an erroring state. I think it's perfectly reasonable to assume that 1 out of 100 contributor could not commit some time to the bounty and therefore their payout turned out to be 0...

Let me know if I'm making some wrong assumptions.

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+                revert Module__BountyManager__InvalidContributorAmount();
+            }
+
+            contrib = contributors[i].addr;
+            if (
+                contrib == address(0) || contrib == address(this)
+                    || contrib == address(proposal())
```
saxenism: repeated function calls made to the `proposal()` function inside the loop. Cache it, perhaps?

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+    /// @inheritdoc IBountyManager
+    function addBounty(
+        uint minimumPayoutAmount,
+        uint maximumPayoutAmount,
+        bytes calldata details
+    )
+        external
+        onlyRole(uint8(Roles.BountyAdmin))
+        validPayoutAmounts(minimumPayoutAmount, maximumPayoutAmount)
+        returns (uint id)
```
saxenism, line 279: Follow a single convention for return data from functions.

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+        if (_claimRegistry[claimId].bountyId != bountyId) {
+            revert Module__BountyManager__ClaimNotBelongingToBounty();
+        }
+        _;
+    }
+
+    modifier notClaimed(uint bountyId) {
+        //Its not claimed if claimedBy is still 0
+        if (_bountyRegistry[bountyId].claimedBy != 0) {
+            revert Module__BountyManager__BountyAlreadyClaimed();
```
saxenism: This revert message will be thrown up even when the bounty is locked and the `claimedBy` value is type(unit).max, which will not be an entirely accurate error message.

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+        bytes calldata details
+    )
+        external
+        onlyRole(uint8(Roles.ClaimAdmin))
+        validBountyId(bountyId)
+        notClaimed(bountyId)
+        returns (uint id)
+    {
+        validContributorsForBounty(contributors, _bountyRegistry[bountyId]);
+        // Note ids start at 1.
```
saxenism: Not an accurate comment. Since `claimId` and `bountyId` both share the same variable: `_nextId`, all you can say is the IDs are non-zero, because it is impossible for both the IDs to start with 1.

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+        // Add Claim id to the list.
+        _claimList.addId(claimId);
+
+        Claim storage c = _claimRegistry[claimId];
+
+        // Add Claim instance to registry.
+        c.bountyId = bountyId;
+
+        uint length = contributors.length;
+        for (uint i; i < length; ++i) {
```
saxenism: unchecked pre-increments

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+    function addClaim(
+        uint bountyId,
+        Contributor[] calldata contributors,
+        bytes calldata details
+    )
+        external
+        onlyRole(uint8(Roles.ClaimAdmin))
+        validBountyId(bountyId)
+        notClaimed(bountyId)
+        returns (uint id)
```
saxenism, line 333: again, named return convention not followed accurately, uniformly.

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+        external
+        validClaimId(claimId)
+        onlyClaimContributor(claimId)
+        validBountyId(bountyId)
+    {
+        validContributorsForBounty(contributors, _bountyRegistry[bountyId]);
+        Claim storage c = _claimRegistry[claimId];
+
+        uint length = c.contributors.length;
+        for (uint i; i < length; ++i) {
```
saxenism: for loop optimization.

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+            //remove ClaimId for each contributor address
+            contributorAddressToClaimIds[c.contributors[i].addr].remove(claimId); //@note c.contributors[i].addr -> is there a more gas efficient alternative to this?
+        }
+
+        delete c.contributors;
+
+        length = contributors.length;
+
+        for (uint i; i < length; ++i) {
```
saxenism: for loop optimization.

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+        uint length = contribs.length;
+
+        //total amount needed to verifyBounty
+        uint totalAmount;
+
+        //current contributor in loop
+        Contributor memory contrib;
+
+        //For each Contributor add payments according to the claimAmount specified
+        for (uint i; i < length; i++) {
```
saxenism: for loop optimization.

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+        if (
+            totalAmount > bounty.maximumPayoutAmount
+                || totalAmount < bounty.minimumPayoutAmount
+        ) {
+            revert Module__BountyManager__ClaimExceedsGivenPayoutAmounts();
+        }
+    }
+
+    modifier claimBelongingToBounty(uint claimId, uint bountyId) {
+        //Its not claimed if claimedBy is still 0
```
saxenism: Stale comment?

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+        _bountyRegistry[bountyId].claimedBy = claimId;
+
+        emit ClaimVerified(claimId, bountyId);
+    }
+
+    //----------------------------------
+    // Role Functions
+
+    /// @inheritdoc IBountyManager
+    function grantBountyAdminRole(address addr) external onlyAuthorized {
```
saxenism, line 469: Should we put zero address checks here?

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,528 @@
...
+
+    function _ensureTokenAllowance(IPaymentProcessor spender, uint amount)
+        internal
+        override(PaymentClient)
+    {
+        IERC20 token = __Module_proposal.token();
+        uint allowance = token.allowance(address(this), address(spender));
+
+        if (allowance < amount) {
+            token.safeIncreaseAllowance(address(spender), amount - allowance);
```
saxenism, line 547: we gotta fork-test with multiple market leading ERC20 tokens to make sure that `safeIncreaseAllowance` works with them. As originally only `approve` was part of the ERC20 standard and not `safeIncreaseAllowance`.

src/modules/logicModule/BountyManager.sol
FHieser: You mean like this?
for (uint i; i < length; ) {
//loop content
unchecked{++i;}
}

src/modules/logicModule/BountyManager.sol
FHieser: If you know an easy implementation that is actually cheaper gas wise than sure we can do that
If you dont I think its probably not a high priority target for optimization, because the functions that use it probably wont be in use that often

src/modules/logicModule/BountyManager.sol
FHieser: Why would you include a contributor qho isnt get paid anything?
I think this is more of a design decision, there are points for both sides
If at some point there is a cry for a change in production we can still change it, but until then i would keep it

src/modules/logicModule/BountyManager.sol
FHieser: What convention are you referring to?

src/modules/logicModule/BountyManager.sol
FHieser: Can you route a call via address(0) ?
If not there is no harm in adding the zero address to the role i think

src/modules/logicModule/BountyManager.sol
FHieser: This should be tackled in #261 and #208

src/modules/logicModule/BountyManager.sol
```diff
@@ -0,0 +1,548 @@
...
+        onlyRole(uint8(Roles.BountyAdmin))
+        validBountyId(bountyId)
+    {
+        _bountyRegistry[bountyId].details = details;
+
+        emit BountyUpdated(bountyId, details);
+    }
+
+    /// @inheritdoc IBountyManager
+    function lockBounty(uint bountyId)
```
saxenism, line 312: Well, there is a `lockBounty` function, and I do not specifically know why it is here, but would it make sense to include another access controlled `unlockBounty` function?

FHieser: I would say no because if you wanted to change the Bounty in the first place you could have just used updateBounty
Even if lockBounty is final, you could just create a new Bounty
The initial intend for the function was to be able to actually clean up remaining bounties that werent reasonable to be fulfilled in the first place

0xNuggan (approved):

Issue #258, opened 2023-05-22
Update contract naming(s)


marvinkruse:
In the spec and everywhere else, we use new terms, i.e., Orchestrator instead of Proposal, etc. We should probably adapt this in the contracts as well.

FHieser:
Rename configdata to configData and dependencydata to dependencyData

PR #294, opened 2023-07-18
Late dependency injection

saxenism:
Solves the late dependency injection issue mentioned [here](https://github.com/InverterNetwork/inverter-contracts/issues/207).

The approach taken here is to pass the dependency for each module along with it's `configdata`.

The above approach is idle for 0,1 or 2 dependencies. For more than that, it could become a little cumbersome.

As for the tests, all the tests have been updated to be compatible with the late dependency injection. The `ProposalFactory` tests were unclear to me, so I have commented them out for now. Could include them after discussions with the team about the intention of those tests.

EDIT: The issues surrounding the `ProposalFactory` tests have been cleared and all those tests are running fine now.


src/modules/MetadataManager.sol
```diff
@@ -42,6 +50,40 @@ contract MetadataManager is IMetadataManager, Module {
         _setTeamMetadata(teamMetadata_);
     }
 
+    function init2(IProposal proposal_, bytes memory configdata)
+        external
+        initializer2
+    {
+        __Module_initialization = true;
```
FHieser: Couldnt you just do it like this?
modifier initializer2() {
        if (__Module_initialization == true) {
            revert Module__CannotCallInit2Again();
        }
__Module_initialization = true;
        _;
    }

src/modules/MetadataManager.sol
```diff
@@ -42,6 +50,40 @@ contract MetadataManager is IMetadataManager, Module {
         _setTeamMetadata(teamMetadata_);
     }
 
+    function init2(IProposal proposal_, bytes memory configdata)
```
FHieser: In general I would have everyone inherit an empty init2 function via the module
If someone needs it they can just overwrite it
That way you dont have to explicitly include it in every module implementation

src/proposal/Proposal.sol
```diff
@@ -119,6 +136,202 @@ contract Proposal is IProposal, OwnableUpgradeable, ModuleManager {
...
+        }
+
+        return moduleAddress;
+    }
+
+    //--------------------------------------------------------------------------
+    // Module address verification functions
+    // Note These set of functions are not mandatory for the functioning of the protocol, however they
+    //      are provided for the convenience of the users since matching the names of the modules does not
+    //      fully guarentee that the returned address is the address of the exact module the user was looking for
```
FHieser: *guarantee

src/modules/MetadataManager.sol
saxenism: That makes sense, can try that out.

src/proposal/Proposal.sol
```diff
@@ -119,6 +136,202 @@ contract Proposal is IProposal, OwnableUpgradeable, ModuleManager {
...
+
+    /// @inheritdoc IProposal
+    function verifyAddressIsRoleAuthorizerModule(address roleAuthAddress)
```
FHieser: Instead to test for all contracts that we developed I would maybe go for the basic Interfaces:
So IAuthorizer
PaymentProcessor, Fucningmanager and logicmodule
otherwise the contract size will at some point be too big to support this

src/modules/MetadataManager.sol
saxenism: Agreed, initially I wanted to change the existing structures as little as possible, but this seems like a good suggestion.

Will try this.

src/proposal/Proposal.sol
```diff
@@ -8,7 +8,23 @@ import {OwnableUpgradeable} from "@oz-up/access/OwnableUpgradeable.sol";
 import {IERC20} from "@oz/token/ERC20/IERC20.sol";
 
 // Internal Dependencies
+import {RoleAuthorizer} from "src/modules/authorizer/RoleAuthorizer.sol";
```
FHieser: See comment below:
imports could then be removed

src/proposal/base/ModuleManager.sol
```diff
@@ -64,7 +64,7 @@ abstract contract ModuleManager is
     }
 
     modifier moduleLimitNotExceeded() {
-        if (uint8(_modules.length) >= MAX_MODULE_AMOUNT) {
+        if (_modules.length >= uint(MAX_MODULE_AMOUNT)) {
```
FHieser: Why dont you just make MAX_MODULE_AMOUNT uint instead of uint8 ?

test/e2e/CreateNewProposal.t.sol
```diff
@@ -241,23 +241,36 @@ contract ProposalCreation is Test {
...
+        address initialManager = makeAddr("initial manager address");
 
         IProposalFactory.ModuleConfig memory authorizerFactoryConfig =
         IProposalFactory.ModuleConfig(
-            authorizerMetadata, abi.encode(initialAuthorizedAddresses)
+            authorizerMetadata,
+            abi.encode(
+                initialAuthorizedAddresses,
+                initialManager,
+                hasDependency,
```
FHieser: is hasDependency ever used?

test/proposal/base/ModuleManager.t.sol
```diff
@@ -51,6 +51,10 @@ contract ModuleManagerTest is Test {
     //--------------------------------------------------------------------------
     // Tests: Initialization
 
+    /////////////////////////////////////////////////////////////////////////////////////////////
+    // Test passes apart from the cases where number of modules exceed ModuleLimit
+    /////////////////////////////////////////////////////////////////////////////////////////////
+    /*
```
FHieser: Outcommented

src/proposal/IProposal.sol
```diff
@@ -21,6 +21,12 @@ interface IProposal is IModuleManager {
     /// @notice Execution of transaction failed.
     error Proposal__ExecuteTxFailed();
 
+    /// @notice Passed module name is invalid
+    error DependencyInjection__InvalidModuleName();
```
FHieser: Are these errors used somewhere?

src/factories/ProposalFactory.sol
```diff
@@ -125,6 +129,24 @@ contract ProposalFactory is IProposalFactory {
             IPaymentProcessor(paymentProcessor)
         );
 
+        // Second round of module initializations to satisfy cross-referencing between modules
+        // This can be run post the proposal initialization. This ensures a few more variables are
+        // available that are set during the proposal init function.
+        for (uint i; i < modulesLen; ++i) {
+            IModule(modules[i]).init2(
+                IProposal(clone), moduleConfigs[i].configdata
```
FHieser: Instead of using the configdata to save the dependency, couldnt you create a new property in the ModuleConfig struct?
 struct ModuleConfig {
        IModule.Metadata metadata;
        bytes configdata;
        bytes dependecydata;
    }
If it is empty you could skip the init2 step
and you wouldnt need to decode the whole configdata again in the init2 step

test/factories/ProposalFactory.t.sol
```diff
@@ -74,82 +74,83 @@ contract ProposalFactoryTest is Test {
...
-    function testProposalMapping(uint proposalAmount) public {
-        // Note to stay reasonable
-        proposalAmount = bound(proposalAmount, 0, 50);
-
-        for (uint i = 1; i < proposalAmount; ++i) {
-            address proposal = _deployProposal();
-            assertEq(proposal, factory.getProposalByID(i));
-        }
-    }
+    // function testValidProposalId(uint getId, uint proposalsCreated) public {
```
FHieser: ModuleFactoryMock actually doesnt create any real modules so it cant call init2 functions in there, thats why it fails

FHieser (changes requested):

src/modules/MetadataManager.sol
saxenism: So, yeah, it works.

Tried that out here: https://pastebin.com/0DLKt9TL

Also, now, added it to the codebase.

src/proposal/IProposal.sol
```diff
@@ -107,25 +104,9 @@ interface IProposal is IModuleManager {
         address rebasingFundingManagerAddress
     ) external view returns (bool);
 
-    /// @notice Verify whether the given address is the payment client
-    function verifyAddressIsPaymentClient(address paymentClientAddress)
+    /// @notice Verify whether the given address is an IAuthorizer module
```
FHieser, line 132: What about the other major modules?
You changed the Authorizer Module, but not Payment Processor or fundingmanager

src/factories/ProposalFactory.sol
```diff
@@ -163,4 +174,13 @@ contract ProposalFactory is IProposalFactory {
...
         return _proposalIdCounter;
     }
+
+    function _dependencyInjectionRequired(bytes memory dependencydata)
+        internal
+        pure
+        returns (bool)
+    {
+        (bool hasDependency,) = abi.decode(dependencydata, (bool, string[]));
+        return hasDependency;
```
FHieser: Cant you just look if dependencyData is not empty?
if you set dependencyData initially to bytes("") instead of actually doing the abi.encode(emptyStringarray) this should be possible no?

saxenism: if I only check `dependencyData` is not empty then something like `abi.encode(uint256)` would also pass. This ensures that the bytes data at least has a boolean as the first param.

saxenism: this was agreed upon in an internal call. marking this as resolved.

src/proposal/Proposal.sol
```diff
@@ -207,13 +205,15 @@ contract Proposal is IProposal, OwnableUpgradeable, ModuleManager {
     }
 
     /// @inheritdoc IProposal
-    function verifyAddressIsRebasingFundingManager(
-        address rebasingFundingManagerAddress
-    ) external view returns (bool) {
-        RebasingFundingManager rebasingFundingManager =
-            RebasingFundingManager(rebasingFundingManagerAddress);
+    function verifyAddressIsFundingManager(address fundingManagerAddress)
```
FHieser, line 203: And the PaymentProcessor?

saxenism: yeah, missed it 😅 

Have included it now.

FHieser (approved):
Looks good

src/modules/base/Module.sol
```diff
@@ -87,6 +92,15 @@ abstract contract Module is IModule, Initializable, ContextUpgradeable {
         _;
     }
 
+    /// @dev same function as OZ initializer, but for the init2 function
+    modifier initializer2() {
```
FHieser, line 96: This is not tested right?

src/modules/base/Module.sol
```diff
@@ -123,6 +137,12 @@ abstract contract Module is IModule, Initializable, ContextUpgradeable {
         __Module_metadata = metadata;
     }
 
+    function init2(IProposal proposal_, bytes memory dependencydata)
+        external
```
FHieser, line 148: Same here

test/modules/logicModule/RecurringPaymentManager.t.sol
```diff
@@ -318,7 +337,11 @@ contract RecurringPaymentManagerTest is ModuleTest {
 
     function testRemoveRecurringPaymentModifierInPosition() public {
         //Init Module
-        recurringPaymentManager.init(_proposal, _METADATA, abi.encode(1 weeks));
+        recurringPaymentManager.init(
+            _proposal,
+            _METADATA,
+            abi.encode(1 weeks, hasDependency, dependencies)
```
FHieser: These encodings still contain hasDependency although that should have been moved over to the proposalfactory
I think this happened on multiple places in the code

FHieser (changes requested):

test/modules/logicModule/RecurringPaymentManager.t.sol
saxenism: Yes, true. Have fixed this.

src/modules/base/Module.sol
saxenism: Tested now

src/modules/base/Module.sol
saxenism: tested now

FHieser (approved):

src/factories/IModuleFactory.sol
```diff
@@ -42,7 +42,7 @@ interface IModuleFactory {
     /// @notice Creates a module instance identified by given metadata.
     /// @param metadata The module's metadata.
     /// @param proposal The proposal's instance of the module.
-    /// @param configdata The configdata of the module.
+    /// @param configdata The configdata of the module
```
marvinkruse, line 45: Was your `fmt` going wild, or was this intentional?

src/factories/IProposalFactory.sol
```diff
@@ -28,6 +31,7 @@ interface IProposalFactory {
     struct ModuleConfig {
         IModule.Metadata metadata;
         bytes configdata;
+        bytes dependencydata;
```
marvinkruse, line 34: Why are we not using camelCase here? I see that you just copied the style that already existed, so just as a general q.

src/modules/base/Module.sol
```diff
@@ -87,6 +92,22 @@ abstract contract Module is IModule, Initializable, ContextUpgradeable {
         _;
     }
 
+    /// @dev same function as OZ initializer, but for the init2 function
+    modifier initializer2() {
+        if (__Module_initialization == true) {
```
marvinkruse: This looks like it could be simplified slightly - or is this on purpose? Haven't found any other instance of this in our codebase though :D

src/proposal/base/ModuleManager.sol
```diff
@@ -64,7 +64,7 @@ abstract contract ModuleManager is
     }
 
     modifier moduleLimitNotExceeded() {
-        if (uint8(_modules.length) >= MAX_MODULE_AMOUNT) {
+        if (_modules.length >= MAX_MODULE_AMOUNT) {
```
marvinkruse, line 67: Not sure whether I need my next coffee or not, but is it really "exceeding" the limit if `length == max_amount`? Shouldn't this be a `>` instead of `>=`? Note: I might be dumb, so feel free to ignore :D

marvinkruse (commented):
Good stuff! Just added a few annoying comments to compensate for not finding anything important :D The logic and implementation itself look really good! Good stuff! 🙏

src/factories/IProposalFactory.sol
FHieser: I would agree to change that.
Lets move it to #258 
Otherwise this PR gets a bit bloated

marvinkruse: I agree, we can fix things like this in #258!

src/proposal/base/ModuleManager.sol
FHieser: I would agree on that front
Should be > instead of >=

FHieser: I redact my approval xD
Because moduleLimitNotExceeded is a modifier that runs through before the new Module is added it has to include the max_module_amount to work properly

marvinkruse: Hm, okay, that makes slight sense to me, but still the math doesn't fully make sense then. Why isn't this then just changed from 

```Solidity
modifier moduleLimitNotExceeded() {
        if (_modules.length >= MAX_MODULE_AMOUNT) {
            revert Proposal__ModuleManager__ModuleAmountOverLimits();
        }
        _;
    }
```

to this

```Solidity
modifier moduleLimitNotExceeded() {
        _;
        if (_modules.length > MAX_MODULE_AMOUNT) {
            revert Proposal__ModuleManager__ModuleAmountOverLimits();
        }
    }
```

So the code runs first (`_;`) and then is being checked afterwards on whether it exceeded the actual limit?

saxenism: I know I'm a bit late to the party, but I agree with @marvinkruse  here. Should I open a new PR to tackle this or was it internally discussed to be correct?

@FHieser

PR #295, opened 2023-07-21
Proposal Maintenance Funcs + Fixes

### LINKED ISSUE #263: add setAuthorizer and setPaymentProcessor functions (and probably setFundingManager)
Right now once the address is set there's no way to change it - even if we remove the module itself from the proposal. This is clearly bad. There should be an onlyAuthorized function to modify it

### LINKED ISSUE #278: Fix pre-commit hook
Currently, the pre-commit hook only works of certain environment variables are loaded since it runs the scripts as well - which is not ideal for a pre-commit hook. We will need to adapt the makefile to include the required env variables (just like with the `make testFuzz` command) so that it works without any prior work.

Currently the output is:
```
Error: 
Failed to deploy script:
Execution reverted: EvmError: Revert (gas: 1500082)
make[1]: *** [testScripts] Error 1
make: *** [pre-commit] Error 2
```

marvinkruse:
This pr introduces:
* Fixes #278 (allowing the pre-commit hook to be ran without sourcing any `.env` file)
* Fixes #263 (adding maintenance funcs to the proposal + tests)

The failing test on mainnet should also be done now.


FHieser (approved):

src/proposal/Proposal.sol
```diff
@@ -136,6 +136,33 @@ contract Proposal is IProposal, OwnableUpgradeable, ModuleManager {
     //--------------------------------------------------------------------------
     // onlyAuthorized Functions
 
+    /// @inheritdoc IProposal
+    function setAuthorizer(IAuthorizer authorizer_) external onlyAuthorized {
```
saxenism, line 140: Perhaps it'll be cool to emit events when such major changes are taking place. Might come in handy for the FE team?

test/proposal/Proposal.t.sol
```diff
@@ -231,4 +359,23 @@ contract ProposalTest is Test {
...
+    {
+        uint length = addrs.length;
+        vm.assume(length > 0); //Array has to be at least 1
+
+        if (length <= size) {
+            return addrs;
+        }
+
+        address[] memory cutArry = new address[](size);
+        for (uint i = 0; i < size - 1; i++) {
```
saxenism: unchecked pre-increments can make the tests run faster.

Also no need to initialize the `uint i` variable.

src/proposal/Proposal.sol
```diff
@@ -136,6 +136,33 @@ contract Proposal is IProposal, OwnableUpgradeable, ModuleManager {
...
+    }
+
+    /// @inheritdoc IProposal
+    function setFundingManager(IFundingManager fundingManager_)
+        external
+        onlyAuthorized
+    {
+        addModule(address(fundingManager_));
+        removeModule(address(fundingManager));
+        fundingManager = fundingManager_;
```
saxenism, line 154: Same. Can we emit an event about the change?

src/proposal/Proposal.sol
```diff
@@ -136,6 +136,33 @@ contract Proposal is IProposal, OwnableUpgradeable, ModuleManager {
...
+    }
+
+    /// @inheritdoc IProposal
+    function setPaymentProcessor(IPaymentProcessor paymentProcessor_)
```
saxenism, line 159: Same. Can we emit an event about the change?

test/proposal/Proposal.t.sol
marvinkruse: It's in a test, where we have quite some loops that are not optimized, but good point still - @FHieser fixed it 💪

PR #306, opened 2023-07-27
Create ProposalFactoryDemo

### LINKED ISSUE #304: Expand the proposal factory with create Demo functions
- Milestonemanager
- RecurringPaymentManager
- BountyManager

FHieser:
Solves #304


FHieser:
All in all this is probably going to work but its pretty Hacky in my opinion.
Maybe its best to not include it in main and just use it for the immediate deployment.
Opinions @marvinkruse @0xNuggan ?

0xNuggan:
I agree that it's a bit hacky, yeah. It's probably best to mantain  a separate "demo" branch and deploy from there, at least for now

PR #307, opened 2023-07-28
e2e test streamingPayments

### LINKED ISSUE #301: E2E Test StreamingPaymentProcessor
There is still a e2e test missing here

FHieser:
Solves #301


FHieser:
One thing i noticed while writing the tests is that the claimAll and the claimForSpecificWalletId functions in the StreamingPaymentProcessor need as a parameter IPaymentClient. I think for endusers it would be nice if they could just use the PaymentClient address to call upon these functions instead.
Your opinion?  @saxenism

saxenism:
> One thing i noticed while writing the tests is that the claimAll and the claimForSpecificWalletId functions in the StreamingPaymentProcessor need as a parameter IPaymentClient. I think for endusers it would be nice if they could just use the PaymentClient address to call upon these functions instead. Your opinion? @saxenism

Yep, that makes absolute sense. And this change shouldn't cause any major re-factoring. Perhaps, I'll open a new PR with these changes or maybe make this change in this PR itself and if nothing breaks, we can merge this PR into the main branch.

FHieser:
> > One thing i noticed while writing the tests is that the claimAll and the claimForSpecificWalletId functions in the StreamingPaymentProcessor need as a parameter IPaymentClient. I think for endusers it would be nice if they could just use the PaymentClient address to call upon these functions instead. Your opinion? @saxenism
> 
> Yep, that makes absolute sense. And this change shouldn't cause any major re-factoring. Perhaps, I'll open a new PR with these changes or maybe make this change in this PR itself and if nothing breaks, we can merge this PR into the main branch.

Yeah in the last Meeting we had a discussion about issues and PRs and we agreed upon making both as granular as possible to make it easier for future visitors of the repo to follow the changes.
So I would say lets create its own Issue and PR for it and change it there.
Ill create the issue and assign you @saxenism

FHieser:
@saxenism Could your review?

saxenism (approved):

PR #309, opened 2023-08-01
Rename Contributor to paymentReceiver

### LINKED ISSUE #275: OnlyContributor definition
We currently have an interesting usecase where the role of Contributor is based on if an address has a running payment in the streamingPaymentProcessor or not.
In the case of the RecurringPaymentManager Logicmodule we have a possibility that the payments are not triggered for longer periods of time, because no one uses the trigger function, which can lead to the case where someone on the "payroll" who should have the contributor access actually doesnt have it

Its not very critical, because the only place where its used is in the MilestoneManager anyway, but for later implementation this might be something to consider

FHieser:
Fixes #275

### DIFF
--- a/src/modules/paymentProcessor/IStreamingPaymentProcessor.sol
+++ b/src/modules/paymentProcessor/IStreamingPaymentProcessor.sol
@@ -11,13 +11,13 @@ interface IStreamingPaymentProcessor is IPaymentProcessor {
 
     // Structs
 
-    /// @notice This struct is used to store the payment order for a particular contributor by a particular payment client
+    /// @notice This struct is used to store the payment order for a particular paymentReceiver by a particular payment client
     /// @dev for _vestingWalletID, valid values will start from 1. 0 is not a valid vestingWalletID.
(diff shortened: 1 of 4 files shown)


src/modules/paymentProcessor/IStreamingPaymentProcessor.sol
```diff
@@ -89,37 +89,37 @@ interface IStreamingPaymentProcessor is IPaymentProcessor {
...
     );
 
-    /// @notice contributor's walletId for the paymentClient is no longer active
+    /// @notice paymentReceiver's walletId for the paymentClient is no longer active
     error Module__PaymentProcessor__InactiveWallet(
-        address paymentClient, address contributor, uint walletId
+        address paymentClient, address paymentReceiver, uint walletId
     );
 
-    /// @notice the contributor for the given paymentClient does not exist (anymore)
```
SchneiderNejc, line 107: the *paymentReceiver* for the given...

SchneiderNejc (approved):
I think it looks good.

Issue #231, opened 2023-04-28
Add a ILogicModule Interface


FHieser:
Because we started to group specific functionalities under Interfaces (Payment Processor, Authorizer etc) it might make sense to create on for the LogicModules too.

FHieser:
My take on this is that all Logicmodules will have to implement the IPaymentClient Interface in some form to work properly
We could rename that and create abstract Paymentclients for individual purposes
Say ERC20PaymentClient or ERC721PaymentClient for different usecases
This ties directly into issue #261
Any opinions @marvinkruse @0xNuggan ?

PR #312, opened 2023-08-02
Refactor PaymentClient to be logicModule base

### LINKED ISSUE #261: Move and Refactor Payment Client

FHieser:
Solves #261 
Solves #231


src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
```diff
@@ -208,4 +184,53 @@ abstract contract ERC20PaymentClient is
             revert Module__ERC20PaymentClient__InvalidRecipient();
         }
     }
+
+    //--------------------------------------------------------------------------
+    // {ERC20PaymentClient} Function Implementations
+
+    /// @dev Ensures `amount` of payment tokens exist in address(this).
+    function _ensureTokenBalance(uint amount) internal virtual {
+        uint balance = __Module_orchestrator.token().balanceOf(address(this));
```
saxenism: As discussed in the team meet earlier, we might do away with the pattern of storing the token address (IERC20) in the orchestrator contract in the future.

Possible reasons for this decision was to support multi-token payments and overall agreement that the proposal contract wasn't the best place to store the token address.

Therefore, I think querying the token from the orchestrator contract here, might not be a good move. I'd suggest querying this from the FundingManager contract.

src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
```diff
@@ -208,4 +184,53 @@ abstract contract ERC20PaymentClient is
...
+            // to address(this).
+            bool ok;
+            (ok, /*returnData*/ ) = __Module_orchestrator.executeTxFromModule(
+                address(__Module_orchestrator.fundingManager()),
+                abi.encodeWithSignature(
+                    "transferOrchestratorToken(address,uint256)",
+                    address(this),
+                    amount - balance
+                )
+            );
```
saxenism: Well, this is neat. But, there is an edge-case here.

Suppose we are dealing with tokens that take a fee on transfer ([examples](transferOrchestratorToken)) then even if you call the `transferOrchestratorToken` for `amount-balance` it wouldn't necessarily ensure that the contract has enough balance, because some fees will be deducted.

To remedy this, I recommend adding an assert statement that checks the contract balance again after the transfer, and only then you can *actually* be sure that the contract has enough balance.

Another remedy for this could be to simply create a whitelist of allowed proposal tokens and not allow any tokens that take a fee on transfer.

src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
```diff
@@ -208,4 +184,53 @@ abstract contract ERC20PaymentClient is
...
+    function _ensureTokenBalance(uint amount) internal virtual {
+        uint balance = __Module_orchestrator.token().balanceOf(address(this));
+
+        if (balance < amount) {
+            // Trigger callback from orchestrator to transfer tokens
+            // to address(this).
+            bool ok;
+            (ok, /*returnData*/ ) = __Module_orchestrator.executeTxFromModule(
+                address(__Module_orchestrator.fundingManager()),
+                abi.encodeWithSignature(
```
saxenism: I'd suggest using `abi.encodeCall` instead of `abi.encodeWithSignature` as it offers more safety. Typo and type errors will not compile.

I actually suggest to make this change across the repo if we using low-level calls anywhere else.

src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
```diff
@@ -208,4 +184,53 @@ abstract contract ERC20PaymentClient is
...
+        }
+    }
+
+    /// @dev Ensures `amount` of token allowance for payment processor(s).
+    function _ensureTokenAllowance(IPaymentProcessor spender, uint amount)
+        internal
+        virtual
+    {
+        IERC20 token = __Module_orchestrator.token();
+        uint allowance = token.allowance(address(this), address(spender));
```
saxenism, line 223: Not storing these values in separate variables can save gas. 

But it might make the code less neat than it is now.

Your call.

src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
```diff
@@ -208,4 +184,53 @@ abstract contract ERC20PaymentClient is
...
+
+        if (allowance < amount) {
+            token.safeIncreaseAllowance(address(spender), amount - allowance);
+        }
```
saxenism, line 227: This opens up another can of worms. Maybe something that we should decide as a team over a call.

So, the question is, do you want to also *decrease* the allowance of the *spender* if it is more than the `amount` ?

I'm asking this because the way the contracts are right now, depending on the front-end applications built on top of it, the users could be asked to give infinite approval to some contracts.

And if that is the case, there's this huge debate around safety vs convenience regarding infinite approvals. [This blog](https://hackernoon.com/erc20-infinite-approval-a-battle-between-convenience-and-security-lk60350r) encapsulates the essence of the argument.

src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
```diff
@@ -208,4 +184,53 @@ abstract contract ERC20PaymentClient is
             revert Module__ERC20PaymentClient__InvalidRecipient();
         }
     }
+
+    //--------------------------------------------------------------------------
+    // {ERC20PaymentClient} Function Implementations
+
+    /// @dev Ensures `amount` of payment tokens exist in address(this).
+    function _ensureTokenBalance(uint amount) internal virtual {
```
saxenism, line 194: Since this involves a low-level call to an unknown token, we should assume that it could be an ERC-777 token as well which is capable of re-entrancy.

Consider using a ReEntrancy guard on this function?

saxenism (commented):
Solid PR. Looks good to me.

I have flagged a few minor issues. Let's discuss whether they need to be handled or not.

src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
FHieser: This issue will be tackled in the upcoming discussion about the fee system.
In the MilestoneManager we currently solved this by having a seperate ensureTokenbalance call for the section where the fees are taken as well as the ensureTokenbalance call that is integrated in the addPaymentOrder Functions themselves

src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
FHieser: This calls the fundingManager not an unknown token
I dont think this is necessary here

src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
FHieser: I think this should be brought up in a new issue instead of here, because this was more of an refactoring PR than anything else.
I'll create one for it.

FHieser: But isnt this purely restricted to paymentProcessor that is connected to its own payment Processor?
I dont really see the problem here

FHieser:
@saxenism could you review?

src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
saxenism: Yeah, my comment was on the broader theme, but I think your point is valid and we can ignore that comment.

saxenism:
> @saxenism could you review?

Yep, reviewed again. Looks strong.

saxenism (approved):

PR #323, opened 2023-08-11
Feat/setup contracts bonding curve

Zitzak:
## Thoughts about the structure
- Most base contract (BondingCurveFundingManagerBase) implements functions concerning the issuing (buying) of tokens, which every bonding curve funding manager must have. It also defines an abstract function `_issueTokensFormulaWrapper` which must be used in downstream contracts to define through which formula price is calculated
- The next base contract (RedeemingBondingCurveFundingMangerBase) implements the functions concerning the redemption of tokens back for their collateral. It also defines an abstract function `_redeemTokensFormulaWrapper` which must be used in downstream contracts to define through which formula price is calculated. The reason to have two of these functions is because the Bancor contract has a different interface for calculating issuing tokens vs redeeming tokens
- There are 2 more abstract contracts which introduce a virtual issuing token supply (VirtualTokenSupplyBase) and a virtual collateral supply (VirtualCollateralSupplyBase). The would be considered even more specialised in terms of functionality and therefor would be added more in a compositing fashion instead of strict inheritence
- The final contract has all the features above and is called BancorVirtualSupplyBondingCurveFundingManger


Zitzak:
For this draft PR the only interesting commits are the last 5, excluding the very last one. Happy to rebase if needed for clarity of review @FHieser

src/modules/fundingManager/bondingCurveFundingManager/formula/SafeMath.sol
FHieser, line 1: I think safemath isnt needed anymore since compiler version 0.8 so you can probably remove this and refactor the bancor Formular accordingly

src/modules/fundingManager/bondingCurveFundingManager/formula/BancorUtils.sol
FHieser, line 1: This can just directly be included in the Bancor formular contract in my opinion.

src/modules/fundingManager/bondingCurveFundingManager/formula/BancorFormula.sol
FHieser, line 1: Where exactly did you get this from?
Are there any tests for it?

src/modules/fundingManager/token/IVirtualSupplyToken.sol
FHieser, line 1: Do these token contracts serve any purpose currently?

src/modules/fundingManager/token/VirtualSupplyBase.sol
FHieser, line 1: Do these token contracts serve any purpose currently?

src/modules/fundingManager/token/VirtualSupplyTokenUpgradeable.sol
FHieser, line 1: Do these token contracts serve any purpose currently?

test/modules/fundingManager/BondingCurveFundingManager.t.sol
FHieser, line 1: Outcommented

test/modules/fundingManager/PrimaryMarketMaker.t.sol
FHieser, line 1: Outcommented

test/modules/fundingManager/token/SetVirtualSupply.t.sol
FHieser, line 1: Outcommented

test/modules/fundingManager/token/VirtualSupplyTokenDeployment.t.sol
FHieser, line 1: Outcommented

test/modules/fundingManager/token/VirtualSupplyTokenTest.sol
FHieser, line 1: Outcommented

test/modules/fundingManager/token/utils/mocks/VirtualSupplyTokenUpgradeableMock.sol
FHieser, line 1: Outcommented

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/IVirtualTokenSupply.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+
+pragma solidity 0.8.19;
+
+interface IVirtualTokenSupply {
```
FHieser, line 5: I would add the getters here too

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/IVirtualCollateralSupply.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+
+pragma solidity 0.8.19;
+
+interface IVirtualCollateralSupply {
```
FHieser, line 5: I would add the getters here too

src/modules/fundingManager/bondingCurveFundingManager/formula/BancorFormula.sol
FHieser: Do we want to do either:
Import it into lib
or integrate it into the Repository directly (+ Tests)

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/IVirtualCollateralSupply.sol
FHieser: If you want to Force the usage of modifiers in these functions you can instead create virtual/abstract functions
maybe in that regard you want to use super.setVirtualCollateralSupply()

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
```diff
@@ -0,0 +1,143 @@
...
+    function decimals()
+        public
+        view
+        override(ERC20Upgradeable)
+        returns (uint8)
+    {
+        return tokenDecimals;
+    }
+
+    function deposit(uint amount) external {}
```
FHieser, line 84: Can you remind me to bring this up in the upcoming Devmeeting that these functions are not necessary in the IFundingManager Interface every time

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
```diff
@@ -0,0 +1,143 @@
...
+        buyIsOpen = true;
+    }
+
+    function _closeBuy() internal {
+        // Validate if not closed
+        buyIsOpen = false;
+    }
+
+    function _setCollateral(address _collateral) internal {
+        collateral = _collateral;
```
FHieser, line 113: Isnt the collateral just the address(this).balance?
Wasnt that the main point of the Virtual Collateral to be able to be set?

FHieser (commented):
All in all looks good

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
Zitzak: Could you explain this point again for me please. You are right that the collateral address is just the token of defined by the Orchestrator contract, However we talked about also the future possibility to add other collateral tokens in our first call, which is why I included this setter function.

I'm unsure what is meant by balance

FHieser: Forget what I said here 🙈
This is to set the underlying payment token address right?
Because I apparently cant read I assumed that this would somehow set a "virtual" Collateral Amount here, which it straight up doesnt

src/modules/fundingManager/bondingCurveFundingManager/formula/SafeMath.sol
0xNuggan: I'm not sure we should touch the Bancor Formula if not necessary...

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
0xNuggan: Yeah, let's discuss this in the meeting. I may be ready to see the light myself in that regard.

0xNuggan:
Looks good! I was slightly confused about moving virtual balance and supply into separate contracts, but it's true that it keeps everything somewhat tidier.

src/modules/fundingManager/bondingCurveFundingManager/formula/BancorFormula.sol
0xNuggan: Yes, maybe we can just import this as a library? What do you think @Zitzak  ?

src/modules/fundingManager/bondingCurveFundingManager/formula/SafeMath.sol
Zitzak: Ended up adding the files back to the repo, see conversation on [Discord](https://discord.com/channels/788422482087641089/1006521951445078117/1141012690505039943)

I have added a detailed description of the changes I made within each contract, as well as added the files to the ignore list for the `fmt`, so it wouldn't completely change the file

I will document these changes within our documentation

src/modules/fundingManager/bondingCurveFundingManager/formula/BancorFormula.sol
Zitzak: See first comment

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/IVirtualTokenSupply.sol
Zitzak: Done

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/IVirtualCollateralSupply.sol
Zitzak: Done

remappings.txt
```diff
@@ -1,2 +1,2 @@
 @oz-up/=lib/openzeppelin-contracts-upgradeable/contracts/
-@oz/=lib/openzeppelin-contracts/contracts/
+@oz/=lib/openzeppelin-contracts/contracts/
```
FHieser, line 2: lol what is happening here?
it added it and then removed it again?

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -0,0 +1,158 @@
...
+        _addCollateralAmount(_depositAmount);
+    }
+
+    function sellOrder(uint _depositAmount)
+        external
+        payable
+        override(RedeemingBondingCurveFundingManagerBase)
+    {
+        // WiP
+        // Q: Deduct fee from token or collateral?
```
0xNuggan, line 101: When selling, deducting the fee from the collateral after the token has been burned simplifies the process: you use the math as intended and end up with collected fees in only one token.

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
```diff
@@ -0,0 +1,143 @@
...
+    using Strings for uint;
+    using SafeERC20 for IERC20;
+
+    //--------------------------------------------------------------------------
+    // Storage
+    uint8 internal tokenDecimals;
+    bool internal buyIsOpen;
+    uint internal buyFee;
+    address internal collateral;
+    // Q: Do we want to set other beneficiary in the contract then the module itself for the fees?
```
0xNuggan, line 43: I think transferOrchestratorToken() is good enough to move the fees out when the time comes (if the fees are going to go to the worklflow owners anyway). We'd probably need a way to track the collected amount, though

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
```diff
@@ -0,0 +1,143 @@
...
+    function _issueTokens(uint _depositAmount, address _collateral)
+        internal
+        returns (uint mintAmount)
+    {
+        mintAmount = _issueTokensFormulaWrapper(_depositAmount, _collateral);
+        _mint(msg.sender, mintAmount);
+    }
+
+    function _setTokenDecimals(uint8 _decimals) internal {
+        // Input check
```
0xNuggan, line 129: What's the thought behind this? This would be the decimals of the token we mint and give out, right? Is there a reason we wouldn't want to stick with 18?

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
```diff
@@ -0,0 +1,143 @@
...
+        tokenDecimals = _decimals;
+    }
+    //--------------------------------------------------------------------------
+    // OnlyOrchestrator Mutating Functions
+
+    function transferOrchestratorToken(address to, uint amount)
+        external
+        onlyOrchestrator
+    {
+        __Module_orchestrator.token().safeTransfer(to, amount);
```
0xNuggan, line 139: Since we are going to remove orchestrator.token() anyway, maybe go ahead and use the collateral address directly?

0xNuggan:
I like it a lot.  Apart from the comments I left, one question about the fees: Do you want to implement them already for this first version, or leave them for later? 
Also because we need on "main" test for the BancorVirtualS[...]anager.sol and it would probably save time if we have a fee structure in place already at that point (and can test with everything from 0 fees upwards)

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
Zitzak: Thank you for the feedback. Deducting fee will be implemented in a different PR

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
Zitzak: I agree and will not be implementing a beneficiary on top of the standard way to pull out tokens

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
Zitzak: I brought this up in a call with Felix. He suggested that I could stick to what was implemented before and follow the same pattern. To be more specific, the RebaseFundingManager also enables the setting of decimals so I decided to do the same

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
Zitzak: So the initial idea was to add a collateral address to the contract. The reason I discussed with Felix was because in the future we might want to make use of multiple collateral tokens.

I have however removed the reference to the collateral token in the code I'm currently writing (next PR) because setting up the contract for a future with multiple collateral tokens makes me assume to many things. I decided instead to go with what we need right now which means that the code should work with one collateral token, which is the one defined by the Orchestrator.

For this reason, I would leave the transfer of the collateral like this, and update once we remove the `orchestrator.token()` to however it will be implemented once this is needed.

Happy to hear your thoughts around this? :)

Zitzak:
So currently working on the next PR already, because we decided on smaller PRs, I want to merge this one first.

In the next PR I have already implemented the `buyFee` for the `BondingCurveFundingMangerBase.sol` contract, and aim to also implement the `sellFee` within the inherited `RedeemingBondingCurveFundingManager.sol`
@0xNuggan

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
0xNuggan: Ok yeah, makes sense to keep it consistent

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
0xNuggan: Yeah, I think we can keep it like that then. Once we remove `orchestrator.token()` we can refactor it, but not get ahead of ourselves until then. 
Having multiple tokens seems a change big enough to wait for the rest to be in place. Also I think it's out of the scope of the BCRG doc?

0xNuggan:
Ok then it's all clear :)

0xNuggan (approved):

FHieser (approved):

FHieser:
I think in general this is a solid foundation for the upcoming finalization of the BondingCurveFundingManager (BCFM).
The problem I have with it is that the BCFM is technically neither finished in its final functionality nor tested on an individual contract level. This is not your fault Marvin, because we didnt get to this point of talking yet (so its probably my fault, cause I didnt step in directly when you created the Review request, but I didnt have the time 🥲), but in the past we tried to only merge in case the PR is truly finished (functionality + testing). 
I will merge this now, because it seems its necessary for you to proceed to the next stage of its implementation,
but this should be the exception in upcoming PRs.

PR #326, opened 2023-08-15
icebox MilestoneManager

0xNuggan:
As discussed with @FHieser  , this PR removes the MilestoneManager for now since we don't have users for it right now and it was frequently creating issues due to the contract size. For the scripts and e2e tests, it has been replaced with the bountyManager


script/setup/SetupToyOrchestratorScript.s.sol
```diff
@@ -217,45 +211,29 @@ contract SetupToyOrchestratorScript is Test, DeploymentScript {
...
-        );
-        assertTrue(orchestratorCreatedMilestoneManager.isExistingMilestoneId(1));
-
-        assertTrue(
-            orchestratorCreatedMilestoneManager.isContributor(1, alice.addr)
-        );
-        assertTrue(
-            orchestratorCreatedMilestoneManager.isContributor(2, alice.addr)
-        );
+        /*  IBountyManager.Bounty memory confirmBounty =
```
FHieser: outcommented

FHieser (approved):
Looks good
One File still contains some out commented code

script/setup/SetupToyOrchestratorScript.s.sol
0xNuggan: you're right, fixed

Issue #317, opened 2023-08-08
Code review results - Orchestrator


0xNuggan:
Another picture dump for now - for the next one they'll probably do the PR directly in this repo so everybody can comment

![Screenshot from 2023-08-08 05-53-08](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/71cf8a66-2af0-492f-a979-d7fa1c841e56)
![Screenshot from 2023-08-08 05-53-13](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/93c39970-0c30-405d-945b-78ee77346544)
![Screenshot from 2023-08-08 05-53-22](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/8fb632d4-361f-4eb1-a80a-b7576b71df1c)
![Screenshot from 2023-08-08 05-53-32](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/ba504575-82c7-4fe4-92db-7de1d01b0f7c)
![Screenshot from 2023-08-08 05-53-41](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/85ca3e24-8344-4c4b-983e-d8f9d0043af2)
![Screenshot from 2023-08-08 05-53-50](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/f2b14b85-14ea-4b4d-b798-140e24bf6496)
![Screenshot from 2023-08-08 05-59-10](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/10971cee-e7e9-4273-af2a-a2ea6b6ff375)
![Screenshot from 2023-08-08 05-59-14](https://github.com/InverterNetwork/inverter-contracts/assets/82726722/f1eeb98e-cff0-42ad-9ac5-fb54fedf7270)

saxenism:
Ok, so here is the reasoning for adding the `verifyAddresIsXYZModule` functions in the Orchestrator contract:

The ideal work-flow involving these functions are:
1. A new module needs to be added to the inverter work-flow
2. That module has a cross-dependency of some other module (ie it requires late dependency injection)
3. So, the logic of initializing this function would require some additional logic referring to the other contracts in the function `init2`
4. Now, the process of using another module for the initialization of the new module would involve: finding the dependency module address and then verifying whether the address is indeed of the type of module that we require.
5. So, for the finding part we use the function: `findModuleAddressInProposal`
[...]

PR #327, opened 2023-08-15
Created an option to find modules via identifiers

saxenism:
As a fix for the issue mentioned by the auditors in issue #317 regarding the `findModuleAddressInOrchestrator` not designed to handle the cases where multiple modules with the same name are being used, I have made a few changes to `findModuleAddressInOrchestrator`.

Since the URL of a module is much less likely to be the same as other modules, I have replaced the query string from the module title to the module URL.

And since, it is technically possible to have multiple modules with the same URL, i have also given an option to search for module address using a **unique module identifier**.


FHieser:
Where tests

src/orchestrator/Orchestrator.sol
FHieser, line 1: The tests are missing

FHieser (changes requested):

src/orchestrator/Orchestrator.sol
saxenism: image

The tests are no longer missing 👉👈

saxenism:
Considering this comment:


Added the documentation for `dependencyData`.

The `dependencyData` now expects 3 fields:
1. bool hasDependency: True if dependencies are expected, false otherwise
2. string[] dependenciesURLs: URL of the modules that can be counted as dependencies
3. bytes additionalData: Any additional data that you might want access to, inside of `init2`.

FHieser:
The changes will be implemented with the #371 PR

PR #331, opened 2023-08-16
Authorizer codereview fixes

### LINKED ISSUE #316: Code review results - RoleAuthorizer
Another picture dump for now - for the next one they'll probably do the PR directly in this repo so everybody can comment

Screenshot from 2023-08-08 05-55-03
Screenshot from 2023-08-08 05-55-08
Screenshot from 2023-08-08 05-55-12
Screenshot from 2023-08-08 05-55-22
Screenshot from 2023-08-08 05-57-07
Screenshot from 2023-08-08 05-57-15
Screenshot from 2023-08-08 05-57-18
Screenshot from 2023-08-08 05-57-23
Screenshot from 2023-08-08 05-57-30
Screenshot from 2023-08-08 05-57-37
Screenshot from 2023-08-08 05-57-42
Screenshot from 2023-08-08 05-57-51
Screenshot from 2023-08-08 05-57-56
Screenshot from 2023-08-08 05-58-02
Screenshot from 2023-08-08 05-58-07
Screenshot from 2023-08-08 05-58-12
Screenshot from 2023-08-08 05-58-16

0xNuggan:
Changes to the RoleAuthorizer after Audit review. closes #316 
Big changes:
-**removed the toggleSelfManagement logic**: all modules can freely register their own roles by default, and still use the global ones if wanted.
-**removed the isAuthorized function**: Since we are always using roles, having extra authorization methods that don't explicitly state the role and default to some value is unnecessary and confusing. Better always be clear and use hasRole().
-**Roles are now stored as constant bytes32 values**: Since enums are just fancy uint8, it became easier to use uint8(0) instead of using the verbose Module.Roles.ROLE_0 syntax. Now the roles are just public constant bytes32 values declared in the contract (like the OWNER) that can easily be queried. This should be more intuitive and readable. 

Smaller changes:
-Modules can now set the threshold for their token gated roles with the function setThresholdFromModule.
-Global roles can also be made token gated
-Global owner and manager roles no longer encode the proposal address. They are now constants. (Modules still encode their roles to avoid collisions)


lib/forge-std
FHieser, line 1: Why the lib changes?

src/modules/authorizer/TokenGatedRoleAuthorizer.sol
```diff
@@ -57,29 +65,17 @@ contract TokenGatedRoleAuthorizer is
...
-        if (selfManagedModules[_msgSender()]) {
-            roleId = generateRoleId(_msgSender(), role);
-            //check if token gated:
-            if (isTokenGated[roleId]) {
-                return hasTokenRole(roleId, who);
-            } else {
-                return hasRole(roleId, who);
-            }
-            // If not, check the account against the orchestrator roles
+        if (isTokenGated[roleId]) {
```
FHieser, line 75: When you combine this with the onlytokengatedRole from the hasTokenRole then your checking if it is tokenGatedmultiple time in a single function
you might wanna do an internal _hasTokenRole and just use the modifier on the external version

test/modules/authorizer/SingleVoteGovernor.t.sol
```diff
@@ -224,13 +224,13 @@ contract SingleVoteGovernorTest is ModuleTest {
     function testInit() public override(ModuleTest) {
         assertEq(_orchestrator.isModule(address(_governor)), true);
 
-        assertEq(_authorizer.isAuthorized(address(_governor)), true);
+        assertEq(_authorizer.hasRole("0x01", address(_governor)), true); // Owner role
```
FHieser: Direclty using role 0x01 is pretty hard to understand without any prior knowledge, maybe you can instead reference the authorizer versions directly (authorizer.OWNER())

FHieser (approved):
Looks good
Small details

FHieser (approved):

PR #332, opened 2023-08-16
Remove _ensureTokenBalance checks when adding payment Orders and refactor triggerFor()

### LINKED ISSUE #324: EnsureTokenbalance not needed in createPaymentOrders function
The intended purpose of ensureTokenbalance was to pre-shift tokens to the paymentClient Module, so they would be "reserved" for the later payment via the Payment Processor. This works if all Payment Orders are created simultaneously as the ensureTokenbalance function is a called a single time to update up to the threshold needed to fullfill them. But it wont work in case multiple paymentOrders are created on seperate occasions. Take the following example: We create a Payment Order A which has a payment Amount AA. The PaymentClient will be filled up with tokens until the treshhold of AA is reached Now we create a PaymentOrder B which has a Payment Amount of BB AA is bigger than BB so AA>BB The PAymentClient will try to fill up to the treshhold of BB but it already has AA Tokens in there. Therefor the amount is already reached accordingly and nothing happens So the amount of token in the PaymentClient is not AA+BB but just AA

The reason we didnt notice this until now is that the collectPaymentOrders Function of the PaymentClient also calls ensureTokenBalance for all existing PaymentOrders again (in this case it raises the amount to AA+BB), but that is already too late, because we didnt "reserve" the token properly.

We had a talk about it and we agreed to remove the ensureTokenbalance functionality from the create payment orders altogether.
As it currently stands we immediately call processPayments after we create the PaymentOrders, which removes the time in which "reserves" could be of use anyway.

Based on the ensureTokenbalance functionality i made a refatoring of the TriggerFor Function in the RecurringPaymentManager that increased the complexity of the function in favor of improving on the performance of the function
with the removing of the ensureTokenBalance function in the CreatePaymentOrders this can be reverted

_Originally posted by @FHieser in https://github.com/InverterNetwork/inverter-contracts/issues/319#issuecomment-1674478592_

0xNuggan:
closes #324


lib/forge-std
FHieser, line 1: Why are you updating the submodules here?

src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
```diff
@@ -100,9 +97,6 @@ abstract contract ERC20PaymentClient is IERC20PaymentClient, Module {
 
         // Add total orders' amount to current outstanding amount.
         _outstandingTokenAmount += totalTokenAmount;
```
FHieser, line 99: totalAmount is not necessary anymore here

src/modules/logicModule/RecurringPaymentManager.sol
```diff
@@ -283,55 +283,42 @@ contract RecurringPaymentManager is
             ++index;
```
FHieser: Everything above this (up to line 234) is not necessary anymore

FHieser (changes requested):

src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
FHieser, line 1: Also _ensureTokenBalance didnt get changed to just add tokens to the contracts

src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
FHieser: Forget what I said here

src/modules/logicModule/paymentClient/ERC20PaymentClient.sol
0xNuggan: As I understood it, since we are now checking _ensureTokenBalance at the time of processing the orders (and starting the payout), there was no need to remove the (amount > balance ) check, and that way we still keep the possibility of using available funds (from f.ex. cancellations). Example
-PaymentOrder A gets added, amount: 50. No funds are pulled
-PaymentOrder B gets added, amount: 100. no funds are pulled
-PaymentOrders get Processed: _ensureTokenBalance of 150 gets called, and the 150 tokens get pulled
---> **IF** there already were 75 tokens from a cancellation/refund/whatever, we'd only need to call _ensureTokenBalance with 75.

FHieser: Yeah
That would be in case we want to drain the contract of unnecessary funds, that found their way there for some reason
But in case we would like to use the "reserve" functionality of the ensureTokenBalance that was initially proposed, this would not work
I think we should stick to the changing it to its original intended purpose

FHieser (approved):

PR #337, opened 2023-08-21
add checks to setters

### LINKED ISSUE #336: Orchestrator setters don't validate module type
The setters don't check that the address we are setting actually implements the interface. This makes it possible to set an authorizer that doesn't implement the IAuthorizer interface, effectively bricking the proposal

0xNuggan:
closes #336


FHieser:
Where tests

FHieser (approved):
I mean simple enough but please include the tests accordingly

saxenism (approved):

PR #339, opened 2023-08-25
add new scripts for the bountyManager

0xNuggan:
After some merge shenanigans with the last code updates I chose to just rebase and open a new PR.

This PR adds a couple of utility scripts and modifies the setupToyProposal script to correspond to the Bloom usecase


saxenism (approved):
This PR looks good to me.

One change I would like to suggest to make this neater would be to create a new **constants.sol** file that contains all the addresses and values hardcoded in the scripts right now. Then, simply import the values from that file to use across all the scripts.

Ofc, this does not have any implication on performance/security.

0xNuggan:
That's a great suggestion @saxenism ! I'll merge this now and create an issue

PR #342, opened 2023-09-06
Test/add test for bonding curve contract setup

0xNuggan:
Tests for the Bonding Curve contracts. Notes:

- I think I found a logic issue in _convertAmountToRequiredDecimal and provided an alternative. I may be missing something, please review. @Zitzak

- The formula starts reverting when supplying some numbers that don't seem that high to me, specially when selling (assuming 18 decimals). I still want to look closer into it, but extra eyes can only help.

- Since there is some rounding and precision loss involved, some test use foundry 'assertApproxEqRel' instead of 'assertEq'.

- The 4 functions in the coverage of the Bancor funding manager are the stubs for the IFundingManager interface. We should think about removing those functions from the interface and only leave transferOrchestratorToken().

- There are a couple of other bugfixes and changes, usually pretty straightforward. They are also tagged with the @review tag


src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
```diff
@@ -81,63 +105,132 @@ abstract contract BondingCurveFundingManagerBase is
...
+    /// @param _depositAmount The amount of collateral to deposit for buying tokens.
+    function _buyOrder(address _receiver, uint _depositAmount)
+        internal
+        returns (uint mintAmount)
+    {
+        if (_depositAmount == 0) {
+            revert BondingCurveFundingManager__InvalidDepositAmount();
+        }
+        // Transfer collateral, confirming that correct amount == allowance
+        __Module_orchestrator.token().safeTransferFrom(
```
FHieser, line 157: __Module_orchestrator.token() wont exist  in the future.
It might make sense to reference a token specified in the BondingCurveFundingManagerBase

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
```diff
@@ -20,58 +20,82 @@ import {ERC20Upgradeable} from "@oz-up/token/ERC20/ERC20Upgradeable.sol";
...
+/// @author Inverter Network.
+/// @notice This contract enables the base functionalities for issuing tokens along a bonding curve.
+/// @dev The contract implements functionalties for:
+///         - opening and closing the issuance of tokens.
+///         - setting and subtracting of fees, expressed in BPS and subtracted from the collateral.
+///         - calculating the issuance amount by means of an abstract function to be implemented in
+///             the downstream contract.
 abstract contract BondingCurveFundingManagerBase is
     IBondingCurveFundingManagerBase,
-    IFundingManager,
```
FHieser, line 29: It says FundingManagerBase in the title,
Why is this not a Ifundingmanager?

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
```diff
@@ -81,63 +105,132 @@ abstract contract BondingCurveFundingManagerBase is
...
-        mintAmount = _issueTokensFormulaWrapper(_depositAmount, _collateral);
-        _mint(msg.sender, mintAmount);
+        mintAmount = _issueTokensFormulaWrapper(_depositAmount);
     }
 
+    /// @dev Sets the number of decimals for the token.
+    /// This function updates the `tokenDecimals` state variable.
+    /// @param _decimals The number of decimals to set for the token.
     function _setTokenDecimals(uint8 _decimals) internal {
-        // Input check
```
FHieser, line 129: Does this not need an input check?
Check for zero
Check / Reference possible Formular borders

src/modules/fundingManager/bondingCurveFundingManager/IBancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -0,0 +1,43 @@
...
+    //--------------------------------------------------------------------------
+    // Errors
+
+    /// @notice Reserve ratio can not be be bigger than 100% expressed in PPM
+    error BancorVirtualSupplyBondingCurveFundingManager__InvalidReserveRatio();
+
+    //--------------------------------------------------------------------------
+    // Events
+
+    // TODO: Contract still needs events. Where are events needed?
```
FHieser: TODO

src/modules/fundingManager/bondingCurveFundingManager/RedeemingBondingCurveFundingManagerBase.sol
```diff
@@ -2,70 +2,193 @@
...
+    /// contract for redemption.
+    function _sellOrder(address _receiver, uint _depositAmount)
+        internal
+        returns (uint redeemAmount)
+    {
+        if (_depositAmount == 0) {
+            revert RedeemingBondingCurveFundingManager__InvalidDepositAmount();
+        }
+        // Transfer issued token to contract, confirming deposit amount == allowance
+        IERC20(address(this)).safeTransferFrom(
```
FHieser: Does this work?
Isnt this an internal call?

FHieser: Also just out of curiosity:
Shouldnt the tokens be burned here?

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/IVirtualTokenSupply.sol
```diff
@@ -3,9 +3,31 @@
...
+    /// @notice Subtracting would result in an underflow.
     error VirtualTokenSupply__SubtractResultsInUnderflow();
+
+    /// @notice Adding would result in and overflow.
     error VirtualTokenSupply_AddResultsInOverflow();
 
+    //--------------------------------------------------------------------------
+    // Events
+
+    // TODO: Contract still needs events. Where are events needed?
```
FHieser: TODO

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/VirtualCollateralSupplyBase.sol
```diff
@@ -4,36 +4,66 @@ pragma solidity 0.8.19;
...
-    uint internal virtualCollateralSupply;
-    uint private constant MAX_UINT = 2 * 256 - 1;
+    //--------------------------------------------------------------------------
+    // Storage
 
-    function setVirtualCollateralSupply(uint _virtualSupply) external virtual;
+    /// @dev The internal state variable to keep track of the virtual collateral supply.
+    uint internal virtualCollateralSupply;
+    /// @dev Maximum unsigned integer value for overflow checks.
+    uint private constant MAX_UINT = type(uint).max; // TODO: The maximum supply has to be divided by the BPS precision used. If not, it can come to overflow errors in downstream contracts
```
FHieser: TODO

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/VirtualCollateralSupplyBase.sol
```diff
@@ -4,36 +4,66 @@ pragma solidity 0.8.19;
...
     function getVirtualCollateralSupply() external view returns (uint) {
         return _getVirtualCollateralSupply();
     }
 
-    function _addCollateralAmount(uint _amount) internal {
+    //--------------------------------------------------------------------------
+    // Public Functions Implemented in Downstream Contract
+
+    /// @inheritdoc IVirtualCollateralSupply
+    function setVirtualCollateralSupply(uint _virtualSupply) external virtual;
```
FHieser, line 33: why is this virtual?
Is this coupled to a downstream contract and if so maybe an explanation in the docs is in order

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/VirtualTokenSupplyBase.sol
```diff
@@ -4,35 +4,66 @@ pragma solidity 0.8.19;
...
+/// various internal and external functions for managing it.
 abstract contract VirtualTokenSupplyBase is IVirtualTokenSupply {
+    //--------------------------------------------------------------------------
+    // Storage
+
+    /// @dev The internal state variable to keep track of the virtual token supply.
     uint internal virtualTokenSupply;
-    uint private constant MAX_UINT = 2 ** 256 - 1;
+    /// @dev Maximum unsigned integer value for overflow checks.
+    uint private constant MAX_UINT = type(uint).max; // TODO: The maximum supply has to be divided by the BPS precision used. If not, it can come to overflow errors in downstream contracts
```
FHieser: TODO

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/VirtualTokenSupplyBase.sol
```diff
@@ -4,35 +4,66 @@ pragma solidity 0.8.19;
...
     function getVirtualTokenSupply() external view returns (uint) {
         return _getVirtualTokenSupply();
     }
 
-    function _addTokenAmount(uint _amount) internal {
+    //--------------------------------------------------------------------------
+    // Public Functions Implemented in Downstream Contract
+
+    /// @inheritdoc IVirtualTokenSupply
+    function setVirtualTokenSupply(uint _virtualSupply) external virtual;
```
FHieser, line 34: See virtual collateral Token supply

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -55,104 +94,386 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
...
         payable
         override(BondingCurveFundingManagerBase)
+        buyingIsEnabled
     {
-        // WiP
-        // Deduct fee from incoming value. Fee is paid in collateral token
-        uint amountIssued = _issueTokens(_depositAmount, collateral);
-        _addTokenAmount(amountIssued);
-        _addCollateralAmount(_depositAmount);
+        _virtualSupplyBuyOrder(_msgSender(), _depositAmount);
```
FHieser: hm maybe go with:
```
super().buyOrder(...);
_virtualSupplyBuyOrder(...);
```
out of the name _virtualSupplyBuyOrder i would just assume that it creates a buyorder for the virtualsupply only

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -55,104 +94,386 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
...
+    /// @param _receiver The address that will receive the redeemed tokens.
+    /// @param _depositAmount The amount of issued token to deposited.
+    function sellOrderFor(address _receiver, uint _depositAmount)
+        external
+        payable
+        override(RedeemingBondingCurveFundingManagerBase)
+        validReceiver(_receiver)
+        sellingIsEnabled
+    {
+        _virtualSupplySellOrder(_receiver, _depositAmount);
```
FHieser: See comment above

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -55,104 +94,386 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
...
         payable
         override(RedeemingBondingCurveFundingManagerBase)
+        sellingIsEnabled
     {
-        // WiP
-        // Q: Deduct fee from token or collateral?
-        uint redeemAmount = _redeemTokens(_depositAmount, collateral);
-        _subTokenAmount(_depositAmount);
-        _subCollateralAmount(redeemAmount);
+        _virtualSupplySellOrder(_msgSender(), _depositAmount);
```
FHieser: same here

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -55,104 +94,386 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
...
+        _subVirtualTokenAmount(_depositAmount);
+        _subVirtualCollateralAmount(redeemAmount);
+    }
+
+    /// @dev Sets the reserve ratio for buying tokens.
+    /// The function will revert if the ratio is greater than the constant PPM.
+    ///
+    /// @param _reserveRatio The reserve ratio to be set for buying tokens. Must be <= PPM.
+    function _setReserveRatioForBuying(uint32 _reserveRatio) internal {
+        // TODO: Qs - TEST: What happens when set to 0? -> Reserve ratio of 0 is not allowed
```
FHieser: TODO

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -55,104 +94,386 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
...
+        reserveRatioForBuying = _reserveRatio;
+    }
+
+    /// @dev Sets the reserve ratio for selling tokens.
+    /// Similar to its counterpart for buying, this function sets the reserve ratio for selling tokens.
+    /// The function will revert if the ratio is greater than the constant PPM.
+    ///
+    /// @param _reserveRatio The reserve ratio to be set for selling tokens. Must be <= PPM.
+    function _setReserveRatioForSelling(uint32 _reserveRatio) internal {
+        // TODO: Qs - TEST: What happens when set to 0? -> Reserve ratio of 0 is not allowed
```
FHieser: TODO

test/e2e/E2eTest.sol
```diff
@@ -87,6 +91,43 @@ contract E2eTest is Test {
...
+        address(0x3BEAC0);
+    IModule.Metadata bancorVirtualSupplyBondingCurveFundingManagerMetadata =
+    IModule.Metadata(
+        1,
+        1,
+        "https://github.com/inverter/bonding-curve-funding-manager",
+        "BancorVirtualSupplyBondingCurveFundingManager"
+    );
+
+    function setUpBondingCurveFundingManager() private {
```
FHieser: setUpBancorVirtualSupplyBondingCurveFundingManager

test/modules/fundingManager/bondingCurveFundingManager/marketMaker/utils/mocks/BancorVirtualSupplyBondingCurveFundingManagerMock.sol
```diff
@@ -0,0 +1,54 @@
...
+    "src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol";
+import {IBancorFormula} from
+    "src/modules/fundingManager/bondingCurveFundingManager/formula/IBancorFormula.sol";
+import {Module} from "src/modules/base/Module.sol";
+
+contract BancorVirtualSupplyBondingCurveFundingManagerMock is
+    BancorVirtualSupplyBondingCurveFundingManager
+{
+    //--------------------------------------------------------------------------
+    // The BancorVirtualSupplyBondingCurveFundingManager is not abstract, so all the necessary functions are already implemented
```
FHieser, line 23: Just FYI:
Instead of creating a mock here to test internal functions, sometimes you can also just inherit from the contract in the test file and that way you can call it internally

test/e2e/BondingCurve.t.sol
```diff
@@ -0,0 +1,138 @@
...
+        }
+        vm.stopPrank();
+
+        // If the orchestrator spends half of the deposited tokens in the fundingmanager, i.e. for a milestone,
+        // alice and bob are still able to withdraw their respective leftover
+        // of the tokens.
+        // Note that we simulate orchestrator spending by just burning tokens.
+        uint halfOfDeposit = token.balanceOf(address(fundingManager)) / 2;
+        fundingManager.setVirtualCollateralSupply(halfOfDeposit);
+        /*         token.burn(
```
FHieser: outcommented the burning of the tokens

test/modules/fundingManager/bondingCurveFundingManager/RedeemingB_CurveFundingManagerBase.t.sol
```diff
@@ -0,0 +1,461 @@
...
+        assertEq(
+            _token.balanceOf(address(bondingCurveFundingManager)),
+            (bondingCurveBalanceBefore - redeemAmount)
+        );
+        assertEq(bondingCurveFundingManager.balanceOf(seller), 0);
+        assertEq(_token.balanceOf(seller), 0);
+    }
+
+    /* Test sellOrder and _sellOrder function
+        ├── when the sell amount is 0
```
FHieser, line 187: @0xNuggan did you write this tree yourself?

test/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.t.sol
```diff
@@ -0,0 +1,1036 @@
...
+    @NOTE: The functions:
+
+    - deposit(uint amount) external {}
+    - depositFor(address to, uint amount) external {}
+    - withdraw(uint amount) external {}
+    - withdrawTo(address to, uint amount) external {} 
+
+    are not tested since they are empty and will be removed in the future.
+
+    Also, since the following functions just wrap the formula contract, their content is assumed to be tested in the formula tests, not here:
```
FHieser: Currently there arent any formular tests

FHieser (commented):
Structure wise it looks good
This was more of a general review and things i noticed while looking broadly over it

test/e2e/BondingCurve.t.sol
0xNuggan: Forgot to remove the comment. Since we are using a virtual supply we don't need to actually burn anything.

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
0xNuggan: This ties to the comment above. Probably a good call, if BondingCurveFundingManagerBase becomes a IFundingManager, it can call self.token() and avoid getting it through the proposal, which will be deprecated anyway. What do you think @Zitzak  ?

src/modules/fundingManager/bondingCurveFundingManager/RedeemingBondingCurveFundingManagerBase.sol
0xNuggan: If there's enough allowance, it works. To your second point: you're absolutely right, we should burn there. The virtual balances are accounted for in the Bancor tests, but the actual on-chain one slipped through the cracks. :facepalm:

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/VirtualCollateralSupplyBase.sol
0xNuggan: I'd say in case a downstream contract wants to set specific limitations? It's not supposed to be able to work by itself anyway...

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/VirtualTokenSupplyBase.sol
0xNuggan: same as above

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
0xNuggan: Not sure about this one... It does create a buy order with the virtual supply, and not the actual one. I created the function to avoid code duplication. Of course we can also make _buyOrder virtual and override it, but to me this seems more explicit?

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
0xNuggan: Yep, forgot to remove that one

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
0xNuggan: same as above

test/e2e/E2eTest.sol
0xNuggan: Supercalifragilisticexpialidocious

test/modules/fundingManager/bondingCurveFundingManager/marketMaker/utils/mocks/BancorVirtualSupplyBondingCurveFundingManagerMock.sol
0xNuggan: Yeah, I know, but it always seems confusing to me... I prefer to instantiate mocks and test them. But if we choose to enforce a specific code standard I'm happy to oblige.

test/modules/fundingManager/bondingCurveFundingManager/RedeemingB_CurveFundingManagerBase.t.sol
0xNuggan: Yes. But apparently missed one statement to be controlled for...

test/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.t.sol
0xNuggan: Because we are taking the formula wholesale from Aragon, so we don't re-test it. Updated the wording

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
0xNuggan: maybe renaming to virtualBuyOrder is better ?

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
Zitzak: Do we already know when this change will be implemented? This would mean that until then the module would not be able to work with the current setup.

Also seeing that all `FundingManagers` will need to have a token specified, might it make sense to adapt the base `Module` and/or `IFundingmanager` contract/interface?

FHieser: IFundingManager has already the token() function implemented and can just be referenced via orchestrator.fundingManager.token() i think

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
FHieser: I was a bit confused, when i read it, because you combine the two of them together basically, but the name suggests that you use just the virtual

test/e2e/E2eTest.sol
FHieser: xD
At some point this probably isnt the only one that we use lol

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
Zitzak: I think it makes sense indeed.

The redeeming contract however also has FundingManagerBase in the title. Would you suggest renaming it? @FHieser

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
Zitzak: Will implement the IFundingManager interface as you suggested as well as the way to reference the token :) thanks

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/VirtualCollateralSupplyBase.sol
Zitzak: Could you comment what was meant by this @0xNuggan. I was unsure what to do here

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/VirtualCollateralSupplyBase.sol
Zitzak: The reason is that the implementing function needs to be restricted by a modifier, which cannot be accessed in this contract. For that reason this function is set to abstract.

I'll add a comment to clarify this

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/VirtualTokenSupplyBase.sol
Zitzak: Same as the other one :)

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
Zitzak: I think the idea is that the base contract has the `buyOrder` function of course. But because our implementation makes use of a virtual supply, we prepend the virtual supply part to the buyOrder.

I think overriding _buyOrder would defeat the purpose of our abstract formula wrappers

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
0xNuggan: I think since it inherits from bondingCurvefundingManagerBase (and therefore inherits IFundingManager) it's ok to keep the name as is

src/modules/fundingManager/bondingCurveFundingManager/marketMaker/VirtualCollateralSupplyBase.sol
0xNuggan: Old comment. I removed it now

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
FHieser: But arent you overriding it currently anyway?
I think for me it was more of a design thing anyway.
Because your using the original _buyOrder function anyway this doesnt make so much of a difference,
but I thought because you extend it anyway you could also write it the way I described it before

FHieser (approved):
Although Im giving the approve I would highly suggest we get another pair on eyes on this, because of the volume of this PR

Zitzak:
Agree @FHieser, I've added @saxenism to the reviewers as per this ticket :)
https://app.clickup.com/t/860rrm6c3

PR #349, opened 2023-09-27
Add extra E2E tests

### LINKED ISSUE #298: E2E Test RoleAuthorizer
There is still a e2e test missing here

### LINKED ISSUE #299: E2E Test TokenGatedRoleAuthorizer
There is still a e2e test missing here

### LINKED ISSUE #300: E2E Test SingleVoteGovernor
There is still a e2e test missing here

0xNuggan:
adds explicit E2E tests for:
- RoleAuthorizer ( closes #298 )
- TokenGatedRoleAuthorizer ( closes #299  )
- SingleVoteGovernor ( closes #300 )


test/e2e/RoleAuthorizerE2E.t.sol
```diff
@@ -0,0 +1,172 @@
...
+
+        // we authorize the bountySubmitter to submit bounty claims
+        bountyManager.grantModuleRole(
+            bountyManager.CLAIMANT_ROLE(), address(bountySubmitter)
+        );
+
+        // we grant manager role to managerAddress
+        bytes32 managerRole = authorizer.getManagerRole();
+        authorizer.grantRole(managerRole, address(orchestratorManager));
+        authorizer.renounceRole(managerRole, address(this));
```
FHieser, line 101: Maybe comment why you have to remove the managerRole from address(this)

test/e2e/RoleAuthorizerE2E.t.sol
```diff
@@ -0,0 +1,172 @@
...
+        assertTrue(authorizer.hasRole(managerRole, orchestratorManager));
+        assertEq(authorizer.getRoleMemberCount(managerRole), 1);
+
+        //we grant owner role to ownerAddress
+        bytes32 ownerRole = authorizer.getOwnerRole();
+        authorizer.grantRole(ownerRole, address(orchestratorOwner));
+        authorizer.renounceRole(ownerRole, address(this));
```
FHieser, line 108: see above

test/e2e/RoleAuthorizerE2E.t.sol
FHieser: Maybe even do the shuffle to the correct ownership earlier
I think having a clear view of who has access rights right at the start makes more sense

test/e2e/SingleVoteGovernorE2E.t.sol
```diff
@@ -105,22 +105,15 @@ contract SingleVoteGovernorLifecycle is E2eTest {
...
 
-        // Funders deposit funds
-
-        // IMPORTANT
-        // =========
-        // Due to how the underlying rebase mechanism works, it is necessary
-        // to always have some amount of tokens in the orchestrator.
-        // It's best, if the owner deposits them right after deployment.
+        //--------------------------------------------------------------------------------
+        // Set up seed deposit and initial deposit by users
```
FHieser: The initial deposit isnt necessary anymore is it?
I think we should remove the fundingmanager inclusion in this e2e test completely and maybe just focus on the role management side of the singlevote governor
Otherwise it might confuse potential users that the payment part may be strictly necessary although it isnt

You could just do a Role change instead or maybe change something via setter

test/e2e/SingleVoteGovernorE2E.t.sol
```diff
@@ -105,22 +105,15 @@ contract SingleVoteGovernorLifecycle is E2eTest {
 
         authorizer.renounceRole(ownerRole, address(this));
```
FHieser, line 107: Maybe explain which role the singlevote governor takes

test/e2e/TokenGatedRoleAuthorizerE2E.t.sol
```diff
@@ -0,0 +1,219 @@
...
+        vm.prank(orchestratorManager);
+        bountyManager.verifyClaim(claimId, contribs);
+
+        // the worker can't verifiy it
+        vm.expectRevert();
+        vm.prank(bountySubmitter);
+        bountyManager.addBounty(100e18, 500e18, "This is a test bounty");
+    }
+
+    function _createNewOrchestratorWithAllModules_withTokenGatedRoleAuthorizer(
```
FHieser, line 196: Lol this here is a great idea
Why didnt we use this in the other e2e tests?
Basically building the Orchestrator directly in here instead of the base e2e test
I always felt like it was pretty ugly to see in the base e2e test

FHieser (commented):
I think the changes I recommended would increase the readability of the e2e tests
any opinions @0xNuggan ?

test/e2e/RoleAuthorizerE2E.t.sol
0xNuggan: added a comment. We cannot renounce ownership before assigning manager role without using prank(), that's why I it's in that order

test/e2e/SingleVoteGovernorE2E.t.sol
0xNuggan: I think the initial deposit is still necessary for the rebasingFundingManger in general, but since in this case we aren't paying out anything it doesn't matter

test/e2e/SingleVoteGovernorE2E.t.sol
0xNuggan: done :+1:

test/e2e/TokenGatedRoleAuthorizerE2E.t.sol
0xNuggan: Yeah, maybe we should refactor the other e2e into that format

test/e2e/SingleVoteGovernorE2E.t.sol
0xNuggan: BTW it wasn't, I removed it

FHieser (approved):

Issue #308, opened 2023-08-01
Feat: setup bonding curve module contracts


Zitzak:
As **a dev**, **I want to** setup the contracts and testing, **so that** I can develop the features of the contract with the structure in place

## Acceptance Criteria
- [x] Bonding Curve Module contract setup
  - [x] Setup test contract for Bonding Curve Module contract
- [x] VirtualSupplyToken contract setup
  - [x] Setup test contract
- [x] MarketMaker contract setup 
  - [x] Setup test contract
- [x] BancorFormula and util contracts in repo

PR #355, opened 2023-10-13
Bonding curve extra scripts

0xNuggan:
Ties into #308 I guess, but is separate. This deploys an Investable Workstream in the type wanted by the BCRG


Zitzak:
What should we do with this issue PR @0xNuggan ? Maybe close it for now, as it has been open for a while

script/deployment/DeploymentScript.s.sol
```diff
@@ -83,13 +90,22 @@ contract DeploymentScript is Script {
         1, 1, "https://github.com/inverter/bounty-manager", "BountyManager"
     );
 
+    IModule.Metadata bondingCurveFundingManagerMetadata = IModule.Metadata(
+        1,
+        1,
+        "https://github.com/inverter/bonding-curve-funding-manager",
+        "BancorVirtualSupplyBondingCurveFundingManager"
```
Zitzak, line 97: Should this URL be accessible?  It not pointing anywhere atm

script/utils/InvestableWorkstreams/01-DeployInvestableWorkstream.s.sol
```diff
@@ -0,0 +1,273 @@
...
+    IBountyManager
+} from "src/modules/logicModule/BountyManager.sol";
+import {BancorVirtualSupplyBondingCurveFundingManager} from
+    "src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol";
+
+import {BancorFormula} from
+    "src/modules/fundingManager/bondingCurveFundingManager/formula/BancorFormula.sol";
+
+import {ERC20} from "@oz/token/ERC20/ERC20.sol";
+//import {ERC20Mock} from "test/utils/mocks/ERC20Mock.sol";
```
Zitzak: Not being used. Same for the line beneath it. Could be removed

script/utils/InvestableWorkstreams/01-DeployInvestableWorkstream.s.sol
```diff
@@ -0,0 +1,273 @@
...
+    uint orchestratorOwnerPrivateKey =
+        vm.envUint("ORCHESTRATOR_OWNER_PRIVATE_KEY");
+    address orchestratorOwner = vm.addr(orchestratorOwnerPrivateKey);
+
+    address collateralTokenAddress =
+        vm.envAddress("BONDING_CURVE_COLLATERAL_TOKEN");
+    ERC20 collateralToken = ERC20(collateralTokenAddress);
+
+    address bancorFormulaAddress = vm.envAddress("BANCOR_FORMULA_ADDRESS");
+    BancorFormula formula = BancorFormula(bancorFormulaAddress);
```
Zitzak, line 48: Curious if we store the address in the `.env` file manually or do we export them?

script/utils/InvestableWorkstreams/01-DeployInvestableWorkstream.s.sol
```diff
@@ -0,0 +1,273 @@
...
+            owner: orchestratorOwner,
+            token: collateralToken
+        });
+
+        // Funding Manager: Virtual Supply Bonding Curve Funding Manager
+        IOrchestratorFactory.ModuleConfig memory
+            bondingCurveFundingManagerConfig = IOrchestratorFactory
+                .ModuleConfig(
+                bondingCurveFundingManagerMetadata,
+                abi.encode(
```
Zitzak, line 133: The contract expect 2 objects and 1 token address, as well as the decimals of the issuance token. [See link](https://github.com/InverterNetwork/inverter-contracts/blob/dev/test/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.t.sol#L163)

script/utils/InvestableWorkstreams/01-DeployInvestableWorkstream.s.sol
```diff
@@ -0,0 +1,273 @@
...
+            paymentProcessorMetadata,
+            bytes(""),
+            abi.encode(hasDependency, dependencies)
+        );
+
+        // Authorizer: Metadata, initial authorized addresses
+        IOrchestratorFactory.ModuleConfig memory authorizerFactoryConfig =
+        IOrchestratorFactory.ModuleConfig(
+            authorizerMetadata,
+            abi.encode(orchestratorOwner, orchestratorOwner),
```
Zitzak, line 153: Are we expecting the Owner to have also the Manager role here?

script/utils/giveBountyManagerRolesToAddress.s.sol
```diff
@@ -0,0 +1,48 @@
...
+
+contract giveBountyManagerRoles is Script {
+    uint orchestratorOwnerPrivateKey =
+        vm.envUint("ORCHESTRATOR_OWNER_PRIVATE_KEY");
+    address orchestratorOwner = vm.addr(orchestratorOwnerPrivateKey);
+
+    // ===============================================================================================================
+    // Introduce corresponding bounty manager and user addresses here
+    // ===============================================================================================================
+    address bountyManagerAddress = address(0); // Introduce deployed bounty manager address here
```
Zitzak: Should this also be taken from the `.env` files as with all the other addresses? Or will the flow be such that who ever executes this script will do so while editing the script?

Zitzak: Same for the other addresses in this script

Zitzak (changes requested):
First time I reviewed/took a good look at the script. Think I caught most things but if we want to be more sure then we might want to request review from someone else as well
@0xNuggan

script/deployment/DeploymentScript.s.sol
0xNuggan: Yeah, we did some example links ages ago as placeholder. It's about time to change them to the real ones, I'll open an issue for it.

script/utils/InvestableWorkstreams/01-DeployInvestableWorkstream.s.sol
0xNuggan: Yes, that is correct

0xNuggan:
Thanks for the review! I've clarified the .env stuff. Basically it goes: If the address exists, it should be specified in the env file and used from there. If the contract doesn't exist, you can uncomment the corresponding part in the script to deploy it on the go.

Thanks for the heads up with the initial config. Not sure what happened there, since we deployed the curve already a couple times :thinking:

script/utils/InvestableWorkstreams/01-DeployInvestableWorkstream.s.sol
Zitzak: I don't think this works, as the BondingCurveFundingManager expects 3 different objects

Zitzak (changes requested):
There is still an issue with the way the init is done in the Bonding Curve Funding Manager

0xNuggan:
If you uncommented line 85 it worked though! Anyway, I changed it to send the objects, it's cleaner :)

Zitzak (approved):
LGTM

Zitzak:
Ohw interesting. Didn't expect that it would work. Do you know why that is? Doesn't it just expect objects? @0xNuggan

0xNuggan:
I think since it is a struct it just reads the storage in the expected chunks? Anyway, it's definitely better to be explicit :)

Issue #290, opened 2023-07-11
EIP165-based Module


marvinkruse:
We want to have a way to ensure that certain functions exist via EIP165.

0xNuggan:
Relevant from the security review: https://github.com/InverterNetwork/inverter-contracts/pull/350#discussion_r1355112732

FHieser:
Closed with #371

PR #371, opened 2023-10-26
EIP 165 implementation

saxenism:
Solves #290


src/modules/lib/LibInterfaceId.sol
```diff
@@ -0,0 +1,192 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
0xNuggan, line 1: What is the benefit of having a big file storing all the interfaceIDs? Isn't it easier to generate them on the fly since any contract checking them will know against what to compare? This way we avoid having to maintain a list, since we are going to be adding modules over time...

0xNuggan:
Also:
whereTests

src/orchestrator/Orchestrator.sol
```diff
@@ -300,6 +265,26 @@ contract Orchestrator is IOrchestrator, ModuleManager {
...
+        require(
+            _contractAddress.code.length != 0,
+            "Contract Address need to passed here"
+        );
+
+        (bool success, bytes memory data) = _contractAddress.call(
+            abi.encodeWithSignature("supportsInterface(bytes4)", _interfaceId)
+        );
+
+        if (success && abi.decode(data, (bool))) {
```
FHieser: Cant u just return success && abi.decode(data, (bool)) here?

test/e2e/SingleVoteGovernorE2E.t.sol
```diff
@@ -60,8 +60,8 @@ contract SingleVoteGovernorE2E is E2eTest {
             orchestratorConfig
         );
 
-        RebasingFundingManager fundingManager =
-            RebasingFundingManager(address(orchestrator.fundingManager()));
+        // RebasingFundingManager fundingManager =
```
FHieser: Outcommented

src/factories/ModuleFactory.sol
```diff
@@ -125,6 +126,18 @@ contract ModuleFactory is IModuleFactory, Ownable2Step {
...
     }
 
+    function supportsInterface(bytes4 interfaceId)
+        public
+        view
+        virtual
+        override(ERC165)
+        returns (bool)
+    {
+        bytes4 interfaceId_IModuleFactory = type(IModuleFactory).interfaceId;
```
FHieser: Cant you just do:
return interfaceId == type(IModuleFactory).interfaceId || ....

FHieser (commented):
Okay I have a comment that you might not like 🙈:
Can we move these to the start of the contract directly under the contract declaration?
I know the supportsInterface function is a public view function, but they dont have anything to do with the functionality of the contracts themselves but more with the interface implementation itself

src/factories/ModuleFactory.sol
saxenism: Done

src/modules/lib/LibInterfaceId.sol
saxenism: My thought process was regarding the late dependency injection usecase. When creating a new contract, you might get an address that either you can accept as a particular type of module or ask additionally for it's interface ID in the metadata, which you can compare against all the interface IDs of the inverter network by making a call to this library. Without this library, you'll probably have to import a particular type of module into the module that receives the metadata to calculate and then compare the interface IDs.

If this sounds like gibberish, happy to delete the library as well.

saxenism: Also, you'll see this being used sparsely in the Orchestrator.sol file.

saxenism:
> Also: whereTests

Added the unit tests.

src/factories/ModuleFactory.sol
```diff
@@ -32,7 +33,16 @@ import {
...
-contract ModuleFactory is IModuleFactory, Ownable2Step {
+contract ModuleFactory is IModuleFactory, Ownable2Step, ERC165 {
+    function supportsInterface(bytes4 interfaceId)
+        public
+        view
+        virtual
+        override(ERC165)
+        returns (bool)
+    {
+        return super.supportsInterface(interfaceId);
```
FHieser: Isnt here the IModuleFactory missing?

src/modules/fundingManager/RebasingFundingManager.sol
```diff
@@ -28,11 +28,22 @@ import {IFundingManager} from "src/modules/fundingManager/IFundingManager.sol";
 
 contract RebasingFundingManager is
     IFundingManager,
-    IRebasingERC20,
```
FHieser, line 31: Why was this removed?

src/modules/base/Module.sol
```diff
@@ -31,7 +32,24 @@ import {IAuthorizer} from "src/modules/authorizer/IAuthorizer.sol";
...
+{
+    function supportsInterface(bytes4 interfaceId)
+        public
+        view
+        virtual
+        override(ERC165)
+        returns (bool)
+    {
+        bytes4 interfaceId_IModule = type(IModule).interfaceId;
+        return interfaceId == interfaceId_IModule
```
FHieser: Why isn this just return interfaceId == type(IModule).interfaceId

src/orchestrator/Orchestrator.sol
FHieser, line 1: didnt we want to include a way to find a module here?

src/modules/fundingManager/bondingCurveFundingManager/formula/BancorFormula.sol
```diff
@@ -9,12 +13,23 @@ import {Utils} from
...
-///         2. Updating the Solidity version
-///         3. The version-specific removal of the 'public' keyword from the constructor
-///         4. The adjustments in the contract import methodology
-contract BancorFormula is IBancorFormula, Utils {
+/// @notice The sole modification applied to these contracts involves the alteration of the Solidity
+///         version and the version-specific removal of the 'public' keyword from the constructor,
+///         coinciding with adjustments in the contract import methodology.
+contract BancorFormula is IBancorFormula, Utils, ERC165 {
+
+    function supportsInterface(bytes4 interfaceId)
```
0xNuggan: I would remove this since we didn't want to touch the Formula file.

src/modules/lib/LibInterfaceId.sol
0xNuggan: Hmm I see. Seems like overhead to me, what do you think, @FHieser ? Also, I think I'd remove the use in the Orchestrator, since it will only validate the core interface types.

src/orchestrator/Orchestrator.sol
FHieser: The Problem I had with this after looking at it again, was that there was a possibility of multiple modules supporting a Interface. This would mean for finding all modules with a specific interface we would have to create a dynamic array of module addresses, as we dont know how many there really are. My initial thought with this comment was that it would be nice to have nice and easy way to access modules that way, but simulating a dynamic array is quite frankly a bit of a pain.
 As we currently already have a way to search for a module via name I chose not to implement this

FHieser (approved):
download

As the original Implementation was done by Rahul and I just did some finishing touches I will approve this myself xD

PR #382, opened 2023-11-07
344 staking contract v2

### LINKED ISSUE #344: Staking Contract
For the upcoming oStake Module we will need to implement a staking Contract.
The contract holds balances and allows for deposit/withdrawal (also of the depositFor() kind).

Possibly the easiest implementation for the staking use case would be to structure it as an [ERC20Snapshot](https://docs.openzeppelin.com/contracts/3.x/api/token/erc20#ERC20Snapshot) that gives out non-transferable receipt tokens (can be thought as transferable for future). This would allow us to query user balance based on a timestamp, which we can then use for reward distribution in other places.

FHieser:
Solves #344


0xNuggan:
Everything looks good, but I think it would be good if you added descriptions to the storage variables

src/modules/logicModule/StakingManager.sol
```diff
@@ -0,0 +1,271 @@
...
+
+    modifier validDuration(uint duration) {
+        if (duration == 0) {
+            revert Module__StakingManager__InvalidDuration();
+        }
+        _;
+    }
+
+    //--------------------------------------------------------------------------
+    // Storage
```
0xNuggan, line 44: Maybe add some descriptions here

0xNuggan (changes requested):
Everything looks good, but I think it would be good if you added descriptions to the storage variables

src/modules/logicModule/StakingManager.sol
FHieser: Done

0xNuggan (approved):

PR #350, opened 2023-10-06
[REVIEW] Feat: BondingCurveFundingManager + Fixes + CleanUp

marvinkruse:
⚠️ **This PR merges from `dev` into the `main` branch.** ⚠️
_At the time of creating this PR (until it is merged), the `main` branch contains the latest state that has received a security review. Anything in `dev` has been reviewed internally, but not received a security review._
<hr>

➡️ **Process**
This PR will receive an external security review. Additionally, at least one of our developers will review it as well, ideally someone who didn't work on it. Please raise any findings/questions directly within this PR, either as a code comment, referencing the corresponding line or as a general comment to the discussion. Please mark comments that are resolved as resolved.
<hr>

📋 **Contents**
This PR contains all of the changes of the current `dev` branch, which is mainly the BondingCurveFundingManager (+ corresponding contracts), fixes to prior security reviews and general cleanups and fixes.

📄 Notes from BondingCurve PRs ([PR1](https://github.com/InverterNetwork/inverter-contracts/pull/323) and [PR2](https://github.com/InverterNetwork/inverter-contracts/pull/342))
> - Most base contract (BondingCurveFundingManagerBase) implements functions concerning the issuing (buying) of tokens, which every bonding curve funding manager must have. It also defines an abstract function _issueTokensFormulaWrapper which must be used in downstream contracts to define through which formula price is calculated
> - The next base contract (RedeemingBondingCurveFundingMangerBase) implements the functions concerning the redemption of tokens back for their collateral. It also defines an abstract function _redeemTokensFormulaWrapper which must be used in downstream contracts to define through which formula price is calculated. The reason to have two of these functions is because the Bancor contract has a different interface for calculating issuing tokens vs redeeming tokens
> - There are 2 more abstract contracts which introduce a virtual issuing token supply (VirtualTokenSupplyBase) and a virtual collateral supply (VirtualCollateralSupplyBase). The would be considered even more specialised in terms of functionality and therefor would be added more in a compositing fashion instead of strict inheritence
> - The final contract has all the features above and is called BancorVirtualSupplyBondingCurveFundingManger

PR #385, opened 2023-11-09
Omega security review #350 : Remaining bonding curve comments

### LINKED ISSUE #380: Security Review #350 : Store decimals in state to limit external calls
### This issue is linked to

- Security Review with PR #350 
### Explanation

To minimize external calls and save gas, we should store the `decimals()` value within `BancorVirtualSupplyBondingCurveFundingManager`

### Category:
- Gas Optimizations

### Link to review comments

- [x] https://github.com/InverterNetwork/inverter-contracts/pull/350#discussion_r1360617517

### LINKED ISSUE #381: Security Review #350 : Exclude BC fee from balances
### This issue is linked to

- Security Review with PR #350 

### Explanation

The current implementation does not exclude the collected fee from the balances. This behaviour results in the balances not tracking the real in and outflow of collateral/issuance tokens, creating unwanted behaviour. The decision was made to exclude the fee from the virtual balances as they get updated, as well as track the fee in a separate state within the implementation contract

### Category:

- Low severity / Information issue

### Link to review comments

- [x] https://github.com/InverterNetwork/inverter-contracts/pull/350#discussion_r1360643084

Zitzak:
# What has been done? (Issue descriptions)
## Add static price getter
**Why:**
- The UI will need to display a price for user which represent a static price for a single unit.
**How:**
- As we are using Aragons Bancor formula, we decided to also use the calculating of the static price from the Aragon contracts
**What:**
- A function got added which returns the price formatted in PPM
- Tests got added and adapted to accommodate this change

## Add slippage to buy/sell functions
**Why:**
- To protect against txs being frontrun and general situation where a user gets back less tokens than expected because someone bought tokens before them
**How:**
- Functions got added that the UI should call to get the expected amount out before executing the tx. This amount can then be adjusted as the user sees fit to define the slippage. During the tx, this amount gets compared with the actual amount out, and if lower the tx reverts
**What:**
- Two functions got added to calculate the expected amount out:
  - `calculatePurchaseReturn`
  - `calculateSaleReturn`
- To all buy and sell functionalities I added a parameter called `minAmountOut`, which defines the min expected amount out
- A conditional check got added in `_buyOrder` & `_sellOrder` which reverts when the actual amount < `minAmountOu`
- Tests got added and adapted to accommodate this change
# Implemented Issues
- closes #381
- closes #380


0xNuggan (approved):
LGTM!

test/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.t.sol
```diff
@@ -1129,14 +1154,62 @@ contract BancorVirtualSupplyBondingCurveFundingManagerTest is ModuleTest {
         assertEq(bondingCurveFundingManager.decimals(), _newDecimals);
     }
 
+    /* Test _staticPricePPM function
+        ├── When calling with reserve ratio for selling
+        |   └── it should return the right value
+        ├── When calling with reserve ratio for buying
+        |   └── it should return the right value
+    */
+    function testStaticPriceWithSellReserveRatio() public {
```
Zitzak: @0xNuggan for the tests of the staticPrice calculation functions  I wanted to ask your input, as I mentioned in the standup.

Do you think we should test the static price formula taken from the Aragon contract somehow, or can we just take it from Aragon's contracts and trust that it is correct. We do a similar thing with the Bancor Formula contract of course.

I tried testing it through the Bancor Formula contract but I was not able to get a return amount that is comparable

Zitzak:
I've added descriptions of what I've done in the PR description instead of opening two new issues. Jfyi @0xNuggan @FHieser

test/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.t.sol
0xNuggan: Yes, that's a good question. Trusting Aragon to be right is not that much of  stretch (and we're doing it with the BacnorFormula anyway)...
Another option would be to perform a very small buy (maybe the smallest valid buy possible) and verify that:
- The static price has risen.
- It has risen less than a certain threshold (no crazy spikes)
- The caller has received a non-zero amount of tokens out
And then do fuzzing on the virtualCollateralSupply, virtualTokenSupply and reserveRatio to make sure it holds along the whole curve

Zitzak: I think this is a good idea, will test for something along those lines. Thanks

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -216,6 +227,71 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
...
+    }
+
+    /// @inheritdoc IBancorVirtualSupplyBondingCurveFundingManager
+    function calculateSaleReturn(uint _depositAmount)
+        external
+        view
+        returns (uint redeemAmount)
+    {
+        if (_depositAmount == 0) {
+            revert
```
FHieser, line 299: Should a purely informative function really fail here?
Couldnt you just return 0 here?
This also counts for ->calculatePurchaseReturn

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -303,6 +377,25 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
...
+    /// based on the provided issuance token supply, collateral supply, and buy or sell reserve ratio.
+    /// Note: The reserve ratio specifies whether the sell or buy price is returned.
+    /// The formula used is: PPM * PPM * collateralSupply / (issuanceTokenSupply * reserveRatio).
+    /// The formula is based on Aragon's BatchedBancorMarketMaker, which can be found here:
+    /// https://github.com/AragonBlack/fundraising/blob/5ad1332955bab9d36cfad345ae92b7ad7dc0bdbe/apps/batched-bancor-market-maker/contracts/BatchedBancorMarketMaker.sol#L415
+    /// @param _issuanceTokenSupply The total supply of the issuance tokens.
+    /// @param _collateralSupply The total supply of the collateral held by the FundingManager.
+    /// @param _reserveRatio The reserve ratio, specified as either sell or buy reserve ratio.
+    /// @return uint The calculated static price for the specified operation, formatted in PPM
+    function _staticPricePPM(
```
FHieser, line 406: Just for my understanding:
What exatcly is the static Price?
Shouldnt this be linked in some form to the BancorFormular, as the price changes based on demand?

src/modules/fundingManager/bondingCurveFundingManager/RedeemingBondingCurveFundingManagerBase.sol
```diff
@@ -108,15 +118,18 @@ abstract contract RedeemingBondingCurveFundingManagerBase is
...
-    function _sellOrder(address _receiver, uint _depositAmount)
-        internal
-        returns (uint redeemAmount)
-    {
+    /// @return feeAmount The amount of collateral token subtracted as fee
+    function _sellOrder(
+        address _receiver,
+        uint _depositAmount,
+        uint _minAmountOut
+    ) internal returns (uint redeemAmount, uint feeAmount) {
```
FHieser, line 144: Again for my understanding:
Do we take a fee from users for selling and buying in this module?

FHieser (approved):

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
0xNuggan: static price is basically "where the needle is at" in the bonding curve. Any actual amount bought will move it up (so the final price will be different), but it kind of represents where the price starts moving from

src/modules/fundingManager/bondingCurveFundingManager/RedeemingBondingCurveFundingManagerBase.sol
0xNuggan: Yes, as it is right now we do. This could be something to think about in relation to our fee structure. In an hypothetical V2, we could have the "sell" function create payment orders, and thus go through our regular fee system.

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
0xNuggan: If we don't, it will revert when calling the formula anyway in line 287. This way we can give an error message

0xNuggan (approved):

PR #392, opened 2023-11-29
376 Create E2E Test for MetadataManager

### LINKED ISSUE #376: Should we remove the MetadataManager and SingleVoteGovernor contracts?
We aren't actually using the MetadataManager (not even in the ModuleManager!), and I'm not sure what their goal was in the first place. My opinion is we should drop the MetadataManager completely, but I'd love hear your thoughts @marvinkruse @FHieser @saxenism  @Zitzak 

*Please note:* The MetadataManager is a separate contract from the regular LibMetadata for the modules

**UPDATE**: After [this](https://github.com/InverterNetwork/inverter-contracts/pull/350#discussion_r1382612000) comment, I'm including the SingleVoteGovernor in this. The reason is that there exist external Governor implementations that are audited and battle-tested. Since many of the conditions under which the SingleVoteGovernor was developed no longer apply, it may be smarter to remove our implementation and use one of those (implementing a small wrapper if necessary)

FHieser:
Solves #376


0xNuggan (commented):
Looks good to me! Quite the team :monocle_face:

0xNuggan (approved):

PR #397, opened 2023-12-13
Static Price Test

Zitzak:
# testStaticPriceWithBuyReserveRatio
The test `testStaticPriceWithBuyReserveRatio()` works as it currently is. It tests if the return value of `staticPricePPM()` increases after a buy of amount 1.

## Questions
- When testing the buyer balance after the buy, it returns back `0`. I'm guessing the reason for that is because maybe the buy amount was to little? Would you just manually increase it until we get a return value? Within your comment you've mentioned "smallest valid possible buy"
- And just to confirm, after I find the minimal amount, then you would add the virtual supplies and reserve ratio to the parameters to do fuzzing, while keeping the same minimal buy amount to test with right?
# testStaticPriceWithSellReserveRatio
I don't get the test `testStaticPriceWithBuyReserveRatio()` to work atm

## Questions
- When selling a single token, the `staticPricePPM` does not change at all. I'm guessing it's the same reason as above


FHieser (approved):

Zitzak:
@0xNuggan will you give this PR another review as well? You have of course seen it during development but might be good

0xNuggan:
Looks good to me. Merging now

PR #399, opened 2023-12-19
394 metatransactions dev

FHieser:
Solves #393


src/factories/ModuleFactory.sol
```diff
@@ -1,9 +1,11 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 pragma solidity 0.8.19;
 
-// External Dependencies
-import {Context} from "@oz/utils/Context.sol";
+//External Dependencies
+import {ERC2771Context} from "@oz/metatx/ERC2771Context.sol";
 import {Ownable2Step} from "@oz/access/Ownable2Step.sol";
+import {Ownable} from "@oz/access/Ownable.sol";
```
Zitzak: It seems like `Ownable` is not used anywhere

script/deployment/DeploymentScript.s.sol
```diff
@@ -90,8 +90,9 @@ contract DeploymentScript is Script {
         authorizer = deployRoleAuthorizer.run();
 
         moduleFactory = deployModuleFactory.run();
-        orchestratorFactory =
-            deployOrchestratorFactory.run(orchestrator, moduleFactory);
+        orchestratorFactory = deployOrchestratorFactory.run(
+            orchestrator, moduleFactory, address(0)
```
Zitzak: Still has @todo comment, as well as passing `address(0)` as forwarder. Assuming the Forwarder address comes from the `.env` file in the scripts?

src/factories/ModuleFactory.sol
```diff
@@ -1,9 +1,11 @@
...
+import {Context} from "@oz/utils/Context.sol";
```
Zitzak: The `ERC2771Context` already inherits from `Context`. Any particular reason why you left the import, because you removed `ContextUpgradable` within the `ModuleManager` when importing `ERC2771Context`.

src/factories/OrchestratorFactory.sol
```diff
@@ -61,7 +64,11 @@ contract OrchestratorFactory is IOrchestratorFactory {
     //--------------------------------------------------------------------------
     // Constructor
 
-    constructor(address target_, address moduleFactory_) {
+    constructor(
+        address target_,
+        address moduleFactory_,
+        address _trustedForwarder //@todo should this be changeable?
```
Zitzak, line 82: Todo still left in code.

And for my own understanding, the question here is if we want to be able to change the address of the Forwarder after factory deployment right? My 2 cents here would be that it makes sense to be able to edit it because it doesn't impose a centralization risk like with the `moduleFactory`. I'm guessing that no one cares who will pay their fees, and if that changes over time, as long as it works :D

Zitzak:
One other thing to note is that is that it seems OZ stepped away from the `MinimalForwarder` in their [newer library version](https://github.com/OpenZeppelin/openzeppelin-contracts/tree/master/contracts/metatx). Not sure if this is important for us, if we going to implement something other than for the tests, but in [Marvin K's PR](https://github.com/InverterNetwork/inverter-contracts/pull/393), he stated as todo: "Finalize MinimalForwarder to be ready" which is why I wanted to bring this up

src/factories/ModuleFactory.sol
FHieser: Yeah I removed it :+1:

src/factories/ModuleFactory.sol
FHieser: Because the Factory is inheriting from Ownable2Step and ERC2771Context on two seperate occasions I have to explicitly override them for the implementation of the ERC2771. For that I need the reference.
In the fix I shifted it so that it reads the import from Ownable2Step specifically maybe that makes it more readable.

script/deployment/DeploymentScript.s.sol
FHieser: Yeah seems I overlooked that one.
For now yes. For the multicall implementation I will probably change the Forwarder anyway, so this more of a temporary solution

src/factories/OrchestratorFactory.sol
FHieser: Yeah I wanted to ask that the reviewees ^^
"as long as it works" is the key argument here. Because the Forwarder basically circumvents the Authorization, by customizing the msg.sender we have to be careful here. I think this directly ties into the topic of safe module upgrades.
Also: Who does the updating of the forwarder address? If its each own orchestrator this has to be done by each owner seperately. A beacon system might make sense here.

FHieser:
> One other thing to note is that is that it seems OZ stepped away from the `MinimalForwarder` in their [newer library version](https://github.com/OpenZeppelin/openzeppelin-contracts/tree/master/contracts/metatx). Not sure if this is important for us, if we going to implement something other than for the tests, but in [Marvin K's PR](https://github.com/InverterNetwork/inverter-contracts/pull/393), he stated as todo: "Finalize MinimalForwarder to be ready" which is why I wanted to bring this up

Yeah also in the current Version of the minimal Forwarder it says:
MinimalForwarder is mainly meant for testing, as it is missing features to be a good production-ready forwarder. This contract does not intend to have all the properties that are needed for a sound forwarding system. A fully functioning forwarding system with good properties requires more complexity. We suggest you look at other projects such as the GSN which do have the goal of building a system like that.

I will pick this up in the multicall implementation.

PR #401, opened 2024-01-11
OpenZeppelin V5 update

Zitzak:
## What has been done?
- Updated paths
- Pass explicit owner to Ownable contract
- Adapt `OZErrors.sol` and tests using it do work with custom errors

## Notes for the review:
- In the context of Ownable, the owner needs to be explicitly set. I chose to go for the msg.sender as it reflect what we expect to happen with the v4 version of OZ. We might opt for passing an address explicitly if we want to follow their pattern
- `Address.isContract` has been removed, which was used in `Beacon.sol`. Reason being its misleading nature, which you can read more about within [their PR](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3945). To replace it, I went with implementing the logic that the function used to implement. We might want to update this in the future for a better check


FHieser (commented):
All in all looks good.
Did we agree on updating to pragma solidity 0.8.20 ? @Zitzak

Zitzak:
No we didn't, but the OZ contract need to be compiled with `^0.8.20` so I had to adjust our contracts. @FHieser

FHieser (approved):

marvinkruse:
I think upgrading the Solidity version makes sense, as we wanted to do it anyway at some point. It makes sense to just do it in one go now, imho.

One brief discussion maybe to have now is whether we want to go with a higher version directly. Ideally, for launch, we would like to have a version that has been out for a few weeks/months so the risk of undetected bugs in that version is as low as possible, but not a version that is too old. Even the latest version `0.8.23` has been out since November of 2023. Version `0.8.21` and `0.8.22` added some important bug fixes and little gas improvements like "Unchecked loop increments", etc. 
So maybe let's go with `0.8.23` directly? Even for our testnet launch in March, this solc version would be out for 4-5 months by then, which sounds fine to me.

What do you think, @0xNuggan, @FHieser and @Zitzak?

marvinkruse (approved):
can't find any issues, looks good to me. besides the CI issue of course and potential solc version change 😛

Zitzak:
Updating to `0.8.23` sounds good to me. If the others agree as well then I will push those changes as well

Zitzak:
@marvinkruse, talked with @FHieser and he agreed with your suggestion of updating the version. It is included in the last commit

0xNuggan:
Sounds good to me! Let's go with `0.8.23`

PR #403, opened 2024-01-17
402 remove teststaticpricewithfuzzing

### LINKED ISSUE #402: Remove testStaticPriceWithFuzzing
Context
The test for the function staticPrice() sometimes fails or reverts when fusing fuzzed valued. The reason for this is because the values that are being tested are outside of the possible configuration space of the bonding curve.
@Zitzak  thought he had them limited to a (somewhat) possible realm, but because this has been manual trial and error, and the fuzzing touches unpredictable values, this turns out not to have worked.

Options Forward
- Manual test to find the possible configuration space and limit the static price test accordingly
- Remove the test for the security review PR

Decision
We decided to move forward without the fuzzing test for now because of the following reasoning
- The static price calculation has been tested and works as predicted. The fuzzing test should have tested values along the whole curve but this turns out to be difficult because of the 4 values interplaying for the formula, see next point.
- The real issue this reveals is one we already discussed in a standup, namely that the possible bonding curve parameters are not being limited to the actual workable once. Fuzzing the values in the tests reveals this issue again. We decided that the UI would simulate setting of values and TXs such that it would reject values which do not function. So the issue here is not the specific test but that we don't know they actual boundaries of the math behind the formula
- Lastly, manually trying to figure out the actual configuration space looks like could take quite a while and would block the security review unnecessarily

FHieser:
Solves #402


marvinkruse (approved):
LGTM!

PR #404, opened 2024-01-23
Additional events

0xNuggan:
This adds all necessary events and tests for their emission


src/modules/fundingManager/IFundingManager.sol
```diff
@@ -21,25 +21,25 @@ interface IFundingManager {
     // Events
 
     /// @notice Event emitted when a deposit takes place.
-    /// @param from The address depositing tokens.
-    /// @param to The address that will receive the receipt tokens.
-    /// @param amount The amount of tokens deposited.
+    /// @param _from The address depositing tokens.
```
FHieser, line 24: This is the only place you put an _ before the parameters. Why?

src/modules/fundingManager/bondingCurveFundingManager/RedeemingBondingCurveFundingManagerBase.sol
```diff
@@ -206,8 +206,11 @@ abstract contract RedeemingBondingCurveFundingManagerBase is
         if (_fee > BPS) {
             revert RedeemingBondingCurveFundingManager__InvalidFeePercentage();
         }
-        emit SellFeeUpdated(_fee, sellFee);
+        uint oldSellFee = sellFee;
```
FHieser: Why did you change that?

src/modules/paymentProcessor/StreamingPaymentProcessor.sol
```diff
@@ -518,6 +518,8 @@ contract StreamingPaymentProcessor is Module, IStreamingPaymentProcessor {
             .length - 1];
 
         activeVestingWallets[client][paymentReceiver].pop();
+
+        //TODO check if event is missing?
```
FHieser: TODO

FHieser (approved):
LGTM
I didnt find anything mayor.

src/modules/fundingManager/bondingCurveFundingManager/RedeemingBondingCurveFundingManagerBase.sol
0xNuggan: The event was misplaced on my branch and returning the same number twice. I agree emitting before is the more elegant solution. fixed

src/modules/fundingManager/IFundingManager.sol
0xNuggan: because "for" is a protected keyword, but more intuitive than "to"

src/modules/paymentProcessor/StreamingPaymentProcessor.sol
0xNuggan: Fixed. The event gets emitted in _afterClaimCleanup()

Issue #394, opened 2023-11-30
Metatransactions


FHieser:
Implement the [ERC2771](https://eips.ethereum.org/EIPS/eip-2771) to enable Metatransactions

Issue #206, opened 2023-03-30
Multicall Implementation


FHieser:
It makes sense for some of the modules to support multicall functionalities to be able to call multiple functions in the same transaction.

https://docs.openzeppelin.com/contracts/4.x/api/utils#Multicall

PR #405, opened 2024-01-24
SC-122: Implement MultiCall and MetaTransaction Support

### LINKED ISSUE #395: [SC-56] Implement Multicall Support
<sub>From [SyncLinear.com](https://synclinear.com) | [SC-56](https://linear.app/inverter/issue/SC-56/implement-multicall-support)</sub>

FHieser:
Solves #206 and #394


FHieser:
I for now converted it to a draft, because the target of the PR will be the dev branch, but we have to wait until the openzeppelin update version is through

src/factories/OrchestratorFactory.sol
```diff
@@ -1,5 +1,8 @@
 // SPDX-License-Identifier: LGPL-3.0-only
-pragma solidity 0.8.19;
+pragma solidity 0.8.20;
```
Zitzak: Versions needs to be updated to `0.8.23` to align with the decision in the OZ update PR. Needs to happen with all the src files

src/factories/ModuleFactory.sol
```diff
@@ -1,9 +1,10 @@
 // SPDX-License-Identifier: LGPL-3.0-only
-pragma solidity 0.8.19;
+pragma solidity 0.8.20;
```
Zitzak, line 2: Update version

src/factories/OrchestratorFactory.sol
```diff
@@ -73,7 +76,11 @@ contract OrchestratorFactory is IOrchestratorFactory, ERC165 {
     //--------------------------------------------------------------------------
     // Constructor
 
-    constructor(address target_, address moduleFactory_) {
+    constructor(
+        address target_,
+        address moduleFactory_,
+        address _trustedForwarder //@todo should this be changeable?
```
Zitzak, line 82: Assuming this still needs resolving? Could be part of Fridays high-level call?

src/factories/OrchestratorFactory.sol
FHieser: Because this is based on the openzeppelin PR #401,  I will just rebase to the newer version.
Also because this is merging into the dev branch all the changes to the openzepplin are still marked here although they are not relevant to the PR

src/modules/base/IModule.sol
```diff
@@ -87,4 +87,17 @@ interface IModule {
     function grantModuleRole(bytes32 role, address addr) external;
 
     function revokeModuleRole(bytes32 role, address addr) external;
+
+    //--------------------------------------------------------------------------
+    // ERC2771 Context Upgradeable
+
+    /// @notice Checks if the provided address is the trusted forwarder
+    /// @param forwarder The contract address to be verified.
+    /// @return Is the given address the trusted forwarder
```
Zitzak:
```suggestion
    /// @return bool Is the given address the trusted forwarder
```

Makefile
```diff
@@ -78,20 +78,24 @@ testScripts: ## Run e2e test suite
 
 	@forge script script/factories/DeployModuleFactory.s.sol
 	@forge script script/factories/DeployOrchestratorFactory.s.sol
+	@forge script script/factories/DeployOrchestratorFactory.s.sol
```
Zitzak: Duplicate, prob because of merge

Makefile
```diff
@@ -78,20 +78,24 @@ testScripts: ## Run e2e test suite
...
 
 	@forge script script/modules/governance/DeployRoleAuthorizer.s.sol
 	@forge script script/modules/DeploySingleVoteGovernor.s.sol
-	
+
 	@forge script script/modules/paymentProcessor/DeploySimplePaymentProcessor.s.sol
 	@forge script script/modules/paymentProcessor/DeployStreamingPaymentProcessor.s.sol
 
+	@forge script script/modules/fundingManager/DeployRebasingFundingManager.s.sol
```
Zitzak: Duplicate

Makefile
```diff
@@ -78,20 +78,24 @@ testScripts: ## Run e2e test suite
...
 	@forge script script/modules/fundingManager/DeployRebasingFundingManager.s.sol
 
+	@forge script script/orchestrator/DeployOrchestrator.s.sol
```
Zitzak: Duplicate

Makefile
```diff
@@ -78,20 +78,24 @@ testScripts: ## Run e2e test suite
...
 	@forge script script/orchestrator/DeployOrchestrator.s.sol
 
 	@forge script script/proxies/DeployBeacon.s.sol
 
 	@forge script script/setup/SetupToyOrchestratorScript.s.sol
```
Zitzak: Duplicate

src/external/forwarder/TransactionForwarder.sol
```diff
@@ -0,0 +1,101 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity 0.8.20;
```
Zitzak: Needs updating to newer version. Mentioning it because this contract is not in the OZ branch

src/external/forwarder/TransactionForwarder.sol
```diff
@@ -0,0 +1,101 @@
...
+            if (!success && !calli.allowFailure) {
+                revert CallFailed(calli);
+            }
+
+            //set result correctly
+            results[i] = Result(success, returnData);
+        }
+    }
+
+    // Copied from the ERC2771Forwarder as it isnt declared internal ಠ_ಠ
```
0xNuggan, line 64: That's a weird choice by OZ

FHieser: yeah totally agree

Zitzak:
LGTM, Only thing left is the question about the changeable forwarder. Also would like to mention that I loved reading your piece about the technical side of the ERC2771Multicall, which really helped me to review the code :))

0xNuggan:
Looks good to me. I only have one opinion:

If I understand everything right, the trusted forwarder can basically impersonate any address and the rest of the system will believe it. Since this is a massive trust assumption, I'm not sure the forwarder itself should be an upgradeable proxy. Maybe keeping it immutable, and just allowing a contract to change their forwarder if needed is better?

Also (and this is 99% me just not completely understanding the whole thing correctly): Where does the signature validation happen for the multicall? I was looking for some call to ECDSARecover or OZ's 2771 `verify()` function...
I can see it in the test for the normal metaTx (MetaTxAndMulticallE2E.t.sol, line 146), but where does the multicall check that the received transactions are legit?

FHieser:
@0xNuggan
Forwarder Upgradeability: yeah we had a discussion around this point internally. The point why we decided to make it upgradeable was the following: The user already has to make a trust assumption, when using our contracts anyway. With them being upgradeable we could just upgrade to a storage manipulator contract, that basically does the same as the tainted forwarder. 
So it really doesnt matter if we make the contract upgradeable or not. Because of that we wanted to at least make use of the beacon intervention mechanism to keep a safety check in the contract.

Muticall: You actually dont need signature here :D 
It basically uses the structure that the forwarder established initially to forward the calls and interpret the sender to not be the forwarder but instead it takes the original msg sender. Take a look at the last chapter of the article that I wrote. I think that should clear things up.

Issue #409, opened 2024-02-08
Security Review PR#350: Interface (Lib & ERC165Checker)


FHieser:
https://linear.app/inverter/issue/SC-148/security-review-pr350-interface-lib-and-erc165checker

Link to comment(s)

Link PR comment, grouping them into one addressable issue if desired

https://github.com/InverterNetwork/inverter-contracts/pull/350#discussion_r1474555550

https://github.com/InverterNetwork/inverter-contracts/pull/350#discussion_r1479614870

Acceptance Criteria

Responded to comment with the purpose to signal agreement/rejection or discuss option moving forward

Implemented changes if applicable

Zitzak:
Gm @FHieser , curious why you are not using the issue on Linear, or linking to it in the PR but instead duplicated it on Github

FHieser:
Because I did forgot how to set up the connection properly in linear.
When I asked MarvinK about the other Issue that I tackled afterwards there was something wrong with the bridge to github and Marvin set it up for me, but afterwards it failed for me again for the 3rd issue.
I think it would help me if there was a "How to setup Issue thats linked to GitHub for Dummies" page that explains everything.

PR #410, opened 2024-02-08
SC-148: Security Review - Interface lib ERC165Checker

FHieser:
Solves #409


Zitzak (dismissed):
LGTM

Zitzak (commented):
I did approve, assuming that the test that was failing was prob for the interfaces, that we fixed in the `temp_dev` branch. It seems however that a different test is failing @FHieser. The code itself looks good, but as we agreed, I should only approve when I'm okay with it being merged, which is why I removed the approval.

Update: [This fix](https://github.com/InverterNetwork/inverter-contracts/pull/414/commits/39578b4cc2a4eb40befef54f86cd838abd2719a4) should solve the CI failure related to the address collision in the BondingCurve contract

FHieser:
Should I implement this fix in this branch? @Zitzak

Zitzak:
Thinking about it, I think it's fine. As long as it's eventually in the `dev` branch, and works before merging into `main`
@FHieser

Zitzak (approved):
LGTM. The failing BondingCurve test is addresses in PR#414, which is why I'm approving despite failing CI

PR #418, opened 2024-02-14
SC-140: Implement version and intervention mechanism in beacon

FHieser:
The Beacon needs to return the Major and Minor version of a module
I also added the intervention mechanism into this PR, because they touch the same subject


src/factories/ModuleFactory.sol
```diff
@@ -7,11 +7,13 @@ import {Ownable2Step} from "@oz/access/Ownable2Step.sol";
 import {Ownable} from "@oz/access/Ownable.sol";
 
 // External Interfaces
-import {IBeacon} from "@oz/proxy/beacon/IBeacon.sol";
+
```
marvinkruse: Pedantic-Mode Activated: There is an extra line here now :D

src/factories/ModuleFactory.sol
```diff
@@ -28,8 +30,8 @@ import {
  *
  * @dev An owned factory for deploying modules.
  *
- *      The owner can register module metadata's to an {IBeacon}
- *      implementations. Note that a metadata's registered {IBeacon}
+ *      The owner can register module metadata's to an {IInverterBeacon}
+ *      implementations. Note that a metadata's registered {IInverterBeacon}
```
marvinkruse, line 33: Pedantic-Mode still active: Wouldn't it make more sense to refer to it as `InverterBeacon` here? As users are not interacting **with** the interface but just **through** it. You register the metadata to an InverterBeacon, not it's interface. Might just be me though :D

Or are you calling it `IInverterBeacon` here because you refer to an instance of the interface?

src/factories/beacon/InverterBeacon.sol
```diff
@@ -0,0 +1,150 @@
...
+        minorVersion = newMinorVersion;
+
+        emit Upgraded(newImplementation, newMinorVersion);
+    }
+
+    //--------------------------------------------------------------------------------
+    // onlyOwner Intervention Mechanism
+
+    /// @inheritdoc IInverterBeacon
+    function shutdownImplementation() external {
```
marvinkruse: Pedantic-Mode still active: As you "shut down" the implementation and not "shutdown"; shouldn't this be camel-cased as `shutDownImplementation` instead of `shutdownImplementation`?

src/factories/beacon/InverterBeacon.sol
```diff
@@ -0,0 +1,150 @@
...
+    // State
+
+    /// @dev The beacon's implementation address.
+    address private _implementation;
+
+    /// @dev The beacon's current implementation pointer.
+    address private _currentImplementation;
+
+    /// @dev Is the beacon shut down / in emergency mode
+    bool public _emergencyMode;
```
marvinkruse: Does this one have to be `public`?

marvinkruse:
Besides the comments I like this a lot, good stuff!

src/factories/beacon/InverterBeacon.sol
FHieser: Yeah youre right. I had it intially public without the underscore but changed my mind later

src/factories/ModuleFactory.sol
FHieser: Honestly I just replaced the original IBeacon here that Pascal initially wrote
But looking at it more closely the ModuleFactory really only binds to the Interfaces.
Its not about the users interacting with the Interface its about the factory interacting with it in my opinion.

src/factories/ModuleFactory.sol
FHieser: xD :+1:

src/factories/beacon/InverterBeacon.sol
FHieser: I think you can write it both ways actually 
I had it in camelcase before and it looked kinda weird to me
I would keep it the way it was before :P

FHieser: Lol I looked it up again and shutdown as a verb is really not a thing :see_no_evil: 
Fuuu
Will change it then

src/factories/ModuleFactory.sol
FHieser: "You register the metadata to an InverterBeacon, not it's interface."
-> The factory actually registers only the interface not the InverterBeacon. As long as the interface is fulfilled anything can be hidden behind there at the current point of time

src/factories/beacon/InverterBeacon.sol
```diff
@@ -0,0 +1,159 @@
...
+        if (!(newImplementation.code.length > 0)) {
+            revert Beacon__InvalidImplementation();
+        }
+        _;
+    }
+
+    //--------------------------------------------------------------------------------
+    // State
+
+    /// @dev The beacon's implementation address.
```
0xNuggan, line 55: I feel these names are a bit confusing. Maybe something like "referenceImplementatition" and "activeImplementation" would be clearer. Something that makes clearer that `_implementation` is acting as a backup of sorts

0xNuggan (approved):
Great stuff. Not much to say, apart from my comment on the naming

src/factories/beacon/InverterBeacon.sol
FHieser: Yeah youre right
I adapted the naming to _implementationAddress and _implementationPointer. i think that should clear things up

PR #419, opened 2024-02-14
Fix: Re-add Scripts to CI & Fix Syntax

marvinkruse:
Fixes the CI (syntax error in the workflow file).

Status: works, but scripts fail, fixing that atm.


.vscode/settings.json
```diff
@@ -1,8 +1,9 @@
 {
     "solidity.packageDefaultDependenciesContractsDirectory": "src",
     "solidity.packageDefaultDependenciesDirectory": "lib",
-    "solidity.compileUsingRemoteVersion": "v0.8.19",
+    "solidity.compileUsingRemoteVersion": "v0.8.23",
     "solidity.formatter": "forge",
+    "git.detectSubmodules": false,
```
FHieser, line 6: What exactly does this do now?

FHieser (approved):

.vscode/settings.json
marvinkruse: Explained it in the commit message, copying from there:

> Adding the git.detectSubmodule setting to the config. This tells your local VSCode to ignore the submodules in the "git" tab. Which makes it a bit easier to work with, as we will only touch the submodules via forge commands anyways.

Thought it makes sense to add it? Found it quite nice to have a cleaner git-tab in VSCode.

marvinkruse:
I kinda don't like that current PRs are not using the CI (at least the test-part) at all. Should I just remove the scripts part in this one from the workflow (_as I haven't gotten it fixed yet, will still take a bit of time_), so that at least we have our regular CI back and then do a fresh PR for scripts in CI? @FHieser @0xNuggan @Zitzak

FHieser:
I was one step ahead of you there already @marvinkruse 
look at #421

marvinkruse:
> I was one step ahead of you there already @marvinkruse look at #421

Nice, good thinking - thank you!

FHieser:
@marvinkruse @0xNuggan With the latest change that I made the env variables are loaded directly from the dev.env file.
This has a few implications:
1. We have a central place of truth env variables for testing via the make file (instead of copy pasting them manually in there)
2. I assume (i dont know fully) that the way we handled the env variables before actually wasnt valid, which would have resulted that the standard .env files was used instead of the variables we set in the makefile.
3. We shouldnt use the makefile to deploy our contracts onchain. I assume that you did it via forge script anyway so there shouldnt be a problem, because that will still use the .env file. But we have to keep it in mind at least

.github/workflows/ci.yml
```diff
@@ -16,11 +16,15 @@ jobs:
...
         uses: foundry-rs/foundry-toolchain@v1
-        with:
-          version: nightly
 
-      - name: Install deps
+      - name: Install dependencies
         run: forge install
 
+      - name: Test stuff
+        run: make smth
```
marvinkruse: Is this still supposed to be here? :D

Makefile
```diff
@@ -12,12 +12,18 @@
...
+    include dev.env
+    export
+endif
+
 # -----------------------------------------------------------------------------
 # Common
-
-.PHONY: clean
-clean: ## Remove build artifacts
-	@forge clean
```
marvinkruse: Am I the only one using `make clean`? It's handy for when I'm updating the NatSpec for the integration tags, and this usually doesn't trigger the system to recompile AND redo the build file, so my solution is to run `make clean`. I can live with using `forge clean` by all means, just curious if maybe we should just keep this :D

.github/workflows/ci.yml
FHieser: no it was not xD
Removed it

Makefile
FHieser: I think at some point during the research I saw a error message that the clean command shadowed another command, but I cant currently find where that was the case
I reverted the change, but if we come across the error again i would recommend to remove it

.github/workflows/ci.yml
marvinkruse: thank you! :D

Makefile
marvinkruse: That's fine for me, all good. If it causes an issue just remove it, if not we keep it.

marvinkruse:
I think this all looks good to me. Just one thing which will become important now as we deploy something:
in the metadata of the modules, we currently just have `https://github.com/InverterNetwork` as the URL. we should treat this field properly. This means we probably set this to either
- `https://github.com/InverterNetwork/inverter-contracts`
or
- The URL to the actual contract being deployed, i.e.
  -  `https://github.com/InverterNetwork/inverter-contracts/blob/main/src/modules/logicModule/BountyManager.sol`

I believe the first version would be fine, as the main goal would be to distinguish between two modules called `BountyManager` in the future, one that we developed and one that potentially is being done by someone else (in their own repo). And in that case it would be enough to link to our repo, not the individual contract.
Thoughts?

FHieser:
I just copied the style that @Zitzak introduced.
Im fine with either one
I think for starters the https://github.com/InverterNetwork/inverter-contracts is fine and once we figure out how we properly do the versioning in name and repo we can switch

FHieser:
@marvinkruse changed it accordingly

marvinkruse:
> @marvinkruse changed it accordingly

Thank you, I will review it again in a sec, but I think I didn't find anything.

Zitzak (approved):
LGTM.

In regard to the metadata, I just copied what was already in the code ^^ so not a preference. I do agree that it should lead to the actual repo/contracts.

One thing which we brought up some time ago is that it makes sense to squash the commits after the approvals, so that is the last thing still open which I can see

script/external/DeployTransactionForwarder.s.sol
```diff
@@ -14,47 +14,25 @@ import {TransactionForwarder} from
...
-            "Deployment of TransactionForwarder Implementation at address ",
-            implementation
-        );
-        console2.log(
-            "Deployment of TransactionForwarder Beacon at address ", beacon
-        );
-        console2.log(
-            "Deployment of TransactionForwarder Proxy at address ", forwarder
-        );
+        console2.log("Deployment of implementation at address ", implementation);
```
0xNuggan: Maybe mention in the log that the deployment was for the TransactionForwarder implementation? Else, it gets hard to know when you have several logs

0xNuggan (approved):
LGTM. Only one little thing I commented

marvinkruse:
> One thing which we brought up some time ago is that it makes sense to squash the commits after the approvals, so that is the last thing still open which I can see

Yes, I always choose "squash and merge" (i think it's also set as the default), I agree that this is way better for the structure of our dev/main branches.

script/external/DeployTransactionForwarder.s.sol
marvinkruse: I fixed this 👍 good point!

PR #420, opened 2024-02-14
Security review part two fixes

0xNuggan:
Implements the changes mentioned in the security review

### DIFF
--- a/src/modules/authorizer/RoleAuthorizer.sol
+++ b/src/modules/authorizer/RoleAuthorizer.sol
@@ -87,9 +87,6 @@ contract RoleAuthorizer is
         // so they can whitelist an address which then will have full write access to the roles in the system. This is mainly intended for safety/recovery situations,
         // Modules can opt out of this on a per-role basis by setting the admin role to "BURN_ADMIN_ROLE".
 
-        //We preliminarily grant admin role to the caller
-        _grantRole(ORCHESTRATOR_OWNER_ROLE, _msgSender());
-
         // Set up OWNER role structure:
 
         // -> set OWNER as admin of itself
@@ -105,9 +102,10 @@ contract RoleAuthorizer is
 
         // If there is no initial owner specfied or the initial owner is the same as the deployer, the initial setup is finished at this point
         // If intialOwner corresponds to a different address, we need to set it up at this point and renounce the deployer
-        if (initialOwner != address(0) && initialOwner != _msgSender()) {
+        if (initialOwner != address(0)) {
             _grantRole(ORCHESTRATOR_OWNER_ROLE, initialOwner);
-            renounceRole(ORCHESTRATOR_OWNER_ROLE, _msgSender());
+        } else {
+            _grantRole(ORCHESTRATOR_OWNER_ROLE, _msgSender());
         }
     }
 
--- a/src/modules/authorizer/RoleAuthorizer.sol
+++ b/src/modules/authorizer/RoleAuthorizer.sol
@@ -100,8 +100,8 @@ contract RoleAuthorizer is
         // grant MANAGER Role to specified address
         _grantRole(ORCHESTRATOR_MANAGER_ROLE, initialManager);
 
-        // If there is no initial owner specfied or the initial owner is the same as the deployer, the initial setup is finished at this point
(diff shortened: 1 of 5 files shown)


FHieser (approved):

src/modules/authorizer/RoleAuthorizer.sol
```diff
@@ -105,9 +102,10 @@ contract RoleAuthorizer is
 
         // If there is no initial owner specfied or the initial owner is the same as the deployer, the initial setup is finished at this point
         // If intialOwner corresponds to a different address, we need to set it up at this point and renounce the deployer
```
marvinkruse: We don't renounce the role anymore, I think this line needs to be updated?

src/modules/base/Module.sol
```diff
@@ -97,23 +97,24 @@ abstract contract Module is
...
-            revert Module__CallerNotAuthorized(managerRole, _msgSender());
-        } else if (!authorizer.hasRole(ownerRole, _msgSender())) {
-            revert Module__CallerNotAuthorized(ownerRole, _msgSender());
+        if (
+            authorizer.hasRole(ownerRole, _msgSender())
+                || authorizer.hasRole(managerRole, _msgSender())
+        ) {
+            _;
+        } else {
+            if (authorizer.hasRole(ownerRole, _msgSender())) {
```
marvinkruse: Isn't this missing a `!`? Otherwise, this will never be reached, as the if-clause in line 101/102 is already checking for it. Or am I missing something?

My reasoning would be that the `else` is only entered if the user is neither an owner nor a manager, so the check here would be pointless?

marvinkruse (changes requested):

src/modules/base/Module.sol
Zitzak: Maybe to add to that, not sure if it is important with what kind of revert message it gets reverted, but the way it reverts now is reverse from how it was before, as well as the `revert Module__CallerNotAuthorized(ownerRole, _msgSender());` being the catch all revert message

0xNuggan: Yes to both. I think not being owner is a better catch-all message. Do you guys disagree?

marvinkruse: I think one error message is probably enough to signal that the issue of a failed tx is an authorization issue. So I would be fine with just reverting with the owner catch-all, personally.

marvinkruse (approved):

marvinkruse:
The CI failed, this was the error

> `2024-02-20T10:34:24.5883407Z Failing tests:
> 2024-02-20T10:34:24.5884275Z Encountered 1 failing test in test/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.t.sol:BancorVirtualSupplyBondingCurveFundingManagerTest
> 2024-02-20T10:34:24.5886282Z [31m[FAIL. Reason: BondingCurveFundingManagerBase__InvalidRecipient(); counterexample: calldata=0xe4893a4f0000000000000000000000001d1499e622d69689cdf9004d05ec547d650ff2115626ced1f02490c2e1492d8b135fc301b16684103951ed8ef849596b32634148 args=[0x1d1499e622D69689cdf9004d05Ec547d650Ff211, 38967472586288772459083690345774171327728501382326887155453953198174180163912 [2;49;39m[3.896e76][0m]][0m testBuyOrderFor(address,uint256) (runs: 329, μ: 304017, ~: 318753)`

Is that just an address collision from the fuzzing and known already or something new? @0xNuggan @Zitzak

Zitzak:
By now I fixed two address collition CI failures in the BC tests on two occasions. [Link of the PR](https://github.com/InverterNetwork/inverter-contracts/pull/423) from yesterdays fix.

I think rebasing/merging `dev_temp` should solve this issue, otherwise this might be a new one @marvinkruse.

PS: I'm unable to test these cases locally, otherwise I would go in and fix them all at the same time. Any suggestion on how to do this?

0xNuggan:
I added the fix to this branch @Zitzak . I'm not sure how we'd want to proceed... maybe just merge this (now that it passes) and tackle the temp_dev => dev merging afterwards? @marvinkruse

PR #424, opened 2024-02-21
Feat: Add KPI/Rewarder Module

0xNuggan:


0xNuggan:
Notes on the state of the KPI Rewarder at this point:

1) Integration tests are still todo. Right now all components in the test apart from the KPIModule itself are mocks.
2) E2E tests are still todo. Particularly on chain forks using the deployed instance of the Optimistic Oracle.
3) There seems to be a precision loss issue in the staking manager if the sizes of the stake are wildly different (User1: 4 - User 2: 9990e35 - etc). We need to review how to mitigate or remove that. @FHieser 
4) The stakingManager is still missing a function to claim rewards without unstaking. @FHieser 
5) The Optimistic Oracle Integrator is still missing the ERC165 check in the setter. This will be adressed shortly now that we have a general ERC165 checking infrastructure in place.

test/modules/logicModule/KPIRewarder.t.sol
```diff
@@ -0,0 +1,1040 @@
...
+        if (length > kpiManager.MAX_QUEUE_LENGTH()) {
+            length = kpiManager.MAX_QUEUE_LENGTH();
+        }
+
+        // check earned rewards are correct
+        for (uint i; i < length; i++) {
+            assertEq(kpiManager.balanceOf(users[i]), amounts[i]);
+
+            //=========================================================
+            // This is the place where imprecision issues arise. Needs review
```
FHieser, line 903: @0xNuggan Are you really sure this problem lies in the stakingmanager?
In the last commit I added fuzzing to the reward amounts and it went through the tests without problems
Also the only place this seems to be the problem is the continous section. Everywhere else it doesnt throw any errors right?
I would suggest looking at that part again. Especially the interaction between the estimateReward function from the StakingManager and the the Tranches in general. Especially because you didnt override it in the KPI rewarder

src/modules/logicModule/IKPIRewarder.sol
```diff
@@ -0,0 +1,151 @@
...
+    ) external;
+
+    /// @notice Posts an assertion to the Optimistic Oracle
+    /// @return assertionId The assertionId received for the posted assertion
+    function postAssertion() external returns (bytes32 assertionId);
+
+    /// @notice Creates a KPI for the Rewarder
+    /// @param _continuous Should the tranche rewards be distributed continuously or in steps
+    /// @param _trancheValues The value at which the tranches end
+    /// @param _trancheRewards The rewards to be diitributed at completion of each tranche
```
FHieser: *distributed

src/modules/logicModule/KPIRewarder.sol
```diff
@@ -0,0 +1,365 @@
...
+        uint targetValue,
+        uint targetKPI
+    ) external onlyOrchestratorOwner validBondSetup(asserter) {
+        //we do not control the dataId and data inputs since they are external and just stored in the oracle
+
+        setActiveTargetValue(targetValue);
+        setKPI(targetKPI);
+        _setAssertion(dataId, data, asserter);
+    }
+
```
FHieser, line 156: /// @inheritdoc IKPIRewarder

src/modules/logicModule/KPIRewarder.sol
```diff
@@ -0,0 +1,365 @@
...
+            _continuous,
+            _trancheValues,
+            _trancheRewards
+        );
+
+        return (KpiNum);
+    }
+
+    /// @inheritdoc IKPIRewarder
+    function setKPI(uint _KPINumber) public onlyOrchestratorOwner {
```
FHieser: What is the point of having this function if you have prepareAssertion?

src/modules/logicModule/KPIRewarder.sol
```diff
@@ -0,0 +1,365 @@
...
+        if (_KPINumber >= KPICounter) {
+            revert Module__KPIRewarder__InvalidKPINumber();
+        }
+        activeKPI = _KPINumber;
+    }
+
+    /// @inheritdoc IKPIRewarder
+    function setActiveTargetValue(uint targetValue)
```
FHieser: What is the point of having this function if you have prepareAssertion?

src/modules/logicModule/KPIRewarder.sol
```diff
@@ -0,0 +1,365 @@
...
+    ) public override {
+        if (_msgSender() != address(oo)) {
+            revert Module__OptimisticOracleIntegrator__CallerNotOO();
+        }
+
+        if (assertedTruthfully) {
+            // SECURITY NOTE: this will add the value, but provides no guarantee that the fundingmanager actually holds those funds.
+
+            // Calculate rewardamount from assertionId value
+            KPI memory resolvedKPI =
```
FHieser, line 314: Why do we have a Active KPI if you dont trigger it here?

src/modules/logicModule/KPIRewarder.sol
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    // ========================================================================
+    // Assertion Manager functions:
+
+    /// @inheritdoc IKPIRewarder
+    /// @dev about the asserter address: any address can be set as asserter, it will be expected to pay for the bond on posting.
+    /// The bond tokens can also be deposited in the Module and used to pay for itself, but ONLY if the bond token is different from the one being used for staking.
+    /// If the asserter is set to 0, whomever calls postAssertion will be paying the bond.
+    function prepareAssertion(
```
FHieser: Why do we need an active assertion in general here?
Technically we could have multiple different ones at the same time.
We could just create a AssertionStruct that saves all the accoring data and then sent it in the postAssertion field

FHieser (commented):
From an implementation logic standpoint it looks good I think,
but there is definitely a how to guide needed. There are a lot of keywords flying around here and bringing them in order via a guide would be pretty great.

Also I think we could actually remove the "active" Assertion part (including active KPI, targetValue etc) in here and just keep them saved as a Assertion struct for later reference.   Technically these assertions could be (and I think actually currently can be) run in Parallel if you wanted them. The declaration of them being "Active" doesn't really mean much if I understood it correctly.

0xNuggan:
Thanks for the review @FHieser  ! I agree with you that the whole prepareAssertion/activeKPI is too convoluted. I've removed it and added a `dequeue()` function to allow users to withdraw funds that have not been staked yet (in case `postAssertion` doesn't happen).

I've also added an explainer and some extra tests

FHieser (commented):
Nice cleanup.
I think that makes the code pretty tight.
Do we still have the issues on the tests?

0xNuggan:
The precision part yes sadly. I've updated the corresponding test, but the issue persists

PR #426, opened 2024-02-29
[REVIEW] Feat: MetaTx/Multicall + Intervention Mechanism + OZ v5

### LINKED ISSUE #427: [SC-191] Security Review - MetaTx/Multicall, Intervention + OZv5
Description can be found in the [PR](https://github.com/InverterNetwork/inverter-contracts/pull/350).

<sub>[SC-191](https://linear.app/inverter/issue/SC-191/security-review-metatxmulticall-intervention-ozv5)</sub>

marvinkruse:
⚠️ **This PR merges from `review` into the `dev` branch.** ⚠️
_At the time of creating this PR (until it is merged), the `review` branch contains the latest state that has received a security review, including the fixes. Anything in `review` has been reviewed internally, but not received a security review. We are merging into `dev` as the current `dev` branch is still pending the finalization of the audit._
<hr>

➡️ **Process**
This PR will receive an external security review. Additionally, at least one of our developers will review it as well, ideally someone who didn't work on it. Please raise any findings/questions directly within this PR, either as a code comment, referencing the corresponding line or as a general comment to the discussion. Please mark comments that are resolved as resolved.
<hr>

📋 **Contents**
This PR contains...
- MetaTx and Multicall support ([PR](https://github.com/InverterNetwork/inverter-contracts/pull/405))
- Update to OpenZeppelin v5 ([PR](https://github.com/InverterNetwork/inverter-contracts/pull/401))
- Adds first Intervention Mechanisms ([PR](https://github.com/InverterNetwork/inverter-contracts/pull/418))


src/modules/base/Module.sol
```diff
@@ -290,4 +290,50 @@ abstract contract Module is
...
+        internal
+        view
+        virtual
+        override(ContextUpgradeable)
+        returns (address sender)
+    {
+        if (__Module_orchestrator.isTrustedForwarder(msg.sender)) {
+            // The assembly code is more direct than the Solidity version using `abi.decode`.
+            /// @solidity memory-safe-assembly
+            assembly {
```
omega-audits: Should also verify calldatasize is more than 20 to be sure

src/modules/base/Module.sol
```diff
@@ -290,4 +290,50 @@ abstract contract Module is
...
+    }
+
+    function _msgData()
+        internal
+        view
+        virtual
+        override(ContextUpgradeable)
+        returns (bytes calldata)
+    {
+        if (__Module_orchestrator.isTrustedForwarder(msg.sender)) {
```
omega-audits: Also here should check data length is over 20

src/factories/ModuleFactory.sol
```diff
@@ -27,13 +29,18 @@ import {
...
+ *      implementations. Note that a metadata's registered {IInverterBeacon}
  *      implementation can not be changed after registration!
  *
  * @author Inverter Network
  */
-contract ModuleFactory is IModuleFactory, Ownable2Step, ERC165 {
+contract ModuleFactory is
+    IModuleFactory,
+    ERC2771Context,
+    Ownable2Step,
```
omega-audits, line 41: Inheritance here is wrong, like this the call to super will refer to Context and not ERC2771Context as intended later

src/factories/ModuleFactory.sol
```diff
@@ -161,4 +169,39 @@ contract ModuleFactory is IModuleFactory, Ownable2Step, ERC165 {
...
+
+    /// Needs to be overriden, because they are imported via the Ownable2Step as well
+    function _msgSender()
+        internal
+        view
+        virtual
+        override(Context, ERC2771Context)
+        returns (address sender)
+    {
+        return super._msgSender();
```
omega-audits: Since the Ownable2Step is inherited from after ERC2771Context, super here will use Context, not ERC2771Context. 
It's best to write ERC2771Context._msgSender() instead to not confuse multiple inheritance here.

src/factories/beacon/InverterBeacon.sol
```diff
@@ -0,0 +1,161 @@
...
+
+    //--------------------------------------------------------------------------------
+    // onlyOwner Intervention Mechanism
+
+    /// @inheritdoc IInverterBeacon
+    function shutDownImplementation() external onlyOwner {
+        //Go into emergency mode
+        _emergencyMode = true;
+        //Set implementation pointer to address 0 and therefor halting the system
+        _implementationPointer = address(0);
```
omega-audits, line 121: Making the implementation 0 will make the registerMetadata and createModule revert if the beacon is in an emergency mode. If this is intended then the use of assert in createModule is no longer appropriate, and if not intended maybe use a different burn address here for emergency mode.

src/factories/ModuleFactory.sol
```diff
@@ -161,4 +169,39 @@ contract ModuleFactory is IModuleFactory, Ownable2Step, ERC165 {
...
+
+    /// Needs to be overriden, because they are imported via the Ownable2Step as well
+    function _msgData()
+        internal
+        view
+        virtual
+        override(Context, ERC2771Context)
+        returns (bytes calldata)
+    {
+        return super._msgData();
```
omega-audits: same inheritance problem here as with `_msgSender()`

src/factories/ModuleFactory.sol
```diff
@@ -161,4 +169,39 @@ contract ModuleFactory is IModuleFactory, Ownable2Step, ERC165 {
...
+    }
+
+    function _contextSuffixLength()
+        internal
+        view
+        virtual
+        override(Context, ERC2771Context)
+        returns (uint)
+    {
+        return 20;
```
omega-audits: So I guess hardcoding is not wrong here, but it would be more elegant and future-rpoof to write `ERC2771Context._contextSuffixLenght()` here

src/factories/beacon/InverterBeacon.sol
omega-audits: If this is intended, you should also update your comment:

```
        // As a zero address implementation indicates an unrecoverable state
        // and faulty update from the beacon's owner, the beacon should be
        // considered dangerous. We therefore make sure that nothing else can
        // happen in this tx and burn all remaining gas.
        // Note that while the inverter's beacon implementation forbids an
        // implementation update to non-contract addresses, we can not ensure
        // a module does not use a different beacon implementation.
    
```

src/modules/base/Module.sol
omega-audits: It is better to use the ERC2771Context methods from OZ instead of writing your own, to avoid these kind of mistakes, toh here as in _msgData. You can override isTrustedForwarder easily

src/factories/beacon/InverterBeacon.sol
FHieser: Yes your definitely correct here
Having the address(0) as an implementation should ideally be temporary only
In the case of registerMetadata I think its important to check if the beacon had a implementation in the first place before we register it, so i would leave that implementation as is
In the case of the createmodule function I would remove that check as it is no longer a assumption that holds up

src/modules/base/Module.sol
FHieser: I did it as you recommended and inherited from ERC2771Context
I hope I did the inheritance correctly this time :3

src/modules/base/Module.sol
FHieser: see comment above

0xNuggan:
Last change looks good to me. I still think maybe it's too verbose to declare the isTrustedForwarder in the Orchestrator itself only to overwrite it, but as we said it's just taste, and if it makes interactions from other Modules easier then so be it :)

0xNuggan (approved):

PR #429, opened 2024-03-06
SC-103: Complete all testing for the FundingManager

Zitzak:
## What has been done?
- Completed all the testing
- Moved function from implementation into base contracts
- Added functions to base contract
- Made all functions virtual in base contracts

## Implements
- https://linear.app/inverter/issue/SC-103/complete-all-testing-for-the-funding-manager

### DIFF
--- a/src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
+++ b/src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
@@ -84,7 +84,7 @@ contract ToposBondingCurveFundingManager is
     IToposFormula public formula;
     /// @notice Repayable amount collateral which can be pulled from the contract by the liquidity pool
     uint public repayableAmount;
-    /// @dev The current seize percentage expresses in BPS
+    /// @dev The current seize percentage expressed in BPS
     uint64 public currentSeize;
     /// @dev Address of the liquidity pool who has access to the collateral held by the funding manager
     /// through the Repayer functionality
@@ -136,7 +136,7 @@ contract ToposBondingCurveFundingManager is
         formula = IToposFormula(bondingCurveProperties.formula);
         // Set capital required
         _setCapitalRequired(bondingCurveProperties.capitalRequired);
-        // Set current seize to MAX_SEIZE
+        // Set currentSeize
         _setSeize(bondingCurveProperties.seize);
         // Set base price multiplier
         _setBaseMultiplier(bondingCurveProperties.basePriceMultiplier);
--- a/src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
+++ b/src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
@@ -308,6 +308,7 @@ contract ToposBondingCurveFundingManager is
         __Module_orchestrator.fundingManager().token().safeTransfer(
             _to, _amount
         );
+        emit RepaymentTransfer(_to, _amount);
     }
 
     //--------------------------------------------------------------------------
@@ -508,9 +509,13 @@ contract ToposBondingCurveFundingManager is
 
     /// @dev Precomputes and sets the price multiplier to capital ratio
     function _updateVariables() internal {
-        basePriceToCapitalRatio = _calculateBasePriceToCapitalRatio(
+        uint newBasePriceToCapitalRatio = _calculateBasePriceToCapitalRatio(
             capitalRequired, basePriceMultiplier
         );
+        emit BasePriceToCapitalRatioChanged(
+            basePriceToCapitalRatio, newBasePriceToCapitalRatio
+        );
+        basePriceToCapitalRatio = newBasePriceToCapitalRatio;
     }
 
     /// @dev Internal function which calculates the price multiplier to capital ratio
--- a/src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
+++ b/src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
@@ -138,7 +138,7 @@ contract ToposBondingCurveFundingManager is
         _setCapitalRequired(bondingCurveProperties.capitalRequired);
         // Set current seize to MAX_SEIZE
         _setSeize(bondingCurveProperties.seize);
-        // Set initial
+        // Set base price multiplier
         _setBaseMultiplier(bondingCurveProperties.basePriceMultiplier);
         // Set sell fee to Max fee at initiation
         _setSellFee(bondingCurveProperties.sellFee);
@@ -457,6 +457,7 @@ contract ToposBondingCurveFundingManager is
         if (_seize > MAX_SEIZE) {
             revert ToposBondingCurveFundingManager__InvalidSeize(_seize);
         }
+        emit SeizeChanged(currentSeize, _seize);
         currentSeize = _seize;
     }
 
--- a/src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
+++ b/src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
@@ -67,12 +67,14 @@ contract ToposBondingCurveFundingManager is
     uint public constant MIN_RESERVE = 1 ether;
     /// @dev Max seizable amount is 1% expressed in BPS
     uint64 public constant MAX_SEIZE = 100;
-    /// @dev Max fee for selling is TODO: See Qs, expressed in BPS
-    uint64 public constant MAX_SELL_FEE = 100; // BPS, 2.5%
+    /// @dev Max fee for selling is 1% expressed in BPS
+    uint64 public constant MAX_SELL_FEE = 100;
     /// @dev Time interval between seizes
     uint64 public constant SEIZE_DELAY = 7 days;
 
+    /// @dev Role associated with the managing of the bonding curve values
     bytes32 public constant RISK_MANAGER_ROLE = "RISK_MANAGER";
+    /// @dev Role associated with the managing of setting withdraw addresses and setting the fee
     bytes32 public constant COVER_MANAGER_ROLE = "COVER_MANAGER";
 
     //--------------------------------------------------------------------------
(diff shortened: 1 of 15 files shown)


test/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.t.sol
```diff
@@ -127,10 +183,1571 @@ contract ToposBondingCurveFundingManagerTest is ModuleTest {
...
+            _depositAmount,
+            bondingCurveFundingManager.call_getCapitalAvailable(),
+            bondingCurveFundingManager.basePriceToCapitalRatio()
+        );
+        // Get return value
+        uint functionReturnValue = bondingCurveFundingManager
+            .call_redeemTokensFormulaWrapper(_depositAmount);
+
+        // Because of precision loss, the assert is done to be in range of 0.0000000001% of each other
+        assertApproxEqRel(functionReturnValue, redeemAmount, 0.0000000001e18);
```
Zitzak, line 1289: @marvinkruse I could not get both values equal, which seemed odd to me, as the precision loss should be the same if there is any. Would you know why? The diff is very minimal which I resolved in the testing like this

src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
```diff
@@ -132,13 +134,21 @@ contract ToposBondingCurveFundingManager is
         liquidityPool = ILiquidityPool(_liquidityPool);
         // Set formula contract
         formula = IToposFormula(bondingCurveProperties.formula);
+        // Set capital required
         _setCapitalRequired(bondingCurveProperties.capitalRequired);
+        // Set current seize to MAX_SEIZE
+        _setSeize(bondingCurveProperties.seize);
+        // Set initial
```
marvinkruse: Should this say `/// Set initial base price multiplier` or is this intentional?

src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
```diff
@@ -132,13 +134,21 @@ contract ToposBondingCurveFundingManager is
...
+        _setBaseMultiplier(bondingCurveProperties.basePriceMultiplier);
         // Set sell fee to Max fee at initiation
-        _setSellFee(MAX_SELL_FEE);
+        _setSellFee(bondingCurveProperties.sellFee);
+        // Set buying functionality to open if true. By default buying is false
+        if (bondingCurveProperties.buyIsOpen == true) _openBuy();
+        // Set selling functionality to open if true. By default selling is false
+        if (bondingCurveProperties.sellIsOpen == true) _openSell();
```
marvinkruse, line 148: Just caught my eye that you're using `bondingCurveProperties.buyIsOpen == true` instead of just `bondingCurveProperties.buyIsOpen` which does the same. Do you just do this for simplicity in terms of reading? Was just curious as I think the Bonding Curve contracts are the only ones that do this afaik. Not important and doesn't necessarily need a change, just out of curiosity. It compiles to the same code at least so no difference functionally.

src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
```diff
@@ -457,12 +452,22 @@ contract ToposBondingCurveFundingManager is
     //--------------------------------------------------------------------------
     // Internal Functions
 
+    /// @dev Set the current seize state, which defines the percentage of seizable amount
+    function _setSeize(uint64 _seize) internal {
+        if (_seize > MAX_SEIZE) {
+            revert ToposBondingCurveFundingManager__InvalidSeize(_seize);
+        }
+        currentSeize = _seize;
+    }
```
marvinkruse: Do we want an event for updating the `currentSeize`? Just thinking in terms of our goal that we would ideally like to be able to rebuild the entire onchains state from events for potential indexing solutions.

marvinkruse: There is one for example for `CapitalRequiredChanged`, maybe it's more consistent to have one here as well there?

marvinkruse (changes requested):
No things on the logic front that I found, all looking solid. Just some minor things again.

test/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.t.sol
marvinkruse: Hm.. weird indeed, I expected it to be the same as well. DId you check what the difference is when it appears? Is it just 1 wei or anything bigger than that? Then at least it might be apparent that it's a clear precision loss from the division in for example the `basePriceToCapitalRatio`. But not sure why it would lead to different results, both is done in Solidity with the same values. Or is the order in which these calcs happen different?

I will check again in the morning after a coffee, maybe I see it then..  In any case, the solution you have here is good enough for me 😁

Zitzak: So the difference was not a constant but changed with the fuzzing, and as I tried to define the allowed delta. So the value I now came up with works, but it's indeed weird because basically the same function with the same values get called

src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
Zitzak: You're right, did not finish the sentence. Will update

src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
Zitzak: Yeah I tend to do this for clarity. I do always try to opt for readability as you've prob have seen in the tests as well :))

However a question back to you would be if there are any negative gas associations to this, in which case I would def change it

src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
Zitzak: You're right, makes sense. It wasn't in the original contracts weirdly enough so overlooked it

Zitzak: Quick question in regard to "being able to rebuild the on-chain state for indexing". The collateral fee collected value is internal and no event emit reflects this state. My thought process was that it shouldn't be easily retrievable for everyone how much fee has been collected. However this would be one value that is not rebuildable through emits in this way.

Zitzak:
Added 2 more events emits within the implementation contract, based on your comment about being able to get the on-chain state for indexing @marvinkruse

src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
marvinkruse: Thank you! If I wanna be in turbo-pedantic mode I would say that it should say "expressed" instead of "expresses", but you can also ignore this :D Thanks for changing the sentence!

src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
marvinkruse: No gas implications, as it both compiles to the assembly code, so all good. Ideally, I'd love to have the codebase be consistent, which it currently isn't, but let's just keep this like it is and discuss this at a different time with the rest briefly to decide on a general level, I'd say.

src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
marvinkruse: Hm.. good question. I think we don't need an event for changes to `tradeFeeCollected`, seems a bit overkill for me to have one if I think about it.

As it is exposed via `getCollateralTradeFeeCollected()` in `BondingCurveFundingManagerBase `, it is not hidden though, not sure whether you meant that? In the end, an indexer can still track it as it's publically exposed via a this getter function, so we can make use of it in a potential dashboard to track fees etc. (which we probably want at some point). Users are also able to see the value easily - which to me is not a problem, I wouldn't hide it (just not rub it in their face via an event 😁)

Would you agree?

test/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.t.sol
marvinkruse: Hm.. weird indeed. I don't think this should block the PR though, even if I'd like to understand this eventually. Let's go with the solution you have here and maybe either get the auditors to verify this or revisit this on our own down the road (before we deploy to mainnet).

src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
```diff
@@ -132,13 +134,21 @@ contract ToposBondingCurveFundingManager is
         liquidityPool = ILiquidityPool(_liquidityPool);
         // Set formula contract
         formula = IToposFormula(bondingCurveProperties.formula);
+        // Set capital required
         _setCapitalRequired(bondingCurveProperties.capitalRequired);
+        // Set current seize to MAX_SEIZE
```
marvinkruse: Isn't this set to `bondingCurveProperties.seize` instead of `MAX_SEIZE`?

src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
Zitzak: So the `getCollateralTradeFeeCollected()` function has the `onlyOrchestratorOwner` modifier, which would inhibit it from being publicly available.

However reading your comment and seeing that you don't see an issue with users seeing the value, would it then make sense to just remove the getter function and make the `tradeFeeCollected` public, maybe also renaming it so it reflects that collateral fee is collected here?

src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
Zitzak: You're right. This is still an old comment. Changed it

src/modules/fundingManager/bondingCurveFundingManager/ToposBondingCurveFundingManager.sol
marvinkruse: Oh, I missed that modifier indeed, good point! I think making the variable itself public is fine, also agree on the renaming. 👍

marvinkruse:
I believe the only thing missing is the change to `tradeFeeCollected` (making it public and renaming?) and then potentially removing the `getCollateralTradeFeeCollected()` function, right? 
At least everything else looks good to me, so once that's there we're good to go from my side!

marvinkruse (approved):
great stuff!

PR #430, opened 2024-03-06
Create digest helper function for meta tx forwarder

FHieser:
This solves this issue
https://linear.app/inverter/issue/SC-198/create-digest-helper-function-for-metatx-forwarder


marvinkruse (approved):
Looks good to me!

src/external/forwarder/TransactionForwarder.sol
```diff
@@ -61,6 +76,28 @@ contract TransactionForwarder is
         }
     }
 
+    //--------------------------------------------------------------------------
+    // Internal
+
+    function getStructHash(ERC2771Forwarder.ForwardRequestData memory req)
```
Zitzak: Should have an underscore to keep internal function convention

Zitzak (changes requested):
Found only one small thing. Otherwise looks good!

src/external/forwarder/TransactionForwarder.sol
FHieser: @Zitzak added it accordingly
Will merge with the approve of Marvin K

PR #431, opened 2024-03-12
Fix scripts

FHieser:
As the name suggests


FHieser:
Decided to do this in the #419 PR

PR #435, opened 2024-04-02
Fix: Enable <18 Decimal Tokens as Collateral in Bonding Curve

0xNuggan:
changes to core contracts:
- padding the virtual collateral to have the same amount of decimals than the issued token.

main changes in tests:
- bound the fuzzed input in a way that considers the decimals the token has so we avoid overflows when padding
- use the padded version when we are directly calling the formula

### DIFF
--- a/src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
+++ b/src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
@@ -234,34 +234,6 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
         _virtualSellOrder(_msgSender(), _depositAmount, _minAmountOut);
     }
 
-    /// @inheritdoc IVirtualTokenSupply
-    /// @dev Since the supply is normalized internally to 18 decimals for use in the curve, when an external function queries it we convert it to it's native decimal amount
-    function getVirtualTokenSupply()
-        external
-        view
-        override(VirtualTokenSupplyBase)
-        returns (uint)
-    {
-        uint raw_tokenSupply = _getVirtualTokenSupply();
-        return _convertAmountToRequiredDecimal(
-            raw_tokenSupply, eighteenDecimals, decimals()
-        );
-    }
-
-    /// @inheritdoc IVirtualCollateralSupply
-    /// @dev Since the supply is normalized internally to 18 decimals for use in the curve, when an external function queries it we convert it to it's native decimal amount
-    function getVirtualCollateralSupply()
-        external
-        view
-        override(VirtualCollateralSupplyBase)
-        returns (uint)
-    {
-        uint raw_collateralSupply = _getVirtualCollateralSupply();
-        return _convertAmountToRequiredDecimal(
-            raw_collateralSupply, eighteenDecimals, collateralTokenDecimals
-        );
-    }
-
     /// @inheritdoc IBancorVirtualSupplyBondingCurveFundingManager
     function getReserveRatioForBuying() external view returns (uint32) {
         return reserveRatioForBuying;
@@ -407,21 +379,32 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
         override(BondingCurveFundingManagerBase)
         returns (uint mintAmount)
     {
+        // Convert virtual supply and balance to 18 decimals
+        uint decimalConvertedVirtualTokenSupply =
+        _convertAmountToRequiredDecimal(
+            virtualTokenSupply, decimals(), eighteenDecimals
+        );
+
+        uint decimalConvertedVirtualCollateralSupply =
+        _convertAmountToRequiredDecimal(
+            virtualCollateralSupply, collateralTokenDecimals, eighteenDecimals
+        );
+
         // Convert depositAmount to 18 decimals, which is required by Bancor formula
         uint decimalConvertedDepositAmount = _convertAmountToRequiredDecimal(
             _depositAmount, collateralTokenDecimals, eighteenDecimals
         );
 
         // Calculate mint amount through bonding curve
-        mintAmount = formula.calculatePurchaseReturn(
-            virtualTokenSupply,
-            virtualCollateralSupply,
+        uint decimalConvertedMintAmount = formula.calculatePurchaseReturn(
+            decimalConvertedVirtualTokenSupply,
+            decimalConvertedVirtualCollateralSupply,
             reserveRatioForBuying,
             decimalConvertedDepositAmount
         );
         // Convert mint amount to issuing token decimals
         mintAmount = _convertAmountToRequiredDecimal(
-            mintAmount, eighteenDecimals, decimals()
+            decimalConvertedMintAmount, eighteenDecimals, decimals()
         );
     }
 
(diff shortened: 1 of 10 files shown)


src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -454,7 +460,7 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
         // An input verification is needed here since the Bancor formula, which determines the
         // issucance price, utilizes PPM for its computations. This leads to a precision loss
         // that's too significant to be acceptable for tokens with fewer than 7 decimals.
-        if (_decimals < 7) {
+        if (_decimals < 7 || _decimals > collateralTokenDecimals) {
```
marvinkruse: Is this gonna be a problem for Tokens like USDC and Tether, as they are both 6 decimal tokens?

test/modules/fundingManager/RebasingFundingManager.t.sol
```diff
@@ -46,7 +46,7 @@ contract RebasingFundingManagerTest is ModuleTest {
     uint internal constant DEPOSIT_CAP = 100_000_000e18;
 
     // Other constants.
-    uint8 private constant DECIMALS = 18;
+    //uint8 private constant DECIMALS = 18;  //Question : Is there any reason to hardcode the decimals for this fundingmanager to 18?
```
marvinkruse, line 49: Is this for the collateral token or the issuance token? If it's for the collateral token, then it shouldn't be hardcoded, right?

edit: this might be for the issuance token as well, I think I initially misunderstood.

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
0xNuggan: You are of course completely right, this is what I get for rushing to quickly update the setter before the dev call :laughing: It should be as mentioned in the interface. I pushed the fix

test/modules/fundingManager/RebasingFundingManager.t.sol
0xNuggan: It's for the token that the bonding curve will issue. But I guess since we are going to make changes anyway for minting from an existing token, we shouldn't hardcode it

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
marvinkruse: Haha, all good! 😂 
So the bonding curve (issuance) token decimals need to be at least 7 (for precision) and also need to be at least equal to the collateral tokens decimals, not lower, correct? That's how I understand the fix you pushed now, which looks good to me, just making sure.

Initially, I thought `_decimals` was for the collateral token, but it's for the issuance token, I think I get it now. Is there a good reason why someone would set the issuance tokens decimals to anything but 18? (just out of curiosity)

0xNuggan: Yes, exactly! 
Personally, I can't think of a good reason to not have 18 decimals. But if we are going to "plug in" our bonding curve to an external token maybe that token had a good reason for it? So better to keep it open...

0xNuggan:
Thinking a bit more about what @marvinkruse said, we'll probably need to do similar checks for the issuance token once we allow using external contracts for it. But for now we can limit it to the collateral

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -475,7 +478,12 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
         (uint amountIssued, uint feeAmount) =
             _buyOrder(_receiver, _depositAmount, _minAmountOut);
         _addVirtualTokenAmount(amountIssued);
-        _addVirtualCollateralAmount(_depositAmount - feeAmount);
+        uint normalizedDepositAmount = _convertAmountToRequiredDecimal(
+            (_depositAmount - feeAmount),
+            collateralTokenDecimals,
```
Zitzak: As the amounts now get normalized to 18 decimals, if we indeed use for example USDC with 6 decimals as collateral, and a UI would like to get the virtual supply, don't they get this now in 18 decimals. So the dAPP interfacing with the contract has to know to calculate the value back to it's orginal decimals like this right?

Maybe it makes more sense to always convert it back to the original decimals, i.e. in my example this would be 6. Wdyt @0xNuggan

test/e2e/fundingManager/BondingCurveFundingManagerE2E.sol
```diff
@@ -177,7 +177,8 @@ contract BondingCurveFundingManagerE2E is E2ETest {
             );
 
             fundingManager.sell(
-                fundingManager.balanceOf(bob), fundingManager.balanceOf(bob)
+                fundingManager.balanceOf(bob),
+                fundingManager.balanceOf(bob) // Question: What happens with the fee? Shouldn't the amounts be wildly different since they are different tokens?
```
Zitzak: Unsure what is meant by the questions. Could you elaborate pls.

test/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.t.sol
```diff
@@ -1467,12 +1669,22 @@ contract BancorVirtualSupplyBondingCurveFundingManagerTest is ModuleTest {
         );
         // Static price has increased after buy
         assertGt(staticPriceAfterBuy, staticPriceBeforeBuy);
+        /*
+        Question:Unsure about how to proceed here. If the supply and collateral tokens have a big difference in decimals (i.e. 18 and 6) and we start with virtual 
+        supply/collateral of 1/1, the first buy will move the price by a lot, since we will be effectively depositing 1e12 collateral tokens (1 + padding to 18). 
+        This will exceed the delta by a lot, but it IS expected behavior.
+        If we set the initial supply to 1e18 and the initial collateral t 1e12 (what we can call "realistic values"), the delta holds again. But on the ohter hand, we
+        aren't testing on the empty curve, just a "normal" point in it.
```
Zitzak: Personally I would go with testing for realistic values. The BCRG is also building a simulation tool which will also (I think) help with defining the edge cases which we should then warn the user for

Zitzak (commented):
Only thing I found is the questions around always normalizing it, and how we should handle it for when the virtual supply is retrieved.

0xNuggan:
After a talk with @Zitzak  we decided that external facing function should return the collateral and supply values in the native decimals, instead of the normalized ones. I've changed that and added some tests

Zitzak (approved):
LGTM 🚀

0xNuggan:
@marvinkruse do you think we can merge this?

marvinkruse (approved):
lgtm as well! @0xNuggan

PR #436, opened 2024-04-04
Feat: Add Governance Contract

FHieser:
Solves SC-48


FHieser:
@marvinkruse the e2e and scripts are missing, but the logic is done as well as most of the tests (i want to still do a small improvement there)
With that I think you can already give your feedback

src/external/governance/GovernanceContract.sol
```diff
@@ -0,0 +1,339 @@
...
+        emit BeaconImplementationRestarted(beacon);
+    }
+
+    //---------------------------
+    //Ownable2Step
+
+    /// @inheritdoc IGovernanceContract
+    function acceptOwnership(address adr)
+        external
+        onlyCommunityOrTeamMultisig //@note correct Role?
```
marvinkruse: I can't see any downside of allowing both multisigs to accept contract ownership, as they can't really do anything to them outside of the limits we set here, so I personally think this is fine. I'm curious to see what @0xNuggan or @Zitzak think, though!

src/external/governance/GovernanceContract.sol
```diff
@@ -0,0 +1,339 @@
...
+            //@note should be beacon?
```
marvinkruse: I think we can keep it open to any contract, not just beacons, that we may claim ownership of.

src/external/governance/IGovernanceContract.sol
```diff
@@ -0,0 +1,194 @@
...
+        address intendedImplementation;
+        /// @dev The new intended minor Version of the beacon
+        uint intendedMinorVersion;
+    }
+
+    //--------------------------------------------------------------------------
+    // Errors
+
+    /// @notice The given address is invalid
+    error InvalidAddress(address adr);
```
Zitzak: Should have our usual convention of `Governance__InvalidAddress()`. Same for the other custom errors

src/external/governance/GovernanceContract.sol
```diff
@@ -0,0 +1,339 @@
...
+} from "src/factories/beacon/InverterBeacon.sol";
+
+// External Dependencies
+import {ERC165} from "@oz/utils/introspection/ERC165.sol";
+import {Initializable} from "@oz-up/proxy/utils/Initializable.sol";
+import {AccessControl} from "@oz/access/AccessControl.sol";
+import {Ownable} from "@oz/access/Ownable.sol";
+import {Ownable2Step} from "@oz/access/Ownable2Step.sol";
+
+contract GovernanceContract is
```
Zitzak: Do we want to go with the new naming convention here is well? What ever we do, I think the word "contract" should be removed/replaced. Maybe back to manager 😂 or handler maybe?

src/external/governance/GovernanceContract.sol
```diff
@@ -0,0 +1,339 @@
...
+
+    modifier validAddress(address adr) {
+        if (adr == address(0)) {
+            revert IGovernanceContract.InvalidAddress(adr);
+        }
+        _;
+    }
+
+    modifier validAmount(uint amt) {
+        if (amt == 0) {
```
Zitzak: Should we define a minimum timelock so this can't be used to set a non-zero but very small timelock to circumvent it sort of

src/external/governance/GovernanceContract.sol
```diff
@@ -0,0 +1,339 @@
...
+
+        timelockPeriod = newTimelockPeriod;
+    }
+
+    //--------------------------------------------------------------------------
+    // Getter Functions
+
+    /// @inheritdoc IGovernanceContract
+    function getBeaconTimelock(address beacon)
+        external
```
Zitzak: State mutability can be restricted to view

src/external/governance/GovernanceContract.sol
```diff
@@ -0,0 +1,339 @@
...
+
+    /// @inheritdoc IGovernanceContract
+    function getFeeManager() external view returns (address) {
+        return feeManager;
+    }
+
+    /// @inheritdoc IGovernanceContract
+    function setFeeManager(address newFeeManager)
+        external
+        onlyRole(COMMUNITY_MULTISIG_ROLE)
```
Zitzak: So the community in this context is the wider group of stakeholders (like investors) right? Does this mean that they would have the ability to basically disable our fee collecting mechanism?

src/external/governance/GovernanceContract.sol
```diff
@@ -0,0 +1,339 @@
...
+    modifier validAddress(address adr) {
+        if (adr == address(0)) {
+            revert IGovernanceContract.InvalidAddress(adr);
+        }
+        _;
+    }
+
+    modifier validAmount(uint amt) {
+        if (amt == 0) {
+            revert IGovernanceContract.InvalidAmount(amt);
```
Zitzak: Wouldn't `InvalidAmount()` work without referencing in through the interface?

src/external/governance/GovernanceContract.sol
```diff
@@ -0,0 +1,339 @@
...
+        onlyCommunityOrTeamMultisig
+        accessableBeacon(beacon)
+        upgradeProcessAlreadyStarted(beacon)
+        timelockPeriodExceeded(beacon)
+    {
+        //set timelock to inaktive
+        beaconTimelock[beacon].timelockActive = false;
+
+        //Upgrade beacon
+        IInverterBeacon(beacon).upgradeTo(
```
Zitzak: Just for my own understanding. So this governance contract will be the owner of all the beacons? If so, then do we also need a way to transfer ownership from the beacons to another address in case that is needed in the future?

src/external/governance/GovernanceContract.sol
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    /// @inheritdoc IGovernanceContract
+    function forceUpgradeBeaconAndRestartImplementation(
+        address beacon,
+        address newImplementation,
+        uint newMinorVersion
+    )
+        external
+        onlyRole(COMMUNITY_MULTISIG_ROLE)
```
Zitzak: The Interface states that both community and team multisig should be able to call this function. The comment probably needs to be corrected as this function (if I understand correctly) circumvents the timelock and so should only be callable by the community multisig.

src/external/governance/GovernanceContract.sol
Zitzak: Nvm, just saw further down below that you addresses this :)

src/external/governance/GovernanceContract.sol
Zitzak: Agree with @marvinkruse. I can't think of a way how this would enable the "team" to "backdoor" something that the community wouldn't want

src/factories/ModuleFactory.sol
```diff
@@ -76,18 +79,21 @@ contract ModuleFactory is
     //--------------------------------------------------------------------------
     // Storage
 
+    /// @inheritdoc IModuleFactory
+    address public governanceContract; //@note changeable?
```
Zitzak: If we do make it changeable, who would be authorized to do so? Seems to me that this should also be managed by the community multisig, otherwise its a backdoor.

src/factories/ModuleFactory.sol
```diff
@@ -67,7 +67,10 @@ contract ModuleFactory is
     ///         {IInverterBeacon} instance.
```
Zitzak: Comment doesn't reflect the modification

test/e2e/ModuleTest_Template.txt
```diff
@@ -1,61 +0,0 @@
-// SPDX-License-Identifier: LGPL-3.0-only
```
Zitzak: What was this file used for? Just curious

script/factories/DeployModuleFactory.s.sol
```diff
@@ -23,23 +23,32 @@ contract DeployModuleFactory is Script {
     function run() external returns (address) {
         // Read deployment settings from environment variables.
 
+        address governaceContract = vm.envAddress("COVERNANCE_CONTRACT_ADDRESS");
         address forwarder = vm.envAddress("FORWARDER_ADDRESS");
         // Check settings.
 
+        require(
+            governaceContract != address(0),
+            "DeployOrchestratorFactory: Missing env variable: forwarder"
```
Zitzak:
```suggestion
            "DeployOrchestratorFactory: Missing env variable: governance contract"
```

Zitzak (changes requested):

src/external/governance/GovernanceContract.sol
marvinkruse: The idea of the bigger multisig (aka community multisig) is more that it contains people that are not us or directly benefitting from the decisions like in this case. So ideally, it would, for example, be people from other projects that either advise us or are friends with us, but more importantly, do have a reputation to lose if they would just agree to negative decisions (in either direction) and want to protect our community from any harm. Does that make sense?

src/external/governance/GovernanceContract.sol
marvinkruse: Probably a good idea to have a minimum of something like 48 hours or something. As only the community multisig can change this value, we can't change this ourselves and just "take over".

src/factories/ModuleFactory.sol
FHieser: Through the planned upgradeability of the contract this can be added later, so i wont include it for now due to timeconstraints, but would be nice to have a decision on this @marvinkruse

test/e2e/ModuleTest_Template.txt
FHieser: Yeah my bad
I think initially this was used as an copy paste example of how to structure e2e tests,
but in my cleanup rage I deleted it. 😅
I reverted the change, but I would argue, that with our repo still constantly changing this might be something that we might not think of when we make changes to the e2e tests in general
@0xNuggan any opinion on this, because you created it?

src/external/governance/Governor.sol
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    modifier validTimelockPeriod(uint amt) {
+        if (amt < 48 hours) {
+            revert Governance__InvalidTimelockPeriod(amt);
+        }
+        _;
+    }
+
+    modifier accessableBeacon(address target) {
```
marvinkruse: Sometimes you used "accessible" and sometimes "accessable". Wikipedia lists "accessable" as
> Misspelling of accessible

So maybe let's just eliminate the "accessable" for "accessible" wherever it was used?

src/external/governance/Governor.sol
```diff
@@ -0,0 +1,333 @@
...
+
+    /// @inheritdoc IGovernor
+    function triggerUpgradeBeaconWithTimelock(address beacon)
+        external
+        onlyCommunityOrTeamMultisig
+        accessableBeacon(beacon)
+        upgradeProcessAlreadyStarted(beacon)
+        timelockPeriodExceeded(beacon)
+    {
+        //set timelock to inaktive
```
marvinkruse: "inaktive" -> "inactive"

src/external/governance/Governor.sol
```diff
@@ -0,0 +1,333 @@
...
+import {IGovernor} from "src/external/governance/IGovernor.sol";
+import {
+    InverterBeacon,
+    IInverterBeacon
+} from "src/factories/beacon/InverterBeacon.sol";
+
+// External Dependencies
+import {ERC165} from "@oz/utils/introspection/ERC165.sol";
+import {Initializable} from "@oz-up/proxy/utils/Initializable.sol";
+import {AccessControl} from "@oz/access/AccessControl.sol";
```
marvinkruse: This contract is upgradeable; shouldn't this be the `@oz-up` version? The way the roles are stored internally in the `AccessControlUpgradeable` contract is handled differently in the upgradeable version.

src/external/governance/Governor.sol
```diff
@@ -0,0 +1,333 @@
...
+import {Ownable} from "@oz/access/Ownable.sol";
+import {Ownable2Step} from "@oz/access/Ownable2Step.sol";
```
marvinkruse, line 16: This contract is upgradeable; shouldn't this be the @oz-up version? Same argument as with the `AccessControl`, this has a different handling of the storage slot of the owner.

src/external/governance/Governor.sol
```diff
@@ -0,0 +1,333 @@
...
+import {
+    InverterBeacon,
+    IInverterBeacon
+} from "src/factories/beacon/InverterBeacon.sol";
+
+// External Dependencies
+import {ERC165} from "@oz/utils/introspection/ERC165.sol";
+import {Initializable} from "@oz-up/proxy/utils/Initializable.sol";
+import {AccessControl} from "@oz/access/AccessControl.sol";
+import {Ownable} from "@oz/access/Ownable.sol";
```
marvinkruse: Is this required? As far as I can tell, it's only used for the `target.call(abi.encodeCall(Ownable.owner, ()))` call. As `Ownable2Step` is based on `Ownable` it should just work, and if not - the `owner()` function is probably the same anyways so can just use that one (same function signature iirc)?

script/external/DeployGovernor.s.sol
```diff
@@ -0,0 +1,83 @@
...
+                )
+            );
+            gov.init(communityMultisig, teamMultisig, timelockPeriod);
+        }
+
+        vm.stopBroadcast();
+
+        // Log the deployed Governor address.
+        console2.log(
+            "Deployment of Governor implementatio at address", address(gov)
```
marvinkruse: "implementatio" -> "implementation"

script/factories/DeployModuleFactory.s.sol
```diff
@@ -23,23 +23,32 @@ contract DeployModuleFactory is Script {
     function run() external returns (address) {
         // Read deployment settings from environment variables.
 
+        address governor = vm.envAddress("COVERNANCE_CONTRACT_ADDRESS");
```
marvinkruse: "**C**OVERNANCE_CONTRACT_ADDRESS" -> "**G**OVERNANCE_CONTRACT_ADDRESS"

dev.env
```diff
@@ -41,7 +37,11 @@ export ABO_PRIVATE_KEY=0x47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c
...
 
-# These are just basic Addresses fro testing
+# These are just basic Addresses for testing
 export DEPLOYMENT_ORCHESTRATOR_FACTORY_TARGET=0x0000000000000000000000000000000000000001
 export DEPLOYMENT_ORCHESTRATOR_FACTORY_MODULE_FACTORY=0x0000000000000000000000000000000000000002
-export FORWARDER_ADDRESS=0x0000000000000000000000000000000000000003
\ No newline at end of file
+export COMMUNITY_MULTISIG_ADDRESS=0x0000000000000000000000000000000000000003
+export TEAM_MULTISIG_ADDRESS=0x0000000000000000000000000000000000000004
+export COVERNANCE_CONTRACT_ADDRESS=0x0000000000000000000000000000000000000005
```
marvinkruse: "**C**OVERNANCE_CONTRACT_ADDRESS" -> "**G**OVERNANCE_CONTRACT_ADDRESS"

marvinkruse: Also, Forwarder is just "FORWARDER_ADDRESS" and not "FORWARDER_CONTRACT_ADDRESS", so maybe this one should just be "GOVERNOR_ADDRESS"?

src/factories/ModuleFactory.sol
```diff
@@ -76,18 +80,21 @@ contract ModuleFactory is
...
+
     /// @dev Mapping of metadata identifier to {IInverterBeacon} instance.
     /// @dev MetadataLib.identifier(metadata) => {IInverterBeacon}
     mapping(bytes32 => IInverterBeacon) private _beacons;
 
     //--------------------------------------------------------------------------
     // Constructor
 
-    constructor(address _trustedForwarder)
+    constructor(address newGovernor, address _trustedForwarder)
```
marvinkruse: Wouldn't `_governor` be more consistent compared to `_trustedForwarder`? Or does `_trustedForwarder` have the underscore because it's from an underlying contract?

marvinkruse (changes requested):
Looks good to me (besides my small comments), really good work!

Feel free to just resolve any of my comments that you address/don't think make sense, so we can progress with getting this merged.

src/factories/ModuleFactory.sol
marvinkruse: Oh sorry, just saw this now. It's fine to keep it like it is implemented now (no changeability). The governor is upgradeable, that's fine.

dev.env
Zitzak: +1 also on using same name as contract, i.e. Governor

src/external/governance/IGovernor.sol
```diff
@@ -25,25 +21,25 @@ interface IGovernanceContract {
     // Errors
 
     /// @notice The given address is invalid
-    error InvalidAddress(address adr);
+    error Governance__InvalidAddress(address adr);
```
Zitzak: I assume you've changed these before changing the name of the contract :D They all have governance instead of governor

```suggestion
    error Governor__InvalidAddress(address adr);
```

Zitzak: This comment is meant for all the errors

src/factories/beacon/InverterBeacon.sol
```diff
@@ -35,13 +35,6 @@ contract InverterBeacon is IInverterBeacon, ERC165, Ownable2Step {
     //--------------------------------------------------------------------------
     // Modifiers
 
-    modifier validNewMinorVersion(uint newMinorVersion) {
```
Zitzak, line 38: Why did this modifier get removed? Will this check be done somewhere else? Would be curious about what's the thought behind this

Zitzak (changes requested):

src/external/governance/Governor.sol
FHieser: It threw me an error when I used Ownable2Step

Member "owner" not found or not visible after argument-dependent lookup in type(contract Ownable2Step).
   --> src/external/governance/Governor.sol:319:56:
    |
319 |         (success, result) = target.call(abi.encodeCall(Ownable2Step.owner, ()));

I mean I can always go with manual function signature, but I like the type safety

FHieser: Yeah I went with encodeWithSignature

src/factories/beacon/InverterBeacon.sol
FHieser: https://github.com/InverterNetwork/inverter-contracts/issues/432

Nuggan mentioned it in this issue, so I thought I would tackle it in one go

FHieser: Also it was needed when I used the upgrade function in the beacon constructor, although there might have been ways around that, now that I think about it

src/external/governance/Governor.sol
marvinkruse: Oh, sorry, I thought that this would just work. Either way is fine for me 😅 Encoding it is a good idea as well!

src/factories/beacon/InverterBeacon.sol
marvinkruse: Oh, I noticed this in my scripts PR as well and fixed it there as well 😂 (allowing 0 version, but only if it's the first one that is set in the beacon). But that's on me, I will go through merging hell and see what happens there once we get there.

script/factories/DeployModuleFactory.s.sol
```diff
@@ -23,7 +23,7 @@ contract DeployModuleFactory is Script {
     function run() external returns (address) {
         // Read deployment settings from environment variables.
 
-        address governor = vm.envAddress("COVERNANCE_CONTRACT_ADDRESS");
+        address governor = vm.envAddress("GOVERNANT_ADDRESS");
```
marvinkruse, line 26: "GOVERNANT_ADDRESS" should be "GOVERNOR_ADDRESS" - right?

dev.env
```diff
@@ -42,6 +42,6 @@ export DEPLOYMENT_ORCHESTRATOR_FACTORY_TARGET=0x00000000000000000000000000000000
 export DEPLOYMENT_ORCHESTRATOR_FACTORY_MODULE_FACTORY=0x0000000000000000000000000000000000000002
 export COMMUNITY_MULTISIG_ADDRESS=0x0000000000000000000000000000000000000003
 export TEAM_MULTISIG_ADDRESS=0x0000000000000000000000000000000000000004
-export COVERNANCE_CONTRACT_ADDRESS=0x0000000000000000000000000000000000000005
+export GOVERNANT_ADDRESS=0x0000000000000000000000000000000000000005
```
marvinkruse: "GOVERNANT_ADDRESS" should be "GOVERNOR_ADDRESS" - right?

src/external/governance/Governor.sol
```diff
@@ -31,21 +30,21 @@ contract Governor is ERC165, IGovernor, Initializable, AccessControl {
 
     modifier validAddress(address adr) {
         if (adr == address(0)) {
-            revert Governance__InvalidAddress(adr);
+            revert Govenor__InvalidAddress(adr);
```
marvinkruse, line 33: "Govenor" -> "Governor"

src/external/governance/Governor.sol
```diff
@@ -31,21 +30,21 @@ contract Governor is ERC165, IGovernor, Initializable, AccessControl {
...
         }
         _;
     }
 
     modifier validTimelockPeriod(uint amt) {
         if (amt < 48 hours) {
-            revert Governance__InvalidTimelockPeriod(amt);
+            revert Govenor__InvalidTimelockPeriod(amt);
```
marvinkruse, line 40: "Govenor" -> "Governor"

src/external/governance/Governor.sol
```diff
@@ -31,21 +30,21 @@ contract Governor is ERC165, IGovernor, Initializable, AccessControl {
...
         }
         _;
     }
 
-    modifier accessableBeacon(address target) {
+    modifier accessibleBeacon(address target) {
         if (!isBeaconAccessible(target)) {
-            revert Governance__BeaconNotAccessible(target);
+            revert Govenor__BeaconNotAccessible(target);
```
marvinkruse, line 47: "Govenor" -> "Governor"

src/external/governance/Governor.sol
```diff
@@ -57,22 +56,22 @@ contract Governor is ERC165, IGovernor, Initializable, AccessControl {
             !hasRole(COMMUNITY_MULTISIG_ROLE, sender)
                 && !hasRole(TEAM_MULTISIG_ROLE, sender)
         ) {
-            revert Governance__OnlyCommunityOrTeamMultisig();
+            revert Govenor__OnlyCommunityOrTeamMultisig();
```
marvinkruse, line 59: "Govenor" -> "Governor"

src/external/governance/IGovernor.sol
```diff
@@ -21,25 +21,25 @@ interface IGovernor {
...
-    error Governance__TimelockPeriodNotExceeded();
+    error Govenor__TimelockPeriodNotExceeded();
 
     /// @notice This Upgrade process for this beacon has not yet been started
-    error Governance__UpgradeProcessNotStarted();
+    error Govenor__UpgradeProcessNotStarted();
 
     /// @notice The call to the target contract has failed
-    error Governance__CallToTargetContractFailed();
+    error Govenor__CallToTargetContractFailed();
```
marvinkruse: "Govenor" -> "Governor"

src/external/governance/Governor.sol
FHieser: Nope as this is just to reference the function signature and is not inherited

src/external/governance/Governor.sol
FHieser: yes your correct here

src/external/governance/IGovernor.sol
FHieser: aaaaaaaaaaaaaaaaaaaaah

marvinkruse (approved):
lgtm - really good work! 👏

PR #438, opened 2024-04-04
Chore: Refactor Contract Names, Directories and Function Names

Zitzak:
### Contract Naming
- `BancorVirtualSupplyBondingCurveFundingManager` -> `FM_BC_Bancor_Redeeming_VirtualIssuance_VirtualCollateral_v1`
- `ToposBondingCurveFundingManager` -> `FM_BC_BondingSurface_Repayer_Seizable_v1`
- Removed `FundingManager` out of the base contract. Example of this is:
  - `BondingCurveFundingManagerBase` -> `BondingCurveBase`
- Added the keyword `Base` to the virtual supply base contracts
### General Naming
- Removed all functions/events/comments related to Topos (except for testing comments).
- Changed `baseMultiplier` -> `basePriceMultiplier` within the BondingSurface FM and tests.
- Changed all references of the issuance tokens to include the word "issuance", so it matches the collateral
  - `virtualTokenSupply` -> `virtualIssuanceSupply`
- Adjusted comments here and there
### Directory structure
- Created sub-directories for the types (sub-categories) of Funding Manager, i.e. **bondingCurve** and **rebasing**
- Moved all base contracts for the sub-category of Funding Manager to a dedicated **abstracts** directory as a child of the sub-category directory
- Moved all Interfaces for the sub-category of Funding Manager to a dedicated **interfaces** directory as a child of the sub-category directory
- Moved all Libraries used in the different Funding Managers into `/src/lib/`
- Within the testing directory:
  - Created sub-directory based on the the sub-category of Funding Manager
  - Kept the same structure of having the base contracts in the **abstracts** directory

## **PLEASE NOTE:** No logic has been changed in this PR


Zitzak:
I have created a Linear issue to track this PR only for @marvinkruse, but thought I add @FHieser and @0xNuggan to the PR to see how I re-structured the code

src/modules/lib/FixedPointMathLib.sol
```diff
@@ -8,58 +8,56 @@ library FixedPointMathLib {
                             COMMON BASE UNITS
     //////////////////////////////////////////////////////////////*/
 
-    uint256 internal constant YAD = 1e8;
-    uint256 internal constant WAD = 1e18;
-    uint256 internal constant RAY = 1e27;
-    uint256 internal constant RAD = 1e45;
+    uint internal constant YAD = 1e8;
```
0xNuggan, line 11: Can you give a bit more context on these changes?

0xNuggan:
Left one comment, rest looks good to me. Thanks for making the changes, and also renaming the Token to Issuance in the Bancor Contracts :pray:

src/modules/lib/FixedPointMathLib.sol
Zitzak: Hmm I think that my formatter did that, I didn't go in and changed this manually. Do you think I should revert it so it's equal to the original Solmate version?

0xNuggan: I don't think it's worth it... was just curious.

marvinkruse: It is interesting to me though, as we now have an inconsistency here. Most of the codebase uses `uint256`, we now have some parts using `uint`. I know they are equivalent in terms of compilation, but I find it weird that the formatter seemed to convert some to `uint` and some not? Something isn't right here.
But we can also explore this in a different PR at a later time.

marvinkruse (approved):
Looks good to me as well, just the formatted thing, but we can handle that in a different PR.

It looks a bit weird now with just some of the contracts having the new naming scheme, but we will probably fix that later this week?

PR #439, opened 2024-04-04
Feat: Add Cliff to Streaming Payments

marvinkruse:
**Current Status**: ⚠️ Not done; I added the cliff to the implementation; it has not been tested yet, though. Currently, all payments are added with 0 cliff in the tests.

--------

The StreamingPaymentProcessor needs to be able to offer a proper cliff functionality. Payment shall be created with a `cliff` variable, which contains a duration. Any streaming may only start once this duration has passed (after the payment was created/started). However, the **actual vesting shall already happen in the meantime**. 

Example:
* **Payment A**: Start at 1pm, Ends 5pm, no Cliff, 100 Tokens
  * At 2pm 25 Tokens can be claimed (25% vested)
  * At 3pm 50 Tokens can be claimed  (50% vested)
  * At 4pm 75 Tokens can be claimed  (75% vested)
  * At 5pm 100 Tokens can be claimed (100% vested)
* **Payment B**: Start at 1pm, Ends 5pm, 3 Hour Cliff, 100 Tokens
  * At 2pm 0 Tokens can be claimed (cliff active, 25% vested)
  * At 3pm 0 Tokens can be claimed (cliff active, 50% vested)
  * At 4pm 75 Tokens can be claimed (cliff ended, 75% vested)
  * At 5pm 100 Tokens can be claimed (100% vested) 

Example screenshot


src/modules/paymentProcessor/StreamingPaymentProcessor.sol
```diff
@@ -744,8 +767,10 @@ contract StreamingPaymentProcessor is Module, IStreamingPaymentProcessor {
 
     /// @notice validate uint start input.
     /// @param _start uint to validate.
+    /// @param _cliff uint to validate.
+    /// @param _end uint to validate.
     /// @return True if uint is valid.
-    function validStart(uint _start) internal view returns (bool) {
-        return !(_start < block.timestamp || _start >= type(uint).max);
+    function validTimes(uint _start, uint _cliff, uint _end) internal view returns (bool) {
+        return !(_start >= type(uint).max && _start + _cliff > _end);
```
marvinkruse, line 774: I couldn't see a reason why (a) streaming payments couldn't start in the past (so far they needed to be `block.timestamp`or greater) and (b) streaming payments were allowed to have an end date that was smaller than the start date (I can see a reason for `start == end` for instant payments via the streaming processor, but not for `end < start`.

I adapted the implementation so that `start + cliff <= end` - with `start` not needing to be greater than or equal to `block.timestamp`. I can at least see good reason for vesting schemes to start vesting retroactively, especially with cliffs.

Is there anything I am missing @FHieser on why this was implemented this way? Any reason why my change shouldn't be done?

There was this `testProcessPaymentsWorksForEndTimeThatIsPlacedBeforeStartTime` which I found and why I came to this question. This test seems to work with the new implementation but I assume that is not how it was intended. I will continue looking into this tomorrow :D Any input on why things have been done this way would be great :)

src/modules/paymentProcessor/StreamingPaymentProcessor.sol
```diff
@@ -744,8 +767,14 @@ contract StreamingPaymentProcessor is Module, IStreamingPaymentProcessor {
...
+    /// @param _end uint to validate.
     /// @return True if uint is valid.
-    function validStart(uint _start) internal view returns (bool) {
-        return !(_start < block.timestamp || _start >= type(uint).max);
+    function validTimes(uint _start, uint _cliff, uint _end)
+        internal
+        view
+        returns (bool)
+    {
+        return !(_start >= type(uint).max && _start + _cliff > _end);
```
marvinkruse, line 779: I couldn't see a reason why (a) streaming payments couldn't start in the past (so far they needed to be `block.timestamp`or greater, see the [old implementation here](https://github.com/InverterNetwork/inverter-contracts/blob/a7e2b7f574086fff3b8da664815f30054ff8baee/src/modules/paymentProcessor/StreamingPaymentProcessor.sol#L749)) and (b) streaming payments were allowed to have an end date that was smaller than the start date (I can see a reason for `start == end` for instant payments via the streaming processor, but not for `end < start`. 

I adapted the implementation so that `start + cliff <= end` - with `start` not needing to be greater than or equal to `block.timestamp`. I can at least see good reason for vesting schemes to start vesting retroactively, especially with cliffs.

Is there anything I am missing @FHieser on why this was implemented this way? Any reason why my change shouldn't be done?

There was this `testProcessPaymentsWorksForEndTimeThatIsPlacedBeforeStartTime` which I found and why I came to this question. This test seems to work with the new implementation but I assume that is not how it was intended. I will continue looking into this tomorrow :D Any input on why things have been done this way would be great :)

PR #440, opened 2024-04-10
chore: fix review comment

Zitzak:


src/modules/paymentProcessor/StreamingPaymentProcessor.sol
```diff
@@ -513,7 +513,10 @@ contract StreamingPaymentProcessor is Module, IStreamingPaymentProcessor {
         uint stillReleasable =
             releasableForSpecificWalletId(client, paymentReceiver, walletId);
         //In case there is still something to be released
-        if (stillReleasable > 0) {
+        if (
+            vestings[client][paymentReceiver][walletId]._salary
+                - vestings[client][paymentReceiver][walletId]._released > 0
+        ) {
             //Let PaymentClient know that the amount is not needed to be stored anymore
             IERC20PaymentClient(client).amountPaid(stillReleasable);
```
FHieser: I think this doesnt fully do it.
If I remember correctly stillReleasable uses vestedAmountForSpecificWalletId() which just checks what is releasable until that moment of time, not the full time period, which we actually want.
So the solution like this is probably more accurate:

uint remainingReleasable =  vestings[client][paymentReceiver][walletId]._salary
                - vestings[client][paymentReceiver][walletId]._released;
if (
           remainingReleasable > 0
        ) {
IERC20PaymentClient(client).amountPaid(remainingReleasable); 
}

also stillReleasable can be remove

FHieser (changes requested):

FHieser (approved):

PR #441, opened 2024-04-10
Feat: External Issuance Token for the Bonding Curve and Restricted mint and burn functions

0xNuggan:
This PR adapts the Bonding Curve contracts to prepare for the Paribu and Q/Acc  usecases. 

### New contracts:
- **ERC20Issuance**: ERC20 contract with exposed mint/burn functions and a supply cap. Only a specific address (separate from the owner) is allowed to call the mint and burn fucntions. Any mint which would end up with a total supply above the supply cap reverts.
- **RestrictedBancorVirtualSupplyBondingCurveFundingManager**: Inherits from the BancorVirtualSupplyBondingCurveFundingManager, but defines a CURVE_INTERACTION_ROLE and overrides the minting functions to be gated by it 

### Changes to existing contracts
- **BondingCurveFundingManagerBase**: Is not an ERC20 anymore, but stores a reference to an ERC20Issuance contract. The mint and burn functions call that contract
- **BancorVirtualSuppplyBondingCurveFundingManager**: Control for valid decimals when setting the issuance token


src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -112,26 +114,21 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
...
-        );
+        (_issuanceToken, bondingCurveProperties, _acceptedToken) =
+            abi.decode(configData, (address, BondingCurveProperties, address));
 
-        __ERC20_init(
-            string(abi.encodePacked(issuanceToken.name)),
-            string(abi.encodePacked(issuanceToken.symbol))
-        );
+        // Set issuance token. This also caches the decimals
+        _setIssuanceToken(_issuanceToken);
```
Zitzak, line 124: This change seems to introduce functionality such that the FM_BC **only** works with an external ERC20. I wasn't aware of the fact that we remove the issuance asset from the FM all together across all FM_BC, and not only for Paribu.

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
```diff
@@ -11,12 +11,8 @@ import {IBondingCurveFundingManagerBase} from
 import {IOrchestrator} from "src/orchestrator/IOrchestrator.sol";
 
 // External Interfaces
+import {ERC20Issuance} from "./ParibuChanges/ERC20Issuance.sol";
```
Zitzak, line 14: Curious why you went with the contract and not the interface here? Also it might be good to use the absolute path instead of relative, as that is done for the other imports as well

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -471,23 +477,25 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
...
-    /// @dev Sets the number of decimals for the token.
+    /// @dev Sets the issuance token for the Bonding Curve Funding Manager.
     /// This function overrides the internal function set in BondingCurveFundingManagerBase, adding
     /// an input validation specific for the Bancor Formula utilizing implementation, after which
-    /// it updates the `tokenDecimals` state variable.
-    /// @param _decimals The number of decimals to set for the token.
-    function _setTokenDecimals(uint8 _decimals)
+    /// it updates the `issuanceToken` state variable and caches the decimals as `issuanceTokenDecimals`.
+    /// @param _issuanceToken The token which will be issued by the Bonding Curve.
+    function _setIssuanceToken(address _issuanceToken)
```
Zitzak, line 485: Should we enable a way to set a new token after deployment so the token can be migrated? We are also enabling a way for the minter in the ERC20Issuance to be updated.

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/ERC20Issuance.sol
```diff
@@ -0,0 +1,85 @@
...
+
+// External Dependencies
+import {ERC20Upgradeable} from "@oz-up/token/ERC20/ERC20Upgradeable.sol";
+import {ContextUpgradeable} from "@oz-up/utils/ContextUpgradeable.sol";
+import {OwnableUpgradeable} from "@oz-up/access/OwnableUpgradeable.sol";
+
+// External Libraries
+import {SafeERC20} from "@oz/token/ERC20/utils/SafeERC20.sol";
+
+/// @title Bonding Curve Funding Manager Base Contract.
```
Zitzak, line 15: Title is still from the base contract

Zitzak: Also the @notice & @dev parts need updating

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/ERC20Issuance.sol
```diff
@@ -0,0 +1,85 @@
...
+        _burn(_from, _amount);
+    }
+
+    //------------------------------------------------------------------------------------
+    // Internal Functions
+
+    /// @notice Sets the address of the minter.
+    /// @param _minter The address of the minter.
+    function _setMinter(address _minter) internal {
+        allowedMinter = _minter;
```
Zitzak, line 82: Should, and can we, check that the minter is a FM_BC module? Might be overkill and not required. Also maybe a check for a zero address makes sense. However both of these are not critical considering time constraints

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/ERC20Issuance.sol
```diff
@@ -0,0 +1,85 @@
...
+/// @notice This contract enables the base functionalities for issuing tokens along a bonding curve.
+/// @dev The contract implements functionalties for:
+///         - opening and closing the issuance of tokens.
+///         - setting and subtracting of fees, expressed in BPS and subtracted from the collateral.
+///         - calculating the issuance amount by means of an abstract function to be implemented in
+///             the downstream contract.
+contract ERC20Issuance is
+    IERC20Issuance,
+    ERC20Upgradeable,
+    OwnableUpgradeable
```
Zitzak, line 26: Who will be the owner of this contract, i.e. allowed to set the minter?

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/IERC20Issuance.sol
```diff
@@ -0,0 +1,46 @@
...
+import {Ownable} from "@oz/access/Ownable.sol";
+
+// External Dependencies
+import {ERC20Upgradeable} from "@oz-up/token/ERC20/ERC20Upgradeable.sol";
+import {ContextUpgradeable} from "@oz-up/utils/ContextUpgradeable.sol";
+
+// External Libraries
+import {SafeERC20} from "@oz/token/ERC20/utils/SafeERC20.sol";
+
+/// @title Bonding Curve Funding Manager Base Contract.
```
Zitzak, line 16: Natspec needs updating

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/RestrictedBancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -0,0 +1,109 @@
...
+pragma solidity 0.8.23;
+
+// Internal Dependencies
+
+import {
+    BancorVirtualSupplyBondingCurveFundingManager,
+    IBancorVirtualSupplyBondingCurveFundingManager
+} from
+    "src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol";
+
```
Zitzak, line 11: Natspec missing

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/RestrictedBancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -0,0 +1,109 @@
...
+        onlyModuleRole(CURVE_INTERACTION_ROLE)
+    {
+        _virtualSellOrder(_msgSender(), _depositAmount, _minAmountOut);
+    }
+
+    /// @inheritdoc IBancorVirtualSupplyBondingCurveFundingManager
+    function mintIssuanceTokenTo(address _receiver, uint _amount)
+        external
+        override(BancorVirtualSupplyBondingCurveFundingManager)
+        onlyModuleRole(CURVE_INTERACTION_ROLE)
```
Zitzak, line 104: Shouldn't this be a different role? In this way, why would addresses with this role buy tokens, if they can just mint it

test/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.t.sol
```diff
@@ -122,19 +126,17 @@ contract BancorVirtualSupplyBondingCurveFundingManagerTest is ModuleTest {
...
 
         BancorFormula bancorFormula = new BancorFormula();
         formula = address(bancorFormula);
 
-        issuanceToken.name = bytes32(abi.encodePacked(NAME));
-        issuanceToken.symbol = bytes32(abi.encodePacked(SYMBOL));
-        issuanceToken.decimals = uint8(DECIMALS);
+        issuanceToken = new ERC20IssuanceMock();
+        issuanceToken.init(NAME, SYMBOL, type(uint).max, DECIMALS);
+        issuanceToken.setMinter(address(bondingCurveFundingManager));
```
Zitzak: The `bondingCurveFundingManager` isn't set yet. How does this work?

Zitzak: Ahh you set it 2 times. Why is this necessary? Guessing just for safety measure so that the owner doesn't mint?

test/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.t.sol
```diff
@@ -1516,44 +1515,114 @@ contract BancorVirtualSupplyBondingCurveFundingManagerTest is ModuleTest {
...
                 .selector
         );
         // No authentication since it's an internal function exposed by the mock contract
-        bondingCurveFundingManager.call_setDecimals(_newDecimals);
+        bondingCurveFundingManager.call_setIssuanceToken(
+            address(newIssuanceToken)
+        );
     }
 
     function testSetDecimals_FailsIfLowerThanCollateralDecimals(
```
Zitzak: This function can be removed. Duplicate to the one below

Zitzak (changes requested):

test/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.t.sol
0xNuggan: oops

test/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.t.sol
0xNuggan: I think it actually sets it to 0 the first time since bondingCurveFundingManager is declared but not initialized at that point. I've removed it. Although setting the minter to 0 on setup would indeed be the way to prevent the owner from minting straight away.

0xNuggan:
The content in this PR has been moved to #451

PR #442, opened 2024-04-10
SC-240: Protocol Fee Collection from FM BC

Zitzak:
## What has been done
The change introduces the possibility to collect protocol fee during a buy order as well as the sell order, on both the collateral and issuance token. The fee percentage get pulled from the tax manager

## Todo
- Uncomment function calls to the tax manager to get fee percentages
- Uncomment getting the address of the tax manager
- Testing implementation


src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -275,37 +275,64 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
...
         external
-        view
         returns (uint mintAmount)
     {
         if (_depositAmount == 0) {
             revert
                 BancorVirtualSupplyBondingCurveFundingManager__InvalidDepositAmount(
             );
         }
-        if (buyFee > 0) {
```
FHieser, line 285: You could do the zero check instead in the _calculateNetAmountAndFee:
if(_feePct==0){
return _transactionAmount}
.....

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -275,37 +275,64 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
...
-            (_depositAmount, /* feeAmount */ ) =
-                _calculateNetAmountAndFee(_depositAmount, buyFee);
-        }
-        return _issueTokensFormulaWrapper(_depositAmount);
+        // Get protocol fee percentages
+        (uint collateralBuyFeePercentage, uint issuanceBuyFeePercentage) =
```
FHieser: naming: protocolBuyFee?

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
```diff
@@ -253,6 +265,67 @@ abstract contract BondingCurveFundingManagerBase is
...
+        uint feeAmount;
+        // get net and fee amount
+        (netAmount, feeAmount) =
+            _calculateNetAmountAndFee(_transactionAmount, _feePct);
+
+        // mint fee amount
+        // TODO: Get address of where the tokens need to be minted to
+        // _mint(address(__Module_orchestrator.taxMan()), feeAmount);
+
+        // TODO: Should we add event to protocol fee collection?
```
FHieser: Good question
If we do this then maybe even when the protocol takes their share
Seems kind of overkill, but on the other side makes it easy for users to see what amounts of tax have been taken

0xNuggan: I'd be in favor. Just to be able to easily reconstruct the process. We'll probably be interested in tracking all the amounts sent to the TaxMan and why

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
```diff
@@ -253,6 +280,98 @@ abstract contract BondingCurveFundingManagerBase is
...
 
+    /// @dev Returns the collateral and issuance fee percentage retrieved from the tax manager for
+    ///     buy operations
+    /// @return collateralTreasury The address the protocol fee in collateral should be sent to
+    /// @return issuanceTreasury The address the protocol fee in issuance should be sent to
+    /// @return collateralBuyFeePercentage The percentage fee to be collected from the collateral
+    ///     token being deposited for minting issuance, expressed in BPS
+    /// @return issuanceBuyFeePercentage The percentage fee to be collected from the issuance token
+    ///     being minted, expressed in BPS
+    function _getBuyFeesAndTreasuryAddresses()
```
0xNuggan, line 291: This may be nitpicking, but can't this function be combined with `_getSellFeesAndTreasuryAddresses()` into one that receives the function selector (f.ex. `_getFunctionFeesAndTreasuryAddresses()`)? That way, we avoid duplicated code and open the door for inheriting contracts to query fees of additional functions

0xNuggan: Maybe it could even be moved to the base Module contract (although I'm not sure if I'm getting ahead of myself there)

Zitzak: You're right, that would be more clean

Zitzak:
This PR will be closed, as all the changed introduced here are integrated, and further developed upon in PR #453

PR #444, opened 2024-04-11
 Add batched role grant/revoke

0xNuggan:
This PR adds batched role granting and revoking for global roles and module roles. It also cleans up a bit the AuthorizerMock


marvinkruse (approved):
lgtm

PR #449, opened 2024-04-12
Merge temp_dev into dev

marvinkruse:


test/modules/fundingManager/RebasingFundingManager.t.sol
```diff
@@ -46,7 +46,7 @@ contract RebasingFundingManagerTest is ModuleTest {
     uint internal constant DEPOSIT_CAP = 100_000_000e18;
 
     // Other constants.
-    uint8 private constant DECIMALS = 18;
+    //uint8 private constant DECIMALS = 18;  //Question : Is there any reason to hardcode the decimals for this fundingmanager to 18?
```
FHieser: Still a question left
Sounds like a zitzak question xD

test/modules/fundingManager/RebasingFundingManager.t.sol
```diff
@@ -104,7 +104,7 @@ contract RebasingFundingManagerTest is ModuleTest {
     // Tests: Initialization
 
     function testInit() public override(ModuleTest) {
-        assertEq(fundingManager.decimals(), DECIMALS);
+        //assertEq(fundingManager.decimals(), DECIMALS);
```
FHieser: also outcommented this code
probably in connection with the comment above

FHieser (approved):
Except for the 2 comments i made this looks pretty good

test/modules/fundingManager/RebasingFundingManager.t.sol
marvinkruse: As the token of the `RebasingFundingManager` is initialized with the same decimals as the `orchestratorToken` aka `collateralToken`, it is not hardcoded to 18 in the contract. So I don't see why it should be hardcoded here.

I will change the test to get the correct decimal value from the underlying token instead.

test/modules/fundingManager/RebasingFundingManager.t.sol
marvinkruse: Same as above, yeah.

PR #450, opened 2024-04-15
Feat: Add KPI Rewarder Module

0xNuggan:


0xNuggan:
Finally ready for review. Notes:
- The CI is failing because of the fork tests to Sepolia. What can we do @marvinkruse ?
- Staking bug still in. Maybe the auditors will have insights

src/modules/logicModule/KPIRewarder.sol
```diff
@@ -0,0 +1,376 @@
...
+        ) {
+            revert Module__KPIRewarder__ModuleCannotUseStakingTokenAsBond();
+        }
+
+        // Make sure that we are targeting an existing KPI
+        if (KPICounter == 0 || targetKPI >= KPICounter) {
+            revert Module__KPIRewarder__InvalidKPINumber();
+        }
+
+        // Question: what kind of checks should or can we implement on the data side?
```
FHieser: What is happening on the Oracle side with the data field?
My first thought was that it is a "put a message here" field, but it being bytes32 it seems a bit small for that

.github/workflows/ci.yml
```diff
@@ -4,13 +4,13 @@ on:
...
       - main
       - dev
-      - temp_dev
   pull_request:
     branches:
       - '*'
 
 env:
   FOUNDRY_PROFILE: ci
+  SEPOLIA_RPC_URL: ${{ secrets.SEPOLIA_RPC }}
```
FHieser, line 13: nice
Are you already testing it?

src/modules/logicModule/KPIRewarder.sol
```diff
@@ -0,0 +1,376 @@
...
+        // Technically the value mentioned inside "data" (and posted publicly) wouldn't need to be the same as assertedValue...
+
+        // =====================================================================
+        // Staking Queue Management
+
+        for (uint i = 0; i < stakingQueue.length; i++) {
```
FHieser: We talked about the queue already. It is practically a POC, but at some point we might wanna do a rework if this becomes a problem

test/e2e/logicModules/KPIRewarderLifecycle.t.sol
```diff
@@ -0,0 +1,527 @@
...
+
+// Mocks
+import {ERC20Mock} from "test/utils/mocks/ERC20Mock.sol";
+import {ERC20} from "@oz/token/ERC20/ERC20.sol";
+import {IERC20Metadata} from "@oz/token/ERC20/extensions/IERC20Metadata.sol";
+
+/*
+Fork testing necessary:
+
+forge test --match-contract KPIRewarderLifecycle -vvv --fork-url $SEPOLIA_RPC_URL
```
FHieser: How exactly are you doing this in the CICD?
Can you isolate a single test to be forked tested or are all tests running now via fork?

test/e2e/logicModules/KPIRewarderLifecycle.t.sol
```diff
@@ -0,0 +1,527 @@
...
+        }
+
+        // Check final invariants
+
+        assertEq(
+            rewardToken.balanceOf(address(fundingManager)),
+            (REWARD_DEPOSIT_AMOUNT - totalDistributed)
+        );
+
+        // TODO: Fix Staking precision bug
```
FHieser, line 306: Still TODO

test/e2e/logicModules/KPIRewarderLifecycle.t.sol
```diff
@@ -0,0 +1,527 @@
...
+        assertApproxEqAbs(
+            totalDistributed, totalExpectedRewardsDistributed, 1e6
+        );
+
+        /*
+        console.log("Final Rewards Distributed: ", totalDistributed);
```
FHieser, line 312: Still logged

FHieser (commented):
From an implementation standpoint it looks good to me
The testing is the interesting thing to look at
Asked some questions in the comments you might wanna check that out @0xNuggan

.github/workflows/ci.yml
0xNuggan: There's an issue with the CI running it since there are still conflicts with the base branch, but once they're resolved it should run like magic

src/modules/logicModule/KPIRewarder.sol
0xNuggan: The short answer is: it depends, so there's nothing we can really do at this point. The long answer: The data is supposed to be anything that will indicate the UMA checkers how to confirm/reject the assertion. But: they don't have a standard format for that yet, and basically hardcode it for each project that posts there. Some put an IPFS link, others may just send a number to fill out a fixed template. So this part is somewhat up in the air for the time we do some kind of frontend integration, where we will need to revisit the way we construct the bytemessage

src/modules/logicModule/KPIRewarder.sol
0xNuggan: Yeah, I agree. But for now it seems the most straightforward way to go with.

test/e2e/logicModules/KPIRewarderLifecycle.t.sol
0xNuggan: Do we have a general policy about keeping/removing outcommented logs in tests @marvinkruse ? I personally like keeping them so when getting back to the code in the future I just need to uncomment a couple of lines and can get some extra information about contract state. But I see it can be considered bloat, so I'm happy to remove it if preferred. (Of course I'm referring to well written logs, not just "log("run arrived here") )

test/e2e/logicModules/KPIRewarderLifecycle.t.sol
0xNuggan: The code in line 123 is the one that sets up the fork, so all other tests run as usual. I updated the comment though, since it was outdated

test/e2e/logicModules/KPIRewarderLifecycle.t.sol
0xNuggan: somebody chose violence today.

FHieser: https://imgflip.com/i/8nytnw

test/e2e/logicModules/KPIRewarderLifecycle.t.sol
FHieser: I just marked it as it wasnt a style we used before
If I wanted to check for a specific value I always just did it with asserts (at least in the e2e tests)
But I can understand if the usage of an external contract is a bit tricky here. But technically all the internal stuff of the staking should ideally be pre conceptualized without any further interactions with the external contracts, no?
I dont have a hard opinion on this I was just wondering

test/e2e/logicModules/KPIRewarderLifecycle.t.sol
FHieser: :see_no_evil:

test/e2e/logicModules/KPIRewarderLifecycle.t.sol
marvinkruse: Yeah, the cleaner the better, but I'm fine for us to keep outcommented code/logs in there if there is a high chance of us needing it again. Otherwise we should probably remove it. But it's a good reminder to keep these, for when we are revisiting.

FHieser (approved):

PR #451, opened 2024-04-15
Feat: External Issuance Token for the Bonding Curve and Restricted mint and burn functions (new)

0xNuggan:
This PR adapts the open one to the new dev branch, preparing the Bonding Curve contracts for the Paribu and Q/Acc usecases.

### New contracts:
- **ERC20Issuance**: ERC20 contract with exposed mint/burn functions and a supply cap. Only a specific address (separate from the owner) is allowed to call the mint and burn fucntions. Any mint which would end up with a total supply above the supply cap reverts.
- **RestrictedBancorVirtualSupplyBondingCurveFundingManager**: Inherits from the BancorVirtualSupplyBondingCurveFundingManager, but defines a CURVE_INTERACTION_ROLE and overrides the minting functions to be gated by it 

### Changes to existing contracts
- **BondingCurveFundingManagerBase**: Is not an ERC20 anymore, but stores a reference to an ERC20Issuance contract. The mint and burn functions call that contract
- **BancorVirtualSuppplyBondingCurveFundingManager**: Control for valid decimals when setting the issuance token

There are also some bugfixes, specifically in the _boundForDecimalConversion() test helper


0xNuggan:
PR is ready for review. Since the renaming hasn't happened yet, I've left the new stuff in the ParibuChanges folder, even if that is not supposed to be final

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
```diff
@@ -246,61 +233,30 @@ abstract contract BondingCurveFundingManagerBase is
...
-    }
-
-    function _contextSuffixLength()
-        internal
-        view
-        virtual
-        override(ContextUpgradeable, ERC2771ContextUpgradeable)
-        returns (uint)
-    {
-        return ERC2771ContextUpgradeable._contextSuffixLength();
```
marvinkruse, line 304: What's this doing here, wasn't this a different PR? Or is this some merge magic leftover?

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/ERC20Issuance.sol
```diff
@@ -0,0 +1,93 @@
...
+        _burn(_from, _amount);
+    }
+
+    //------------------------------------------------------------------------------------
+    // Internal Functions
+
+    /// @notice Sets the address of the minter.
+    /// @param _minter The address of the minter.
+    function _setMinter(address _minter) internal {
+        // Note to @Zitzak: Since the token should work independently, I wouldn't control that the minter is a module. Also, setting the minter to zero may be useful in some cases.
```
marvinkruse, line 90: Leftover comment?

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
0xNuggan: No, it's just because that since we don't inherit from ERC20Upgradeable anymore, we don't have a conflict and can just take the implementations from the Module class above

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/ERC20Issuance.sol
0xNuggan: No, I actually want to double check with @Zitzak since he commented it in the last PR. But my opinion is we shouldn't check for it to keep it generalized. That way a Token can completely unbind itself from the workflow it was deployed in

marvinkruse: That makes more sense, thank you for the explainer!

marvinkruse: Let's just not forget about it 😁

src/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.sol
marvinkruse: Yeah that makes sense, gotcha! Thank you!

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/ERC20Issuance.sol
```diff
@@ -0,0 +1,93 @@
...
+/// @title ERC20Issuance Token
+/// @author Inverter Network.
+/// @notice This contract creates an ERC20 token with the ability to mint and burn tokens and a supply cap.
+/// @dev The contract implements functionalties for:
+///         - Whitelisted mint and burn
+///         - A supply cap
+///         - Flexible decimals in initialization
+contract ERC20Issuance is
+    IERC20Issuance,
+    ERC20Upgradeable,
```
Zitzak, line 24: Why are we chosing for upgreadable contract. As there is an owner, and also a minter that can be set it seems we have all the flexibility we need

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/ERC20Issuance.sol
Zitzak: So I agree with the comment, that if it should be functioning as a standalone, then we should restrict it, and setting the minter to address zero and then revoking ownership could be a way to "default" the ERC20 back to normal

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -111,26 +117,35 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
...
-        __ERC20_init(
-            string(abi.encodePacked(issuanceToken.name)),
-            string(abi.encodePacked(issuanceToken.symbol))
+        ERC20Issuance _issuanceToken = new ERC20Issuance();
+        _issuanceToken.init(
+            string(abi.encodePacked(issuanceTokenData.name)),
+            string(abi.encodePacked(issuanceTokenData.symbol)),
+            issuanceTokenData.decimals,
+            issuanceTokenData.maxSupply,
+            tokenAdmin
```
Zitzak: Shouldn't we just set the minter per default to the FM_BC implementation? Otherwise we will still need another TX for updating it, and it would introduce a risk of the admin just minting tokens. What are you thoughts on that? @0xNuggan

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
```diff
@@ -319,9 +338,21 @@ contract BancorVirtualSupplyBondingCurveFundingManager is
     //--------------------------------------------------------------------------
     // OnlyOrchestrator Functions
 
+    /// @inheritdoc IFundingManager
```
Zitzak, line 342: The comment has been removed in the IFundingManager. Maybe add a comment here specifically stating that this function has been left with no restriction?

Zitzak:
We also talked about that Paribu will do a big mint through the BC to initiate the supply. I don't remember how big the amount was, but if large enough, then it might be good to add a test with these amounts to be sure that no over/under flow is happening with the Bancor Formula.
@0xNuggan

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
0xNuggan: We can do that too. I'll add an initialMinter parameter so stuff can be configured out of the box (we need to indicate the addresses explicitly because of factory/initializer/etc shenaningans)

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
0xNuggan: Not sure what you mean here?

0xNuggan:
Good point on the buy-in, will update the lifecycle test to reflect it

src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol
Zitzak: Hmm not sure myself. I thought you had removed the comment of this function within the `IFundingManager`, but isn't the case at all. So just ignore this comment

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/ERC20Issuance.sol
marvinkruse: Has this question been answered already or is this still an open question?

marvinkruse: In any case I tend to agree with @Zitzak  though, as upgradeable tokens are always a very delicate topic. If there is no really good reason to have it like this, I think it shouldn't be upgradeable?

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/ERC20Issuance.sol
```diff
@@ -0,0 +1,94 @@
...
+    OwnableUpgradeable
+{
+    // State Variables
+    address public allowedMinter;
+    uint public MAX_SUPPLY;
```
marvinkruse, line 29: If we switch to a non-upgradeable version, this should probably be `immutable`?

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/IERC20Issuance.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity 0.8.23;
+
+interface IERC20Issuance {
+    //--------------------------------------------------------------------------
+    // Events
+
+    /// @notice Emitted when the minter is set.
+    event minterSet(address indexed minter);
```
marvinkruse: I think we are capitalizing event names, right?

marvinkruse (commented):
Left 1-2 comments, but overall I'm happy! 🙏

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/ERC20Issuance.sol
0xNuggan: We kind of agreed on leaving it upgradeable just in case we need to do some extra changes. The only thing I can think of right now is something like replacing minter address with minter whitelist, but maybe something else comes up...

0xNuggan:
All changes have been moved to #455 , so closing this

PR #452, opened 2024-04-16
Feat: Only Allow Adding of Modules Registered in the Factory

marvinkruse:
This adds a check to the `ModuleManager`, resulting in only being able to add modules to an Orchestrator if they are properly registered through our factory.

⚠️ The tests of this are not done yet, work in progress.


Zitzak:
This PR will be closed, as all the changes have been integrated into the rebased version at PR #455

marvinkruse:
Handled in #483 - see you there!

PR #453, opened 2024-04-18
Feat: Add Protocol Fee (old/unrebased)

FHieser:
Solves SC-18

### DIFF
--- a/script/deployment/DeploymentScript.s.sol
+++ b/script/deployment/DeploymentScript.s.sol
@@ -18,6 +18,7 @@ import {DeployBountyManager} from
     "script/modules/logicModule/DeployBountyManager.s.sol";
 
 import {DeployGovernor} from "script/external/DeployGovernor.s.sol";
+import {DeployFeeManager} from "script/external/DeployFeeManager.s.sol";
 import {DeployTransactionForwarder} from
     "script/external/DeployTransactionForwarder.s.sol";
 import {DeployOrchestrator} from "script/orchestrator/DeployOrchestrator.s.sol";
@@ -80,6 +81,8 @@ contract DeploymentScript is Script {
         new DeployTransactionForwarder();
     //Governor
     DeployGovernor deployGovernor = new DeployGovernor();
+    //FeeManager
+    DeployFeeManager deployFeeManager = new DeployFeeManager();
 
     //Beacon
     DeployAndSetUpBeacon deployAndSetUpBeacon = new DeployAndSetUpBeacon();
@@ -138,6 +141,9 @@ contract DeploymentScript is Script {
 
     //Governor
     address governor;
+
+    //FeeManager
+    address feeManager;
 
     //TransactionForwarder
     address forwarder;
@@ -256,6 +262,23 @@ contract DeploymentScript is Script {
         console2.log(
             "-----------------------------------------------------------------------------"
         );
+
+        console2.log(
+            "-----------------------------------------------------------------------------"
+        );
+        console2.log("Fee Manager \n");
+
+        feeManager = deployFeeManager.run(
+            address(governor), //owner
+            communityMultisig, //@note This is not the real target is it?
+            100, //Colleteral Fee 1%
+            100 //Issuance Fee 1%
+        ); //@todo late dependency injection of feemanager into Governor
+
+        console2.log(
+            "-----------------------------------------------------------------------------"
+        );
+
         console2.log("Deploy orchestrator implementation \n");
         //Orchestrator
         orchestrator = deployOrchestrator.run();
--- a/script/deployment/DeploymentScript.s.sol
+++ b/script/deployment/DeploymentScript.s.sol
@@ -273,7 +273,7 @@ contract DeploymentScript is Script {
             communityMultisig, //@note This is not the real target is it?
             100, //Colleteral Fee 1%
             100 //Issuance Fee 1%
-        ); //@todo late dependency injection of feemanager into Governor
+        );
 
         console2.log(
             "-----------------------------------------------------------------------------"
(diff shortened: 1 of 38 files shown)


test/modules/fundingManager/bondingCurveFundingManager/RedeemingB_CurveFundingManagerBase.t.sol
```diff
@@ -491,7 +864,7 @@ contract RedeemingBondingCurveFundingManagerBaseTest is ModuleTest {
     //      - Mints collateral tokens to a seller and
     //      - Deposits them so they can later be sold.
     //      - Approves the BondingCurve contract to spend the receipt tokens
-    // @note This function assumes that we are using the Mock with a 0% buy fee, so the user will receive as many toknes as they deposit
+    //      - This function assumes that we are using the Mock with a 0% buy fee, so the user will receive as many toknes as they deposit
```
0xNuggan, line 486: typo

test/modules/fundingManager/bondingCurveFundingManager/RedeemingB_CurveFundingManagerBase.t.sol
```diff
@@ -364,6 +364,379 @@ contract RedeemingBondingCurveFundingManagerBaseTest is ModuleTest {
         );
     }
```
0xNuggan: Some comment explicitly indicating what the "case" is would be nice

0xNuggan:
One question: Hadn't we chosen a different name for TaxMan in the end?

marvinkruse:
> One question: Hadn't we chosen a different name for TaxMan in the end?

We did, I believe it was just an old commit still, he pushed it now with `FeeManager` 👍

script/deployment/HowToSetup.md
```diff
@@ -1,19 +1,12 @@
...
-5. Run the following command:
+1. Paste the values of the `dev.env` in your `.env` file:
+2. Adapt the values of the `.env` file to fit your deployment setup.
+3. Open a tab in your terminal
+4. Run the following command: `anvil`
+5. Open another tab in your terminal
+6. Run the following command:
    `forge script script/deployment/DeploymentScript.s.sol --fork-url http://localhost:8545/ --broadcast`
+7. Access the communityMultisig and call setFeeManager() in the governor contract. Use the address of the newly creatd
+   @note is this enough to properly represent that you have to set the feemanager seperately?
```
marvinkruse: Yeah, I'd say that's fine for now. Scripts and Docs will be overhauled in the next 1-2 weeks anyways!

src/external/fees/FeeManager.sol
```diff
@@ -0,0 +1,282 @@
...
+        external
+        onlyOwner
+        validAddress(_defaultProtocolTreasury)
+    {
+        defaultProtocolTreasury = _defaultProtocolTreasury;
+        emit DefaultProtocolTreasurySet(_defaultProtocolTreasury);
+    }
+
+    /// @inheritdoc IFeeManager
+    function setWorkflowTreasuries(address workflow, address treasury)
```
marvinkruse: As this just sets the on treasury of one workflow, should it be called `setWorkflowTreasury()` instead of `setWorkflowTreasuries()`?

src/external/governance/Governor.sol
```diff
@@ -152,7 +153,64 @@ contract Governor is ERC165, IGovernor, AccessControlUpgradeable {
...
+    function setFeeManagerIssuanceWorkflowFee(
+        address workflow,
+        address module,
+        bytes4 functionSelector,
+        bool set,
+        uint fee
+    ) external onlyRole(COMMUNITY_MULTISIG_ROLE) {
+        feeManager.setIssuanceWorkflowFee(
+            workflow, module, functionSelector, set, fee
+        );
```
marvinkruse: I will double-check this with the business side, as I can see arguments for both sides, but I assume that these might be functions that both multisigs, but especially the `TEAM_MULTISIG_ROLE`, should be able to access on its own. This is something that will (or at least can) occur pretty often (i.e. whenever a new workflow is deployed that we might have a custom deal with) and I don't see us asking the community multisig all the time to confirm a new fee value.

Maybe (just spitballing) a good solution here would be that the community multisig can set certain boundaries for the team multisig (i.e., such that we can set the fee to a value from 0-5%, but community multisig can do 0-100%) so that we can't just silently set the fee to a super high value without community confirmation.

If that takes too long for now I would probably just:
* Set it to allow any multisig for now
* Revisit this after the first audit handover (i.e. in a week) with an update, where we introduce this - and have it covered in the subsequent audits/contests

What do you think @FHieser?

script/deployment/DeploymentScript.s.sol
```diff
@@ -256,6 +262,23 @@ contract DeploymentScript is Script {
...
+
+        console2.log(
+            "-----------------------------------------------------------------------------"
+        );
+        console2.log("Fee Manager \n");
+
+        feeManager = deployFeeManager.run(
+            address(governor), //owner
+            communityMultisig, //@note This is not the real target is it?
+            100, //Colleteral Fee 1%
```
0xNuggan, line 274: typo: collateral

test/modules/fundingManager/bondingCurveFundingManager/BondingCurveFundingManagerBase.t.sol
```diff
@@ -268,11 +287,37 @@ contract BondingCurveFundingManagerBaseTest is ModuleTest {
...
+        }
+
+        if (_workflowFee != 0) {
+            vm.prank(owner_address);
+            bondingCurveFundingManager.setBuyFee(_workflowFee);
+        }
 
         address buyer = makeAddr("buyer");
+
+        // vm.assume(buyer != address(bondingCurveFundingManager) && buyer != address(tr) )
```
0xNuggan, line 320: seems like this is a useful check?

FHieser:
Moved to #456

PR #454, opened 2024-04-19
Chore/rename contracts

Zitzak:
# What has been done
## For each contract
- Name contract
- Filename contract
- Natspec contract
- Comments contract (if applicable)
- Imports contract
- Update directory location (if applicable)
- Naam interface
- Filenaam interface
- Imports interface
- Custom error string in interface
- Comments interface (if applicable)
- Update directory location (if applicable)
- Name deployment script (if applicable)
- Filename deployment script (if applicable)
- Name test
- Filename test
- Mocknames (if applicable)
## Additional Notes
- Added `interfaces` folder whenever I project more contracts to be created within the specific category/directory
- Moved `beacons` from the `factories` folder to src/. Same for the test files
    - I followed the example of how the scripts directory was structured. Made more sense to me
- I’ve appended `base` to each abstract contract to signal they are base and not implementation contracts
  - Exception for the rebasing token base contracts
- Aligned all the imports for the contracts so the order is the same
  1. Internal Interfaces
  2. Internal Dependencies
  3. External Interfaces
  4. External Dependencies
- Updated all the error for each contract to match new naming
- Add remapping for the main module categories
- Removed the wording `FundingManager` from the base contracts
- Moved libraries within the module directory which where in sub-directories to the shared lib folder
- All of current Logic Modules (LM) are Payment Clients (PC), and they all use ERC20, which is why I moved them together in the same directory
- 

[Link to google sheet where I kept track of name changes and categorisations can be found here](https://docs.google.com/spreadsheets/d/1mAzfiNHalK0CNJEMYIraWuffmmEDgTvRz1TdcNSEkvA/edit#gid=0)

## Open Questions
- We have a singleVoteGovernor and a normal Governor now, which are two completely different use cases. Maybe rename to InverterGovernor?
- Do we have events where we need it to reconstruct the state off-chain?
- How to structure the directory of the Payment Processor
- What to do about the Utils folder?

### Governor
Failed amount: 3 times
``` 
Encountered 1 failing test in test/external/Governor_v1.t.sol:GovernorV1Test
[FAIL. Reason: call reverted as expected, but without data; counterexample: calldata=0x443a5b3800000000000000000000000000000000000000000000000000000000000003980000000000000000000000004e59b44847b379578588920ca78fbf26c0b4956c args=[920, 0x4e59b44847b379578588920cA78FbF26c0B4956C]] testAccessibleBeacon(uint256,address) (runs: 2, μ: 26877, ~: 26877)
```
### SingleVoteGovernor
Failed amount: 1 time
```
Encountered 1 failing test in test/modules/SingleVoteGovernor.t.sol:SingleVoteGovernorTest
[FAIL. Reason: call did not revert as expected; counterexample: calldata=0xcaf8d2b30000000000000000000000007fa9385be102ac3eac297483dd6233d62b3e1496 args=[0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496]] testOnlyGovernanceIsAuthorized(address) (runs: 102, μ: 38682, ~: 38682)
```


script/proxies/UpgradeInverterBeacon_v1.sol
FHieser, line 1: you changed it so that this is not a script anymore
Add the .s. in the naming

script/utils/InvestableWorkstreams/01-DeployInvestableWorkstream.s.sol
```diff
@@ -6,22 +6,21 @@ import "forge-std/Test.sol";
...
+    ILM_PC_Bounty_v1
+} from "@lm_pc/ERC20PaymentClient/LM_PC_Bounty_v1.sol";
 import {
-    BancorVirtualSupplyBondingCurveFundingManager,
-    IBancorVirtualSupplyBondingCurveFundingManager
-} from
-    "src/modules/fundingManager/bondingCurveFundingManager/BancorVirtualSupplyBondingCurveFundingManager.sol";
+    FM_BC_Bancor_Redeeming_VirtualSupply_v1,
+    IFM_BC_Bancor_Redeeming_VirtualSupply_v1
+} from "@fm/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol";
```
FHieser, line 20: Nice :D

src/external/governance/Governor_v1.sol
```diff
@@ -15,37 +15,51 @@ import {AccessControlUpgradeable} from
     "@oz-up/access/AccessControlUpgradeable.sol";
 import {Ownable2Step} from "@oz/access/Ownable2Step.sol";
 
-contract Governor is ERC165, IGovernor, AccessControlUpgradeable {
+/**
+ * @title   Governor_v1: Governor Contract V1 for the Inverter Network.
+ *
+ * @notice  This contract manages upgrades to beacon contracts through role-based permissions,
```
FHieser: I would reorder this
* @notice This contract manages various administrative functions that can be executed only by specified multisig addresses. It supports upgrades to beacon contracts through role-based permissions, enabling a timelocked upgrade process and emergency procedures.

src/external/governance/interfaces/IGovernor_v1.sol
```diff
@@ -21,25 +21,25 @@ interface IGovernor {
     // Errors
 
     /// @notice The given address is invalid
-    error Governor__InvalidAddress(address adr);
+    error Governor_v1__InvalidAddress(address adr);
```
FHieser: I think changing the naming of the error messages to include the version might be debatable
Its probably due to the Replace functionality you used, but I think it comes with pros and cons
Due to the time constraints I would recommend to keep it that way for now, but I think we might wanna talk about this

src/factories/ModuleFactory_v1.sol
```diff
@@ -1,42 +1,44 @@
...
+import {Context, Ownable} from "@oz/access/Ownable.sol";
 
 /**
- * @title Module Factory
+ * @title   ModuleFactory_v1: Module Factory v1 for the Inverter Network.
  *
- * @dev An owned factory for deploying modules.
+ * @notice  Enables the creation and registration of Inverter Modules,
+ *          facilitating the deployment of module instances linked to specific beacons.
+ *          Provides management functionality to govern module behavior and interactions.
```
FHieser: ~~Provides management functionality to govern module behavior and interactions.~~
-> Allows for configuration of modules starting state via provided deployment data.

src/factories/interfaces/IModuleFactory_v1.sol
```diff
@@ -42,7 +42,7 @@ interface IModuleFactory {
     //--------------------------------------------------------------------------
     // Functions
 
-    /// @notice Returns the governor contract address
+    /// @notice Returns the governor_v1 contract address
```
FHieser, line 45: ~~governor_v1~~
-> governor

src/factories/interfaces/IOrchestratorFactory_v1.sol
```diff
@@ -1,29 +1,29 @@
...
 
     /// @notice The orchestrator owner is address(0)
-    error OrchestratorFactory__OrchestratorOwnerIsInvalid();
+    error OrchestratorFactory_v1__OrchestratorOwnerIsInvalid();
 
     //--------------------------------------------------------------------------
     // Events
 
-    /// @notice Event emitted when a new orchestrator is created.
+    /// @notice Event emitted when a new orchestrator_v1 is created.
```
FHieser: seems like the naming in the @notice tends to be <contract>_v1
Seems kind of weird to me?
@Zitzak You decide. We can probably can also tackle this later
In case we want to change this, the easiest way to find this is probably regex:
Something like this->  /// @notice*_v1

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -4,18 +4,18 @@ pragma solidity ^0.8.0;
...
     "@oz/access/extensions/IAccessControlEnumerable.sol";
 
-interface IAuthorizer is IAccessControlEnumerable {
+interface IAuthorizer_v1 is IAccessControlEnumerable {
     //--------------------------------------------------------------------------
     // Errors
 
     /// @notice The function is only callable by an active Module.
-    error Module__RoleAuthorizer__NotActiveModule(address module);
+    error Module__Authorizer_v1__NotActiveModule(address module);
```
FHieser: Why was this Modul__ somthing in the first place?

src/modules/base/Module_v1.sol
```diff
@@ -12,31 +17,26 @@ import {ERC165} from "@oz/utils/introspection/ERC165.sol";
...
-import {IAuthorizer} from "src/modules/authorizer/IAuthorizer.sol";
-import {IAuthorizer} from "src/modules/authorizer/IAuthorizer.sol";
-
 /**
- * @title Module
+ * @title   Module_v1
  *
- * @dev The base contract for modules.
+ * @dev     The base contract for modules.
  *
```
FHieser, line 27: This Contract is the basic building block for all Modules in the Inverter Network.
It contains references to other contracts, modifier for access restriction, metadata to identify the module type as well as utility functions for general module interactions.

src/modules/base/Module_v1.sol
```diff
@@ -12,31 +17,26 @@ import {ERC165} from "@oz/utils/introspection/ERC165.sol";
...
- *      Each module is identified via a unique identifier based on its major
- *      version, title, and url given in the metadata.
+ *          Each module is identified via a unique identifier based on its major
+ *          version, title, and url given in the metadata.
  *
- *      Using proxy contracts, e.g. beacons, enables globally updating module
- *      instances when its minor version changes, but supports differentiating
- *      otherwise equal modules with different major versions.
+ *          Using proxy contracts, e.g. beacons, enables globally updating module
```
FHieser: ~~Using proxy contracts, e.g. beacons, enables globally updating module~~
~~instances when its minor version changes, but supports differentiating~~
~~otherwise equal modules with different major versions.~~

src/modules/utils/SingleVoteGovernor_v1.sol
```diff
@@ -1,21 +1,23 @@
...
-import {Module, IModule} from "src/modules/base/Module.sol";
+import {Module_v1, IModule_v1} from "src/modules/base/Module_v1.sol";
 
-import {IOrchestrator} from "src/orchestrator/IOrchestrator.sol";
+import {IOrchestrator_v1} from
+    "src/orchestrator/interfaces/IOrchestrator_v1.sol";
 
-import {ISingleVoteGovernor} from "src/modules/utils/ISingleVoteGovernor.sol";
+import {ISingleVoteGovernor_v1} from
+    "src/modules/utils/interfaces/ISingleVoteGovernor_v1.sol";
```
FHieser: title notice and so forth missing

tasks/deployBaseContracts.sh
FHieser, line 1: Just out of Interest:
@InverterNetwork/smart-contract-devs what does this contract do?

src/orchestrator/Orchestrator_v1.sol
```diff
@@ -1,54 +1,56 @@
...
-
-import {IModuleManager} from "src/orchestrator/base/IModuleManager.sol";
-
 /**
- * @title Orchestrator
+ * @title   Orchestrator_v1: Orchestrator v1 for Inverter Network.
  *
- * @dev A new funding primitive to enable multiple actors within a decentralized
- *      network to collaborate on orchestrators.
+ * @dev     A new funding primitive to enable multiple actors within a decentralized
```
FHieser: This Contract is the center and connecting block of all Modules in a Inverter Network Workflow.
It contains references to the essential contracts that make up a workflow.
By inheriting the ModuleManager it allows for managing which modules make up the workflow.

src/orchestrator/abstracts/ModuleManagerBase_v1.sol
```diff
@@ -1,32 +1,33 @@
...
-// Interfaces
-import {IModuleManager} from "src/orchestrator/base/IModuleManager.sol";
-
 /**
- * @title Module Manager
+ * @title   ModuleManagerBase_v1: Module Manager Base v1 for Inverter Network
  *
- * @dev A contract to manage modules that can execute transactions via this
- *      contract and manage own role-based access control mechanisms.
+ * @dev     A contract to manage modules that can execute transactions via this
```
FHieser: A contract to manage Inverter Network modules.
It allows for adding and removing modules in a local registry for reference.
Additional functionality includes the execution of calls from this contract.

src/orchestrator/abstracts/ModuleManagerBase_v1.sol
```diff
@@ -1,32 +1,33 @@
...
+ *          contract and manage own role-based access control mechanisms.
```
FHieser: ~~A contract to manage modules that can execute transactions via this~~
~~contract and manage own role-based access control mechanisms.~~

~~The role-based access control mechanism is based on OpenZeppelin's~~
~~AccessControl contract. Each module has it's own access control context~~
~~which it is able to freely manage.~~

FHieser (changes requested):
First of:
Props to you @Zitzak 
This was huuuuge. Thanks for doing that.

I added some comments for the descriptions of the contract. Some of them were a bit outdated, but that wasnt your doing.

One thing that might take some time for me getting used to is probably the error naming stuff. Especially the adding of Module__ before every error i found a bit over the top. But i think for now this is fine, we can tackle this at a later point of time too.

tasks/deployBaseContracts.sh
marvinkruse: I have no idea and never used it, I use the scripts. @0xNuggan do you use this or could we get rid of it?

marvinkruse:
Some notes from my review:

---

**Payment Clients**
To me, the structure here feels like it's overengineered:
* `src/modules/logicModule/paymentClient/ERC20PaymentClient/LM_PC_Bounty_v1.sol`
* `src/modules/logicModule/paymentClient/ERC20PaymentClient/LM_PC_Recurring_v1.sol`
I understand these are all payment clients and we wanna show that maybe, even if I think the whole `PC`/paymentClient part is redundant in itself as it doesn't reeeaaallly matter, but why not just simplify it like this:
* `src/modules/logicModule/LM_PC_Bounty_v1.sol`

This tells you that it's a logic module and a payment client. There is no need to separate the "non-payment-client" logic modules, as we don't have any anyways for now. Even if we did, I think they could be in the same folder. Like so:
* `src/modules/logicModule/LM_PC_Bounty_v1.sol`
* `src/modules/logicModule/LM_PC_Recurring_v1.sol`
* `src/modules/logicModule/LM_IPFSManager_v1.sol`
* `src/modules/logicModule/LM_PC_NFTGenerator_v1.sol`

I feel like having two subfolders that both tell you it's a payment client is a bit much. I know we might need to distinguish between `ERC20PaymentClient` and `ERC721PaymentClient` or whatnot in the future, but I feel like having that in the contract name is enough for now. If we have many of them at some point, we can still change the folder structure in the future, as it doesn't matter as much as the contract name.

---

**RecurringPaymentManager**
I know I might be alone here, but I don't feel like this is named well with `LM_PC_Recurring_v1.sol`.
Recurring-what? I felt like `RecurringPaymentManager` was a good enough name. We want to capture the essence of the contracts. This one is about creating and managing recurring payments. 

---

**BountyManager**
To me, calling it just `LM_PC_Bounty_v1.sol` seems a bit too simplified, but might just be me. I liked `BountyManager` (so `LM_PC_BountyManager_v1.sol`) because the contract is not just about "a bounty" (whatever that may mean) but about creating and claiming these. I feel like `BountyManager` captured that quite well. Happy to be proven wrong, though 😁

---

Thoughts from you guys @Zitzak @FHieser @0xNuggan?

src/external/governance/interfaces/IGovernor_v1.sol
marvinkruse: Yeah, in general I would just change the name of the contract itself, not errors or events or whatever. If it's not possible time-wise I agree that we can just keep it like that and have a separate PR in the future to fix this, as it's not introducing any logic changes.

src/factories/interfaces/IOrchestratorFactory_v1.sol
marvinkruse: I would also not reference the version here, like I said above: basically only in the contract name. it's still the "orchestrator", doesn't really matter which version.

script/utils/InvestableWorkstreams/01-DeployInvestableWorkstream.s.sol
0xNuggan: yeah huge props on that

tasks/deployBaseContracts.sh
0xNuggan: Never used it. We can get rid of it

script/utils/InvestableWorkstreams/01-DeployInvestableWorkstream.s.sol
marvinkruse: so much better :D

tasks/deployBaseContracts.sh
FHieser: Yeah lets remove it

src/external/governance/interfaces/IGovernor_v1.sol
0xNuggan: Agree. Let's push it to the "non-logic-cleanup" bin :grin:

0xNuggan:
Agree with @marvinkruse  on his points

Zitzak:
@FHieser in regard to the `Module__` for error strings, I actually came across it is some interfaces and others not. So I decided to go this route, but apparently this was a practice which was done in the past already.

I don't think this PR will be added today anymore. Who votes that this should be changed again? React with thumbs up (for changing it by removing Module) or thumbs down (for rejecting the change and keep as is)
@marvinkruse @FHieser @0xNuggan

Zitzak:
In regard to the naming of the `LM_PC_Bounty_v1.sol` and `LM_PC_Recurring_v1.sol` contracts, I do agree that they don't look as nice as the full names. However if we decide on a naming convention then I think we should stick with it. Otherwise it will be back to personal preference again. Also, currently there is only one of these contract, but how to deal with it if more "Bounty" using LM will be written. We decided on the current scheme to make it future proof.

Lastly, the full name is still int the contract natspec.

Let's do another vote on this one with thumb up (for changing it back) or thumbs down (for rejecting changing it back)
@marvinkruse @0xNuggan @FHieser

src/external/governance/interfaces/IGovernor_v1.sol
Zitzak: So I actually did it on purpose, as I through it might be valuable information. Will change it back

Zitzak: Also just seeing that I haven't added `Module__` for these errors (see questions below)

src/modules/authorizer/IAuthorizer_v1.sol
Zitzak: See my message here
https://github.com/InverterNetwork/inverter-contracts/pull/454#issuecomment-2066899767

Zitzak:
I have removed the over-engineered directory structure with the PaymentClient as you suggested @marvinkruse

FHieser (approved):

FHieser:
In regard to the naming of the LM_PC_Bounty_v1.sol and LM_PC_Recurring_v1.sol 
Im fine with it, but I would recommend to change LM_PC_Recurring_v1 to LM_PC_RecurringPayment_v1

marvinkruse:
Shouldn't it then be "LM_PC_RecurringPayments_v1" instead of "LM_PC_RecurringPayment_v1"? As it's not just one payment?

FHieser (changes requested):
Rename LM_PC_Recurring_v1.sol to LM_PC_RecurringPayments_v1.sol

FHieser (approved):

marvinkruse:
Besides the `Bounty` contract it looks good to me, can be merged if everyone else is happy. ✅ 

----

Only thing I found weird was that in the comments at the top of the contracts it says 
```
/**
 * @title   TransactionForwarder_v1: Transaction Forwarder v1 for Inverter Network.
 *
 * @notice  This contract enables users to interact with smart contracts indirectly through
 *          a trusted forwarder. It supports meta transactions, allowing transactions to be
 *          sent by one party but signed and paid for by another. It also handles batch
 *          transactions (multi-call), facilitating complex, multi-step interactions within a single
 *          transaction.
 *
 * @dev     Integrates {ERC2771Forwarder} and Context to manage and relay meta transactions.
 *          It handles nonce management, signature verification, and ensures only trusted calls
 *          are forwarded.
 *
 * @author  Inverter Network.
 */
```

I don't consider the name or title of the contract to be `TransactionForwarder_v1` but to be `TransactionForwarder`. To me, having the file name there looks weird. Also, repeating it twice there doesn't really make much sense and it's not how I have seen it being used elsewhere. Also the addendum of "for the Inverter Network" is redundant. I know that this is not something you (@Zitzak) did, just something I noticed now when looking at it. Also the dot at the end of the author feels off to me, as it's not a sentence. My ideal NatSpec would look like this:

```
/**
 * @title   Meta-Transaction & Multicall Forwarder
 *
 * @notice  This contract enables users to interact with smart contracts indirectly through
 *          a trusted forwarder. It supports meta-transactions, allowing transactions to be
 *          sent by one party but signed and paid for by another. It also handles batch
 *          transactions (multicall), facilitating complex, multi-step interactions within a single
 *          transaction.
 *
 * @dev     Integrates {ERC2771Forwarder} and {Context} to manage and relay meta transactions.
 *          It handles nonce management, signature verification, and ensures only trusted calls
 *          are forwarded.
 *
 * @author  Inverter Network
 */
```

Important note here, though, that I also realize: this is not directly relevant for the renaming PR here, so we could do these types of changes once the audit handover is fully done and handle it in a different PR. I would just love a bit more consistency in the Natspec before the launch.

Zitzak:
I have implemented these contract natspecs, so I would say that they are part of this renaming PR.

I do agree @marvinkruse  that the `@title` could be done better, and happy to change this. However I will need some help in finding a good title for the contracts, where there hasn't been one in the past that I might repurpose

Zitzak:
> I don't consider the name or title of the contract to be TransactionForwarder_v1 but to be TransactionForwarder.

One more question, did you mean that the name of the contract itself should also be changed such to remove the version? @marvinkruse

marvinkruse:
In the interest of time, let's merge this one if everything else is good, as the NatSpec will take a bit of time to get in and go through everything. Can happen later. That's my view at least.

marvinkruse:
> > I don't consider the name or title of the contract to be TransactionForwarder_v1 but to be TransactionForwarder.
> 
> One more question, did you mean that the name of the contract itself should also be changed such to remove the version? @marvinkruse

Sorry, I wrote that quite confusing, was typing things up as I thought about it. I just meant the `_v1` part in the Contracts name **in the NatSpec**. Is that what your question was?

Zitzak:
Let me do a quick change to the contracts, and then we can merge it. Those would be
- remove the `.` at the end of author
- Use the old titles we had for the contracts that had them
- Remove the `for the inverter network`
@marvinkruse

marvinkruse:
If that doesn't take too much time then that's great, yeah!

Zitzak:
@marvinkruse  I've just put the name of the contract in the title of the contracts (except for the one you suggested for the Forwarder), as that seems to be as it was before.

FHieser (approved):
LGTM

PR #455, opened 2024-04-19
Feat: External Issuance Token for the Bonding Curve and Restricted Mint and Burn Functions [REBASED VERSION]

Zitzak:
## Please note
This PR has all the changes introduced in PR #451 as well as all the recently introduced changes concerning the renaming and re-org of the directory structure


Zitzak:
## TODO's 
- Update this branch with changes that might come from the PR #451 
- Update contract Natspec for the contracts added by Pable
- Update name of `RestrictedBancorVirtualSupply...` Implementation contract

## Questions
- Where should the `ERC20Issuance_v1` token live? I now added a new directory in `/FM/BC/` but that might be overkill. Where else?
-

0xNuggan:
Thanks a lot for this! :pray: 
To me it looks like the RestrictedBancor... name change got lost somewhere though?

About the token: I think keeping it in a folder there is a good solution for now, but I woul rename the folder to issuanceToken or something like that to make clearer what's inside. I also think it may be better to have the token Interface in there too? It seems wrong to me to have it together with the Bonding Curves

Zitzak:
So I haven't changed the name of the contract yet. I've tried initially during the rebasing but it created more problems. I've added it to the todos.

@0xNuggan your PR is also still review, so my ideas was to rebase one more time once it is approved, and then change the naming of the contract

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/IERC20Issuance.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
0xNuggan: This file is doubled with the one in interfaces/ (and contains less comments)

0xNuggan (changes requested):
Apart from this one thing, everything seems to be in place. Thanks for the hard job!!

The only open question in my mind is if we leave the issuance token upgradeable or not. What do you think? @marvinkruse @Zitzak 

Thumbs up is keep, down is change

marvinkruse:
I think we should keep the token **non-upgradeable** (for the end-users and their safety's sake) as long as there is no specific client need for it. The ownership is transferrable, and the minter can be replaced; otherwise, it's just a regular ERC20.

0xNuggan:
All clear, will update accordingly

src/modules/fundingManager/bondingCurveFundingManager/ParibuChanges/IERC20Issuance.sol
Zitzak: Ahh didn't see that there was still a directory that I didn't delete yet. Done it now

Zitzak:
Removed the Upgradability from the ERC20Issuance, jfyi :)) @0xNuggan

src/modules/fundingManager/bondingCurve/tokens/ERC20Issuance_v1.sol
```diff
@@ -6,26 +6,28 @@ import {IERC20Issuance_v1} from
     "@fm/bondingCurve/interfaces/IERC20Issuance_v1.sol";
 
 // External Dependencies
-import {ERC20Upgradeable} from "@oz-up/token/ERC20/ERC20Upgradeable.sol";
-import {ContextUpgradeable} from "@oz-up/utils/ContextUpgradeable.sol";
-import {OwnableUpgradeable} from "@oz-up/access/OwnableUpgradeable.sol";
+import {ERC20} from "@oz/token/ERC20/ERC20.sol";
+import {Context} from "@oz/utils/Context.sol";
+import {Ownable} from "@oz/access/Ownable.sol";
```
marvinkruse, line 11: We usually use `Ownable2Step`, but I like the argument of being able to "throw the keys away" by setting it to `address(0)` - especially for the token. So we should probably keep it like this, even if we usually would use the 2-step variant. Does someone have a different viewpoint?

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -129,8 +129,7 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
             (IssuanceToken, address, BondingCurveProperties, address)
         );
 
-        ERC20Issuance_v1 _issuanceToken = new ERC20Issuance_v1();
-        _issuanceToken.init(
+        ERC20Issuance_v1 _issuanceToken = new ERC20Issuance_v1(
             string(abi.encodePacked(issuanceTokenData.name)),
             string(abi.encodePacked(issuanceTokenData.symbol)),
```
marvinkruse: Why is this being converted from `string` (original type according to the `IssuanceToken` struct) to `string` again via `encodePacked`? Am I missing something, or is this just not doing anything?

marvinkruse (commented):
Good stuff, looks good otherwise!

src/modules/fundingManager/bondingCurve/tokens/ERC20Issuance_v1.sol
Zitzak: So that was indeed one of the points, although tbh I didn't know that wasn't possible with `Ownabel2Step`. So from my point we can leave it like this

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
Zitzak: Not sure tbh. I removed it

0xNuggan: That's probably on me. We were using bytes32 before (fabi brought it up in the sc-chat) and I changed it to string at some point. Must have missed that one when refactoring

src/modules/fundingManager/bondingCurve/tokens/ERC20Issuance_v1.sol
0xNuggan: Yeah I agree.

marvinkruse (approved):
thank you everyone - lgtm!

marvinkruse:
Don't forget to rename this PR to the "original" name (which was I think "_Feat: External Issuance Token for the Bonding Curve and Restricted Mint and Burn Functions_") before merging (or at least name the squash commit like that).

0xNuggan (approved):

PR #457, opened 2024-04-26
Feat: Add Functionality to Withdraw Project Collateral Fee

Zitzak:
## What has been done?
- Add functionality to withdraw project collateral fee
- Add changes implemented in the Topos branch
  - Move calculatePurchaseReturn to the base contracts
  - Move calculateSaleReturn to the base contract
  - Adapt test
 ## Implements
- https://linear.app/inverter/issue/SC-250/fm-bc-missing-features


0xNuggan:
About the failing test: It's checking address zero, but also [querying the default identifier](https://github.com/InverterNetwork/inverter-contracts/blob/05283a87479f7442efd749855c88a2fa5ee637d8/src/modules/logicModule/abstracts/oracleIntegrations/UMA_OptimisticOracleV3/OptimisticOracleIntegrator.sol#L135) and seeing if it fails. Short of an ERC165 interface check (which oov3 doesn't implement) it's the best we can do.

marvinkruse:
> About the failing test: It's checking address zero, but also [querying the default identifier](https://github.com/InverterNetwork/inverter-contracts/blob/05283a87479f7442efd749855c88a2fa5ee637d8/src/modules/logicModule/abstracts/oracleIntegrations/UMA_OptimisticOracleV3/OptimisticOracleIntegrator.sol#L135) and seeing if it fails. Short of an ERC165 interface check (which oov3 doesn't implement) it's the best we can do.

So whats the best way to fix that failing test?

0xNuggan:
I'm running it locally and it's not failing :sweat_smile: I'll check if the CI says anything more

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -298,41 +298,6 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
         );
     }
 
-    /// @inheritdoc IFM_BC_Bancor_Redeeming_VirtualSupply_v1
-    function calculatePurchaseReturn(uint _depositAmount)
```
0xNuggan, line 302: Can you remind me of the reason to move these two to the upstream contract? I'm not against it, just curious

Zitzak: The reason being is that we need this function in each implementation, and as far as I've seen with implementing both formula contracts, we use the internal issuance/redeem supply wrapper for calculating the return value. Next to this, we also need to deduct the fees from the return values. So to me it made sense to standardize this in the upstream contract, and leave the option for the implementation to change it if needed

0xNuggan (approved):
All looks good to me! Thanks for going through the merge conflicts

marvinkruse:
Could @FHieser maybe take a look as well, as he was involved with the whole fee implementation part? Or did he already?

Zitzak:
I've added @FHieser to the reviewers, although I think this fee withdraw functionality is a different feature than he was working on
@marvinkruse

FHieser (approved):
LGTM

PR #459, opened 2024-04-30
Fix: PaymentProcessor unclaimable bug

FHieser:


marvinkruse:
Didn't @FHieser mention that the actual fix wasn't contained here yet? I think I had a brief discussion with him in a DM where he mentioned that. That's why it was a draft still?
I did review the current state and found nothing really, just fyi.

FHieser:
@marvinkruse @0xNuggan now its ready for review

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -690,7 +697,13 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
             //Make sure to let paymentClient know that amount doesnt have to be stored anymore
             IERC20PaymentClientBase_v1(client).amountPaid(amount);
         } else {
-            unclaimableAmounts[client][paymentReceiver] += amount;
+            emit UnclaimableAmountAdded(
+                client, paymentReceiver, walletId, amount
+            );
+            //Adds the walletId to the array of unclaimable wallet ids
+            unclaimableWalletIds[client][paymentReceiver].push(walletId);
```
marvinkruse: Could there be a case where the `walletId` is already in that array? Would it be added twice then?

marvinkruse (commented):
Looks good to me, just that one question I had.

src/modules/paymentProcessor/PP_Streaming_v1.sol
FHieser: Yes its possible, but if you look at the _claimPreviouslyUnclaimable() function the loop in line line 734 deletes the unclaimableAmount for that specific walletId. That means if the walletid is contained multiple times, then at the second occurence the unclaimableamount is 0 and therfore this interaction basically doesnt matter anymore

FHieser: fuu

FHieser: I just looked at it again and noticed that the unclaimableAmount Lookup is dependenat on it beign contained only once so i have to change it

FHieser: Here is the fix

FHieser: Good find

marvinkruse: Thanks, good stuff!

marvinkruse (commented):
Fix to the thing I brought up looks good 👍

marvinkruse (approved):

PR #461, opened 2024-05-03
Feat: Implement workflow mutability policy

Zitzak:
## Implements
- https://linear.app/inverter/issue/SC-46/implement-workflow-mutability-policy

### DIFF
--- a/src/orchestrator/Orchestrator_v1.sol
+++ b/src/orchestrator/Orchestrator_v1.sol
@@ -207,7 +207,7 @@ contract Orchestrator_v1 is IOrchestrator_v1, ModuleManagerBase_v1 {
     // onlyOrchestratorOwner Functions
 
     /// @inheritdoc IOrchestrator_v1
-    function setAuthorizer(IAuthorizer_v1 authorizer_)
+    function initiateSetAuthorizerWithTimelock(IAuthorizer_v1 authorizer_)
         external
         onlyOrchestratorOwner
     {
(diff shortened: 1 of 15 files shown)


marvinkruse:
Just did a first pass over the implementation. Looks good to me! Will do a second pass within the next 30 minutes, but haven't found anything yet - just fyi.

marvinkruse (approved):
lgtm - good work! 🙏

src/orchestrator/Orchestrator_v1.sol
FHieser, line 1: Is it possible to remove the Authorizer via the ModuleManager directly and therefor brick it?
We might wanna change that in this PR too

FHieser (changes requested):

src/orchestrator/Orchestrator_v1.sol
Zitzak: Just for my understanding, you mean that we can remove the authorizer through the ModuleManager -> (initiate)removeModule and then we wouldn't be able to call the functions within the Orchestrator anymore that require authentification. Is that correct? @FHieser

marvinkruse: Good point @FHieser - thanks for adding this @Zitzak! 🙏

marvinkruse (approved):
Good stuff, latest commit looks fine to me!

FHieser (approved):

marvinkruse (approved):

PR #462, opened 2024-05-07
Feat: Enable Factory Upgradeability

marvinkruse:
This enables the factories to be upgradeable via our beacon pattern. Implementation is done, and tests as well as scripts should be fine as well.


FHieser (commented):
Looks good to me

0xNuggan (approved):
Looks good to me too

marvinkruse:
Updated the scripts to work now as well. Deployment via beacon of the factories, and the whole deployment run has ownership with the deployer first, and then transfers it. This is required as the ModuleFactory owner (governor) needs to be equal to the Beacon owner of a new module (governor) at the time of registering.

PR #465, opened 2024-05-08
Fix: Enable Batch Role Granting from Module Base

marvinkruse:
The implementation of batch-granting and -revoking of roles (in this case module roles), didn't include to enable this functionality in the `Module` base contract, from where it needs to be used (as the `Authorizer` function itself is `onlyModule`). 

This PR adds this functionality and extends the tests to cover it properly.

-- 

Solves SC-278


0xNuggan:
monkeyLookingAway

0xNuggan (approved):

PR #467, opened 2024-05-14
Fix: Reduce Bonding Curve (and other) Contract Sizes

0xNuggan:
## Description
Some contracts (particularly the Bonding Curve ones) had grown significantly beyond max contract size of 24.576 kb, making them undeployable. 

This PR implements a size reduction of ~25% to the BC contracts, leaving some margin in case there are changes in the upstream contracts that increase size. As a side effect, it also brings the other contracts below limit again.  

Closes [SC-272] .
## What has been done
While I see some of the streamlining as positive, this PR mostly makes the code uglier and more difficult to read. Sadly, this seems to be the reality of contract size reduction.

It has been achieved mostly by inlining functions that are seldom called and replacing local buffer variables with nested function calls.

There have also been some minor changes that affect functionality. I feel these were necessary and shouldn't be a big issue down the line, but would love to hear from anybody who has a different opinion .

### Framework Changes
- Reduced compiler optimization runs to 2000
### Functional Changes 
-  Remove init2 functionalities since they weren't being used anywhere
-  Make _openBuy and _openSell idempotent and inline them in the parent functions. As a consequence remove the corresponding revert errors.
- remove calculateNetAmountAndFee() since it wasn't being used anywhere.
- Generalize get[...]FeesAndTreasuryAddress() into one common function. 
### Structural Changes
- Create FM_BC_Tools library, that has (only) the _convertAmountToRequiredDecimal() function. NOTE: For some reason ([and against what the docs say](https://ethereum.org/en/developers/tutorials/downsizing-contracts-to-fight-the-contract-size-limit/)) making the function public increases contract size, while keeping it internal doesn't. Adding calculateNetSplitAndFees() (prime candidate) to the library also results in a size increase somehow.
- Consolidate errors between inheriting Bonding Curve contracts
- Externalize all modifiers into internal functions
- Externalize the [fee <= 100%] check
- Externalize reserve ratio validation 
### Inlining 
- Minimize buy and sell functions by calling their ...For() counterparts
- Compress _issueTokensFormulaWrapper by not storing local variables
- Inline _calculatePurchasereturn
- Inline _calculateSaleReturn
- Inline _getVirtualCollateralSupply
- Inline _getVirtualIssuanceSupply
- Inline _virtualBuyOrder 
- Inline _virtualSellOrder
- Inline _staticPricePPM
- Inline _withdrawProjectCollateralFee
## TL;DR 
Screenshot from 2024-05-14 16-43-11
IWonButAtWhatCost


script/utils/InvestableWorkstreams/01-DeployInvestableWorkstream.s.sol
```diff
@@ -52,18 +52,18 @@ contract SetupInvestableWorkstream is Test, DeploymentScript {
     // ========================================================================
     // BONDING CURVE PARAMETERS
 
-    string CURVE_TOKEN_NAME = "Conding Burve Token";
```
FHieser: lol

foundry.toml
```diff
@@ -7,7 +7,7 @@ fs_permissions = [{ access = "read-write", path = "./"}]
 # Compilation
 solc_version = "0.8.23"
 optimizer = true
-optimizer_runs = 5_000
+optimizer_runs = 2_000
```
FHieser, line 10: Why the reduce?

0xNuggan:
Since fully removing all the init2 related mentions to hasDependency, would make the diff too convoluted and complicate the review. I've moved the related cleanup to #471. Each PR should be easier to review independently. 

The intended merge path would be:
0) Approve both PRs 
1) #471 into this
2) This into dev

src/modules/base/Module_v1.sol
```diff
@@ -82,46 +77,27 @@ abstract contract Module_v1 is
...
-
-        if (
-            authorizer.hasRole(ownerRole, _msgSender())
-                || authorizer.hasRole(managerRole, _msgSender())
-        ) {
-            _;
-        } else {
-            revert Module__CallerNotAuthorized(ownerRole, _msgSender());
-        }
+        _checkOnlyOrchestratorOwnerOrManagerModifier();
```
FHieser, line 89: Out of interest: Is the newly created function really smaller than just using the way it was before?

src/modules/fundingManager/bondingCurve/FM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -52,47 +52,26 @@ contract FM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1 is
...
-    /// @inheritdoc FM_BC_Bancor_Redeeming_VirtualSupply_v1
-    /// @dev added role check
-    function buy(uint _depositAmount, uint _minAmountOut)
-        external
-        override(FM_BC_Bancor_Redeeming_VirtualSupply_v1)
-        buyingIsEnabled
-        onlyModuleRole(CURVE_INTERACTION_ROLE)
-    {
-        _virtualBuyOrder(_msgSender(), _depositAmount, _minAmountOut);
+        //_virtualBuyOrder(_receiver, _depositAmount, _minAmountOut);
```
FHieser: Outcommented

src/modules/fundingManager/bondingCurve/FM_BC_Tools.sol
```diff
@@ -0,0 +1,57 @@
...
+            return (_amount / conversionFactor);
+        } else {
+            // If the decimal of token is < required decimal, calculate conversion rate and
+            // return amount converted to required decimals
+            uint conversionFactor = (10 ** (_requiredDecimals - _tokenDecimals));
+            return (_amount * conversionFactor);
+        }
+    }
+
+    /*
```
FHieser, line 38: Outcommented

FHieser (approved):
All in all looks good to me
Chapeau to you @0xNuggan for going through this you detective.

I think I got the gist for most of the changes you made for the Bonding Curve, but I would recommend to put more eyes on this.
I will approve it, as the tests run through without problems.

PS:
Could you create Review Issues on Linear for the Future PRs?
It helps me to keep things on track :)

src/modules/base/Module_v1.sol
0xNuggan: Yes it is. Modifiers get copied in full into the function they appear in during compile time, so if you have a modifier that is used 3 times, you are actually duplicating the same code in all three functions separately. On the other hand a function call will redirect all three instances to the one place where you have stored the logic, avoiding all the additional storage. That's actually one of the reasons why some people suggest it's best practice to move away from modifiers entirely (vyper doesn't even implement them, f.ex).

Fun!

src/modules/fundingManager/bondingCurve/FM_BC_Tools.sol
0xNuggan: Yeah, this is the one function were I think having it here SHOULD be more efficient, but it actually isn't when I redirect the calls. I wanted to wait to see if anybody has insights on this, if not I'll just remove it before merging.

foundry.toml
0xNuggan: Optimizer runs make the code bigger on deployment, but more gas efficient on calls once deployed. Reducing it here was necessary to bring the contracts under the line.

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -125,28 +117,66 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
...
+
+        emit ProjectCollateralFeeWithdrawn(_receiver, _amount);
     }
 
     //--------------------------------------------------------------------------
     // Public Functions
 
     /// @notice Returns the address of the issuance token
-    function getIssuanceToken() public view virtual returns (address) {
+    function getIssuanceToken() external view virtual returns (address) {
```
Zitzak, line 179: Why are most function visibilities changes from external to public, but this one the other way around? Just curious

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -252,80 +280,36 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
...
+    ///     a specific operation
     /// @return collateralTreasury The address the protocol fee in collateral should be sent to
     /// @return issuanceTreasury The address the protocol fee in issuance should be sent to
-    /// @return collateralBuyFeePercentage The percentage fee to be collected from the collateral
-    ///     token being deposited for minting issuance, expressed in BPS
-    /// @return issuanceBuyFeePercentage The percentage fee to be collected from the issuance token
-    ///     being minted, expressed in BPS
-    function _getBuyFeesAndTreasuryAddresses()
+    /// @return collateralFeePercentage The percentage fee to be collected from the collateral
+    ///     token being redeemed, expressed in BPS
```
Zitzak:
```suggestion
    ///     token being deposited or redeemed, expressed in BPS
```

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -252,80 +280,36 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
...
+    /// @return issuanceFeePercentage The percentage fee to be collected from the issuance token
+    ///     being deposited, expressed in BPS
```
Zitzak:
```suggestion
    ///     being deposited or minted, expressed in BPS
```

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -2,38 +2,40 @@
...
-    "@fm/bondingCurve/abstracts/RedeemingBondingCurveBase_v1.sol";
-import {BondingCurveBase_v1} from
-    "@fm/bondingCurve/abstracts/BondingCurveBase_v1.sol";
-import {VirtualCollateralSupplyBase_v1} from
-    "@fm/bondingCurve/abstracts/VirtualCollateralSupplyBase_v1.sol";
-import {VirtualIssuanceSupplyBase_v1} from
-    "@fm/bondingCurve/abstracts/VirtualIssuanceSupplyBase_v1.sol";
+
+import {
+    IBondingCurveBase_v1,
```
Zitzak, line 15: So it is actually adding to storage to have multiple imports from the same contract, instead of just one which retrieves all the necessary dependencies?

src/modules/fundingManager/bondingCurve/FM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -52,47 +52,25 @@ contract FM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1 is
     /// @inheritdoc FM_BC_Bancor_Redeeming_VirtualSupply_v1
     /// @dev added role check
     function buyFor(address _receiver, uint _depositAmount, uint _minAmountOut)
-        external
-        override(FM_BC_Bancor_Redeeming_VirtualSupply_v1)
+        public
+        override
         validReceiver(_receiver)
         buyingIsEnabled
```
Zitzak: Would it be possible to remove the modifiers that are also on the `buyFor` function, as you're redirecting it through the `super` keyword?

Zitzak: If possible, same for the `sellFor` function

Zitzak (dismissed):
LGTM, only a few general questions

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
0xNuggan: Honestly, I think I did it just as cleanup. If you think it was better before I can change it back

src/modules/fundingManager/bondingCurve/FM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1.sol
0xNuggan: Good catch! I'll change it

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
0xNuggan: Since the contract can access the variable directly I thought it wasn't needed to make it public. The other changes from external to public were mostly because we are now calling the functions (f.e. buyFor) from the contract itself.

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
Zitzak: No all good. It's just personal preference whereby I imported them separately when changing all the names of the contracts. But it's not important enough to add more work :) Was just wondering if it had something to do with the storage

Zitzak (approved):
LGTM

PR #468, opened 2024-05-14
Feat: Limit the Maximum Fee

FHieser:


test/external/FeeManager_v1.t.sol
```diff
@@ -237,6 +241,36 @@ contract FeeManagerTest is Test {
...
+        amount = bound(amount, 0, feeMan.BPS());
+
+        vm.expectEmit(true, true, true, true);
+        emit MaxFeeSet(amount);
+
+        feeMan.setMaxFee(amount);
+        assertEq(feeMan.maxFee(), amount);
+    }
+
+    function testSetMaxFeeModifierInPosition() public {
```
0xNuggan, line 266: I'd split this one into two tests: failsIfUnauthorized and failsIfFeeAbove max. That also makes it easy to fuzz both

FHieser: I disagree here
The ModifierInPosition tests are just to check that the modifier are here in the first place.
Their functionality is tested in their own test with fuzzing at all. (Which I promptly forgot for the maxfee modifier lol)
Its like testing an internal function separately first and then assuming it works basically

0xNuggan (approved):

marvinkruse (approved):
lgtm! 🚀

PR #469, opened 2024-05-14
Chore: Remove Metadata Manager

FHieser:


0xNuggan (approved):
LGTM

marvinkruse:
@0xNuggan the CI is failing, because of this test: `testSetOptimisticOracleFails_WhenNewOracleIsNotUmaOptimisticOracle`. 

What I can see is that it expects a revert, but the call doesn't revert. This is because a contract address is used as input. `setOptimisticOracle` itself takes this address and calls a function on it to get the default identifier - if this contract doesn't implement it, this **does not** fail, but instead the `defaultIdentifier` is just zero. I believe that is why this test is failing, it's using a contract address as input (one that is not excluded) and thus the call doesn't fail anymore. 

I guess the implementation would need to check whether the `defaultIdentifier` is actually non-zero (which it doesn't currently)? Unless a zero default identifier should be allowed, in which case the test just needs to check for zero instead of a failing call I'd assume.

0xNuggan:
Funnily enough, it wasn't that.

The reason was that even though we were excluding the OptimisticOracleIntegrator from the fuzzing, we weren't excluding the standalone implementation address the proxy points to. That's also why it was returning zero. 

The CI caught this, while my local fuzz run didn't think of trying out the (only briefly stored during setup) impl address.

I've updated the test file.

PR #470, opened 2024-05-14
Chore: Move VotingRoleManager to authorizer extensions

FHieser:


marvinkruse:
Did you merge the MetadataManager removal changes into this PR already or why is it in here as well?

FHieser:
> Did you merge the MetadataManager removal changes into this PR already or why is it in here as well?

Yes exactly

0xNuggan:
I see the "[thing]Manager" nomenclature is still alive and healthy :rofl: I think we wanted to move away from it? Maybe something like "AUT_EXT_Voting_v1" would be better?

FHieser:
> I see the "[thing]Manager" nomenclature is still alive and healthy 🤣 I think we wanted to move away from it? Maybe something like "AUT_EXT_Voting_v1" would be better?

Sure
I have no hard opinion on this. @marvinkruse ?

marvinkruse:
Sorry to be nitpicking, but one thing I noticed is that you called the folder `extension`, while I think it should be `extensions` (plural), just like we use `interfaces` or `abstracts`. Otherwise I'm happy 👍

> I have no hard opinion on this. @marvinkruse?

I am not against the `Manager` naming, I still like it, but he has a point with the `AUT_EXT` part. Maybe `AUT_EXT_VotingRoles_v1`? As I think just having `Voting` is a bit ambiguous in terms of "_voting of what?_". 
(Tagging @0xNuggan as well for a quick check for opinion),

marvinkruse (changes requested):
- Renaming `extension` to `extensions` (if you agree)
- Renaming of contract name according to discussion above

marvinkruse (approved):

PR #473, opened 2024-05-17
Feat: Add Cliff to Streaming Payments

marvinkruse:
⚠️ **READ THIS** ⚠️ 
Because of how GitHub handles diffs in PRs (see details below), the diff here is wrong. Use **[this link](https://github.com/InverterNetwork/contracts/compare/dev..feat/add-cliff-to-stream?diff=split&w=)** to get a clear diff between the `dev` branch and the `feat/add-cliff-to-stream` branch of this commit to do your review, unless you want to suffer. Don't worry, the PR doesn't include all of these changes that you see here.

<details>
  <summary><i>More details on this</i></summary>

> This is a result of the type of diff we use on GitHub. We use git’s three dot diff, which is the difference between the latest commit on the HEAD branch and the last common ancestor commit with the base branch.

> There are two comparison methods for the git diff command; two-dot (git diff A..B) and three-dot (git diff A...B). By default, pull requests on GitHub show a three-dot diff.

You can find more details on this here:
* https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests#three-dot-and-two-dot-git-diff-comparisons
* https://stackoverflow.com/questions/55479729/github-pull-request-shows-wrong-diff
</details>

--------

**Current Status**: ⚠️ Implementation of the cliff is done, but its not specifically tested yet. @0xNuggan taking over for that.

--------

The StreamingPaymentProcessor needs to be able to offer a proper cliff functionality. Payment shall be created with a `cliff` variable, which contains a duration. Any streaming may only start once this duration has passed (after the payment was created/started). However, the **actual vesting shall already happen in the meantime**. 

<details>
  <summary><i>See example</i></summary>

* **Payment A**: Start at 1pm, Ends 5pm, no Cliff, 100 Tokens
  * At 2pm 25 Tokens can be claimed (25% vested)
  * At 3pm 50 Tokens can be claimed  (50% vested)
  * At 4pm 75 Tokens can be claimed  (75% vested)
  * At 5pm 100 Tokens can be claimed (100% vested)
* **Payment B**: Start at 1pm, Ends 5pm, 3 Hour Cliff, 100 Tokens
  * At 2pm 0 Tokens can be claimed (cliff active, 25% vested)
  * At 3pm 0 Tokens can be claimed (cliff active, 50% vested)
  * At 4pm 75 Tokens can be claimed (cliff ended, 75% vested)
  * At 5pm 100 Tokens can be claimed (100% vested) 

Example screenshot

</details>


0xNuggan:
@marvinkruse ready for review! (I cannot tag you as reviewer since you created the PR)

**NOTE**: The changes to the PaymentRouter batched orders were necessary to avoid stack to deep errors. I think they make sense conceptually

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
```diff
@@ -74,20 +81,62 @@ contract LM_PC_PaymentRouter_v1 is
             IERC20PaymentClientBase_v1(address(this))
         );
     }
+    /*
```
FHieser: Outcommented

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
```diff
@@ -61,11 +61,18 @@ contract LM_PC_PaymentRouter_v1 is
...
+        uint end
     ) public onlyModuleRole(PAYMENT_PUSHER_ROLE) {
         _addPaymentOrder(
             PaymentOrder(
-                recipient, paymentToken, amount, block.timestamp, dueTo
+                recipient,
+                paymentToken,
+                amount,
+                start == 0 ? block.timestamp : start,
+                cliff,
```
FHieser, line 74: Are all Cliffs valid?
Even if they are bigger than than end - start?
If so you need to update the addPamyentOrder Modifier

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
```diff
@@ -74,20 +81,62 @@ contract LM_PC_PaymentRouter_v1 is
...
+                ends[i]
+            );
+        }
+
+        // call PaymentProcessor
+        __Module_orchestrator.paymentProcessor().processPayments(
+            IERC20PaymentClientBase_v1(address(this))
+        );
+    }
+    */
```
marvinkruse: Why is this commented out?

test/modules/paymentProcessor/PP_Streaming_v1.t.sol
```diff
@@ -358,14 +366,128 @@ contract PP_StreamingV1Test is //@note do we want to do anything about these tes
         assertEq(totalAmount, paymentClient.amountPaidCounter());
     }
 
-    // @dev Assume recipient can withdraw full amount immediately if dueTo is less than or equal to block.timestamp.
-    function testProcessPaymentsWorksForDueTimeThatIsPlacedBeforeStartTime(
+    // test cannot claim before cliff ends
+
+    function test_claimStreamedAmounts_cannotClaimBeforeCliff(
```
marvinkruse, line 371: This test is good in terms of testing whether:
* User cannot claim before cliff runs out
* User can claim 100% once end date is reached

But what we are not testing yet at all is
* User can claim the right amount directly after cliff is over
  * i.e. if the whole duration is 10 and the cliff is 1 (`start = 0`, `cliff = 1`, `end = 10`), the user should be able to claim 10% (1/10th) after the cliff is over (as the vesting itself already is in progress and doesn't start after the cliff)

src/modules/logicModule/interfaces/ILM_PC_PaymentRouter_v1.sol
```diff
@@ -7,14 +7,17 @@ interface ILM_PC_PaymentRouter_v1 {
...
         address recipient,
         address paymentToken,
         uint amount,
-        uint dueTo
+        uint start,
+        uint cliff,
+        uint end
     ) external;
 
+    /*
```
marvinkruse: Can this be cleaned up if the commented-out version is not required anymore? Looks like a leftover to me at least?

marvinkruse (commented):
Generally looks good 👍 Just found a few commented-out things as well as the tests not covering everything I believe.

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -784,57 +792,70 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
-
-        if (timestamp < startPaymentReceiver) {
+        uint total = streams[client][paymentReceiver][streamId]._total;
+        uint start = startForSpecificStream(client, paymentReceiver, streamId);
+        uint end = endForSpecificStream(client, paymentReceiver, streamId);
+
+        if (
+            timestamp
+                < (
+                    start
```
FHieser, line 814: Lol this is the only real change right?

FHieser (commented):
All in all Looks good to me
Take a look at the comments

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
marvinkruse: The cliff is being checked in the `StreamingPaymentProcessor`, in the sense that `start + cliff <= end` is required. If `cliff > end - start` it would violate that and throw an error. Or do you propose to already check this here as well?

src/modules/paymentProcessor/PP_Streaming_v1.sol
marvinkruse: Yes, this is the only real change here regarding the cliff, ensuring that nothing is sent until cliff is over.

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
0xNuggan: that was a leftover, sorry for that

test/modules/paymentProcessor/PP_Streaming_v1.t.sol
0xNuggan: That would have been tested in the existing tests, since the underlying vest calculation logic hasn't been changed. But I'll add an explicit test for that.

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
FHieser: The problem is that in the case that the paymentOrders arent immediatley processed, then we basically create faulty paymentOrders, that cant be processed later on
Id rather check 2 times

marvinkruse (commented):
Thank you @0xNuggan - looks good to me! Can't approve it as it's my PR, but you can approve it in my stead if you want. I'm happy! 🙏

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
0xNuggan: The paymentRouter _does_ immediately process them though, so we would be double-checking. And the other modules don't implement cliffs. I think we can keep it like that

0xNuggan (approved):

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
marvinkruse: I see your point @FHieser, but also agree with @0xNuggan - we can keep it like this for now I think.

FHieser: We dont enforce this though.
If someone else comes up and wants to develop something this has to be enforced
I will create a linear issue for it so we keep it in mind

PR #474, opened 2024-05-20
[SC-290] Expand Payment Order struct

0xNuggan:
## Summary
To prepare the PaymentRouter Module we need to adapt the basic PaymentOrder struct so the token used can be sepcified. This way, we can send not only collateral tokens,  but also issuance tokens. This PR implements that change.

## What has been done
The main change has been to expand the PaymentOrder struct in IERC20PaymentClientBase_v1 with a field that specifies the token address. This has led to significant refactors in the PaymentClient/PaymentProcessor contracts. 

The main logic changes have been made inside the ERC20PaymentClientBase_v1 contract: the outstandingTokenAmount variable has been replaced with a mapping, and the collectPaymentOrders() function reworked as a consequence. It now returns an array of the tokens in the paymentOrders and the total outstanding amounts for each. Corressponding processPayments() functions in PP_Simple_v1 and PP_Streaming_v1 have been adapted too. 

This PR also re-comments a test in UMA Oracle test base that was causing errors. I've opened an issue on linear (SC-289) to track this.


src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -104,22 +106,10 @@ abstract contract ERC20PaymentClientBase_v1 is
...
-            _ensureValidPaymentOrder(currentOrder);
-
-            // Add order's amount to total amount of new orders.
-            totalTokenAmount += currentOrder.amount;
-
-            // Add new order to list of oustanding orders.
-            _orders.push(currentOrder);
-
-            emit PaymentOrderAdded(currentOrder.recipient, currentOrder.amount);
+            _addPaymentOrder(currentOrder);
```
FHieser: remove currentorder and go with orders[i] directly

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -203,50 +226,59 @@ abstract contract ERC20PaymentClientBase_v1 is
...
-    {
-        __Module_orchestrator.fundingManager().token().safeIncreaseAllowance(
-            address(spender), amount
-        );
+    function _ensureTokenAllowance(
+        IPaymentProcessor_v1 spender,
+        address token,
+        uint amount
+    ) internal virtual {
+        IERC20(token).safeIncreaseAllowance(address(spender), amount);
```
FHieser: Didnt we talk about this being allowance should be upped to outstanding token amount?

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -203,50 +226,59 @@ abstract contract ERC20PaymentClientBase_v1 is
...
+    function _ensureValidPaymentOrder(PaymentOrder memory order) private view {
+        _ensureValidRecipient(order.recipient);
+        _ensureValidToken(order.paymentToken);
+        _ensureValidAmount(order.amount);
+    }
+
     //--------------------------------------------------------------------------
     // {ERC20PaymentClientBase_v1} Function Implementations
 
     /// @dev Ensures `amount` of payment tokens exist in address(this).
```
FHieser: I would love a @dev comment here explaining what exactly is happening here

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -153,39 +148,67 @@ abstract contract ERC20PaymentClientBase_v1 is
         }
 
         // Create a copy of all orders to return.
-        uint totalAmount;
         uint ordersLength = _orders.length;
+        uint tokenCount;
+
+        address[] memory tokens_buffer = new address[](ordersLength);
```
FHieser, line 151: I dont love this
How about we create a struct combination that allows to order the payment orders directly by token

For example:

struct paymentOrderCollection{
address token;
paymentOrders[] orders;
} 

That way you can actually keep the old Payment Order strcuture and you dont need special sorting of the different tokens. You would then basically only check when creating if a collection already exists for that token.
You could probably even optimize it with a mapping and array of current collections to look stuff up cheaper there

FHieser: We agreed together that we were gonna put this on the backlog due to time concerns
We will create a linear issue to rework this

test/modules/logicModule/oracle/OptimisticOracleIntegrator.t.sol
```diff
@@ -209,13 +209,15 @@ contract OptimisticOracleIntegratorTest is ModuleTest {
         ooIntegrator.setOptimisticOracle(address(0));
     }
 
+    /*
```
FHieser, line 212: pedro-pascal

FHieser (changes requested):
All in all looks good
Minor changes needed

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
0xNuggan: Changed it and updated test

test/modules/logicModule/oracle/OptimisticOracleIntegrator.t.sol
0xNuggan: pedroPascal

FHieser (approved):
LGTM

PR #476, opened 2024-05-22
[SC-270]  Add Payment Router Module

0xNuggan:
[Proper PR message TODO]

This PR adds the PaymentRouter Module, which allows an external authorized party to push payments into the PaymentProcessor directly.

It builds on top of #474 **so that one should be merged first**


src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
```diff
@@ -0,0 +1,115 @@
...
+    bytes32 public constant PAYMENT_PUSHER_ROLE = "PAYMENT_PUSHER";
+
+    /// @inheritdoc ILM_PC_PaymentRouter_v1
+    function pushPayment(
+        address recipient,
+        address paymentToken,
+        uint amount,
+        uint dueTo
+    ) public onlyModuleRole(PAYMENT_PUSHER_ROLE) {
+        _addPaymentOrder(
```
FHieser, line 66: Do we want to add events here?

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
```diff
@@ -0,0 +1,115 @@
...
+// Internal Libraries
+
+// External Libraries
+
+// NOTE: Development halted until the payment processor can handle multiple tokens,  and not only the native orchestrator one
+
+/**
+ * @title   Payment Router
+ *
+ * @notice  Tis module is a stopgap solution to enable pushing payments to the Payment Processor.
```
FHieser: Typo
*This

FHieser: Also I think this is a valid Module in General
I wouldnt call it stopgap solution

FHieser (changes requested):
I would label this as a fully fletched out module. 
Also do we want to add events to the PaymentOrder Creation?

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
0xNuggan: We do emit an PaymentOrderAdded event. Do you mean another event?

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
0xNuggan: Stopgap because we are gonna redesign the whole thing anyway. But I'll remove it for now

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
FHieser: Yeah your probably right
I thought because we have extra events in the other logicmodules it might make sense to do it here too
In those cases we usually represent state changes that are not happening here I guess

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
```diff
@@ -22,12 +22,10 @@ import {
 
 // External Libraries
 
-// NOTE: Development halted until the payment processor can handle multiple tokens,  and not only the native orchestrator one
-
 /**
  * @title   Payment Router
  *
- * @notice  Tis module is a stopgap solution to enable pushing payments to the Payment Processor.
+ * @notice  This module is enable pushing payments directly to the Payment Processor.
```
FHieser: This module ~~is~~ enable**s** pushing payments directly to the Payment Processor.

FHieser (changes requested):

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
0xNuggan: I feel in this case an extra event isn't necessary

FHieser (approved):
Approved: Pablo learned how to write :+1:

PR #477, opened 2024-05-23
[SC-138] Feat: Immutability Option for Module Deployments

FHieser:
Changes I made here:
- Create TransparentInverterProxy
- replace orchestratorConfig with workflowConfig
- Enable the creation of the TransparentInverterProxy instead of the BeaconProxy in the ModuleFactory via the workflowConfig
- Make Orchestrator upgradeable via BeaconProxy in OrchestratorFactory
- Enable the creation of the TransparentInverterProxy instead of the BeaconProxy in the OrchestratorFactory via the workflowConfig


marvinkruse:
Haven't reviewed it in detail yet, but do we not also need a way to retrieve the correct version from the new `TransparentInverterProxy` - so do we need to overwrite the `version()` function there as well? Otherwise we'll have the same issue as before with the version not being correct after an upgrade I believe?

FHieser:
@marvinkruse changed it accordingly

src/factories/OrchestratorFactory_v1.sol
```diff
@@ -115,57 +129,63 @@ contract OrchestratorFactory_v1 is
...
-        for (uint i; i < modulesLen; ++i) {
-            modules[i] = IModuleFactory_v1(moduleFactory).createModule(
-                moduleConfigs[i].metadata,
-                IOrchestrator_v1(clone),
-                moduleConfigs[i].configData
-            );
-        }
-
-        if (orchestratorConfig.owner == address(0)) {
-            revert OrchestratorFactory__OrchestratorOwnerIsInvalid();
```
Zitzak, line 162: We might want to add a similar check for the indepened admin, like we did with the owner

Zitzak: To be more specific (as I get a better understanding), if independentUpdates == true then admin can't be zero address

script/factories/DeployOrchestratorFactory_v1.s.sol
```diff
@@ -77,14 +88,25 @@ contract DeployOrchestratorFactory_v1 is Script {
 
         (orchestratorFactoryBeacon, orchestratorFactoryProxy) =
         deployAndSetUpInverterBeacon_v1.deployBeaconAndSetupProxy(
-            governor, orchestratorFactoryImplementation, 1, 0
+            governor,
+            orchestratorFactoryImplementation,
+            1,
+            0 //@note do we have a way to smartly track these Versions?
```
Zitzak, line 94: We don't yet. Still major todo after public audit start

Makefile
```diff
@@ -93,7 +93,9 @@ testScripts: ## Run e2e test suite
 
 	## factories
 	@forge script script/factories/DeployModuleFactory_v1.s.sol
-	@forge script script/factories/DeployOrchestratorFactory_v1.s.sol
+	
+	## @note Because the Beacon doesnt allow to be passed addresses without implementations we cant test the script properly anymore without breaking the script testing structure we used up until now.
```
Zitzak, line 97: Could you add an issue for this in Linear, with some context please to ensure it doesn't get lost. Project could be [DEPLOY] Protocol V1

test/utils/mocks/proxies/InverterBeaconV1Mock.sol
```diff
@@ -43,6 +43,10 @@ contract InverterBeaconV1Mock is IInverterBeacon_v1, ERC165 {
         return (majorVersion, minorVersion);
     }
 
+    function getImplementationAddress() external returns (address) {
```
Zitzak: Compiler warning about state mutability

test/utils/mocks/proxies/InverterTransparentUpgradeableProxyV1AccessMock.sol
```diff
@@ -0,0 +1,44 @@
+pragma solidity ^0.8.0;
+
+import {InverterTransparentUpgradeableProxy_v1} from
+    "src/proxies/InverterTransparentUpgradeableProxy_v1.sol";
+
+import {IInverterBeacon_v1} from "src/proxies/interfaces/IInverterBeacon_v1.sol";
+
+contract InverterTransparentUpgradeableProxyV1AccessMock is
```
Zitzak, line 8: Compiler warning

test/proxies/InverterTransparentUpgradeableProxy_v1.t.sol
```diff
@@ -0,0 +1,143 @@
...
+        );
+
+        vm.expectRevert(
+            InverterTransparentUpgradeableProxy_v1
+                .InverterTransparentUpgradeableProxy__ProxyDeniedAdminAccess
+                .selector
+        );
+
+        vm.prank(admin);
+        address(proxy).call(abi.encode(sig));
```
Zitzak, line 112: Compiler warning

test/proxies/InverterTransparentUpgradeableProxy_v1.t.sol
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    function test_fallbackWorksIfAdminCallsUpgradeToNewestVersion() public {
+        proxyMock.flipUpgradeToNewestVersionActive();
+
+        vm.prank(admin);
+        address(proxyMock).call(
```
Zitzak, line 119: Compiler warning

test/e2e/paymentProcessor/StreamingPaymentProcessorE2E.sol
```diff
@@ -91,14 +91,14 @@ contract StreamingPaymentProcessorE2E is E2ETest {
...
-            token: token
+        IOrchestratorFactory_v1.WorkflowConfig memory workflowConfig =
+        IOrchestratorFactory_v1.WorkflowConfig({
+            independentUpdates: false,
+            independentUpdateAdmin: address(0)
         });
 
-        orchestrator =
-            _create_E2E_Orchestrator(orchestratorConfig, moduleConfigurations);
+        IOrchestrator_v1 orchestrator =
```
Zitzak: Compiler warning

Zitzak (changes requested):
Only have one comment about a check which might be valuable. All others are requests to fix the compiler warnings for newly introduced changes, as this is something which also came up during the audit

src/factories/OrchestratorFactory_v1.sol
FHieser: I thought about it,
but the Proxy prevents the usage of address(0) as the owner anyway so it should be fine

test/proxies/InverterTransparentUpgradeableProxy_v1.t.sol
FHieser: The compiler warnings on the calls are super dumb
I check the functionality of the call via the reverts / the state changes in the first place
I could check if the calls are successes i guess, but I dont know what to do with the data fields.
I tried to do a check for them, but it seemed super random what came back (at least in the function signature check)
Do you have any idea what I should check here for? @Zitzak

src/factories/OrchestratorFactory_v1.sol
FHieser: It does it in the ERC1967Utils.changeAdmin(initialOwner);

Zitzak: Ahh okay, didn't know that. Thank you

Makefile
FHieser: Added it in linear as SC-555

test/proxies/InverterTransparentUpgradeableProxy_v1.t.sol
Zitzak: Hmm not really. Otherwise silence or ignore I guess @FHieser

FHieser: I would ignore it for starters I think
Lets push it to the cleanup phase after launch

test/proxies/InverterTransparentUpgradeableProxy_v1.t.sol
FHieser: see other comment https://github.com/InverterNetwork/contracts/pull/477#discussion_r1620755918

test/proxies/InverterTransparentUpgradeableProxy_v1.t.sol
FHieser: @Zitzak could you approve it then?

src/proxies/InverterTransparentUpgradeableProxy_v1.sol
```diff
@@ -0,0 +1,120 @@
...
+    error InverterTransparentUpgradeableProxy__InvalidBeacon();
+
+    /// @dev If the proxy caller is the current admin then it can only call the admin functions.
+    error InverterTransparentUpgradeableProxy__ProxyDeniedAdminAccess();
+
+    //--------------------------------------------------------------------------------
+    // State
+
+    /// @dev The address of the admin that can update the implementation address of this proxy
+    address internal immutable _admin;
```
marvinkruse, line 50: In the original implementation of OZ this is `immutable` as it's using a ProxyAdmin. Here, it's probably an EOA, so this should be changeable here, right?

FHieser: I thought about this too.
The problem is that every single contract in the workflow would need to be updated for the admin to properly work.
And because only the admin can change the admin, then if the address should be corrupted the whole system would be useless anyway. The recommondation they gave in the Original proxy was to make it a specially created account in the first place so thats what I went with
Should I maybe add another disclaimer either in the workflowConfig or in the proxy itself?

Zitzak (approved):

marvinkruse (approved):
lgtm - great work!

src/proxies/InverterTransparentUpgradeableProxy_v1.sol
FHieser: We agreed to make a seperate admin contract that allows to update the proxies via the immutable admin role
We will create a linear issue for this

PR #483, opened 2024-05-28
Feat: Only Allow Adding of Modules Registered in the Factory

marvinkruse:
This PR limits the ModuleManager to only allow adding modules that have previously been created via our own factory. Basically, whenever a module is created in the `ModuleFactory` via `createModule` (which deploys the proxy), we store this `proxy` in a mapping that returns the address of the Orchestrator, for which it has been created. This way, the Module Manager (which is the Orchestrator by inheritance), can just verify whether the returned address is `address(this)`(aka itself) to ensure that this module is (a) registered and (b) has been created for itself, not for a different Orchestrator.

I adapted the tests and everything should (tm) be okay. 🤞

### DIFF
--- a/src/factories/ModuleFactory_v1.sol
+++ b/src/factories/ModuleFactory_v1.sol
@@ -94,6 +94,9 @@ contract ModuleFactory_v1 is
     /// @dev MetadataLib.identifier(metadata) => {IInverterBeacon_v1}
     mapping(bytes32 => IInverterBeacon_v1) private _beacons;
 
+    /// @dev Mapping of proxy address to orchestrator address.
+    mapping(address => address) private _orchestratorOfProxy;
+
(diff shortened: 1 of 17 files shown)


FHieser (approved):
LGTM

0xNuggan (approved):
But I wanted to ping-pong existing Modules between orchestrators!!1!
Just kidding, looks good to me :rocket: 

Btw good call on renaming implementation to proxy.

marvinkruse:
> But I wanted to ping-pong existing Modules between orchestrators!!1!
> 
> Just kidding, looks good to me :rocket: 
> 
> 
> 
> Btw good call on renaming implementation to proxy.

Haha thank you ser 🙏 Yeah that always bugged me but I always forgot about it 😂 Thanks for the review!

PR #484, opened 2024-05-28
[SC-382] Audit(L-8): Removal of token gated role does not remove role associated data

0xNuggan:
## What has been done
- Add state deletion on token role removal.

While fixing the issue I realized that there was another edge case possible: adding a token role through the grantModuleRole of the parent contract once it was set as tokenGated would set the threshold at the default value (-> 0), effectively circumventing authorization.
- Added extra check on role granting to address that

### DIFF
--- a/src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
+++ b/src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
@@ -110,7 +110,7 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
     /// @param role The role to grant
     /// @param who The address to grant the role to
     /// @return bool Returns if the role has been granted succesful
-    /// @dev Overrides {_grantRole} from AccesControl to enforce interface implementation when role is token-gated
+    /// @dev Overrides {_grantRole} from AccesControl to enforce interface implementation and threshold existence when role is token-gated
     function _grantRole(bytes32 role, address who)
         internal
         virtual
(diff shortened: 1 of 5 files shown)


src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
```diff
@@ -92,39 +92,62 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
...
-                revert Module__AUT_TokenGated_Roles__InvalidToken(account);
+                revert Module__AUT_TokenGated_Roles__InvalidToken(who);
             }
         }
 
-        return super._grantRole(role, account);
+        return super._grantRole(role, who);
+    }
+
+    /// @notice Overrides {_revokeRole} to prevent having an empty OWNER role
```
marvinkruse: Is this comment valid? Doesn't this just prevent the leftover thresholds after removing the role? Or am I missing something?

marvinkruse (commented):
looks good, just one question about the NatSpec 🙌

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
0xNuggan: You are right. The parent function does, but I updated the comment

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
```diff
@@ -110,7 +110,7 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
     /// @param role The role to grant
     /// @param who The address to grant the role to
     /// @return bool Returns if the role has been granted succesful
-    /// @dev Overrides {_grantRole} from AccesControl to enforce interface implementation when role is token-gated
+    /// @dev Overrides {_grantRole} from AccesControl to enforce interface implementation and threshold existence when role is token-gated
```
marvinkruse: Sorry, one more thing: `AccesControl` should be `AccessControl` 😅

marvinkruse:
@0xNuggan Sorry for the pushes here, the address collision fix was still visible here in the diff, so I reset it to your last commit before and rebased from `macro-fixes`. It's clean now

marvinkruse (approved):
lgtm!

0xNuggan:
Thank you! Sorry for the mess

PR #487, opened 2024-05-30
[SC-400] Audit(Q-5) unused events

FHieser:
Removed the event


marvinkruse:
Is this the right fix, @0xNuggan - or should the event actually be used?

0xNuggan:
LGTM - we are emitting the duration in the RewardsSet() event anyway, and do not store it locally

0xNuggan (approved):

PR #488, opened 2024-05-30
[SC-297] Audit(H-4) supportinterface not correctly implemented in

FHieser:
Added the IPaymentProcessor_v1 interface to supportsInterface

### DIFF
--- a/src/modules/paymentProcessor/PP_Streaming_v1.sol
+++ b/src/modules/paymentProcessor/PP_Streaming_v1.sol
@@ -44,6 +44,7 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
         returns (bool)
     {
         return interfaceId == type(IPP_Streaming_v1).interfaceId
+            || interfaceId == type(IPaymentProcessor_v1).interfaceId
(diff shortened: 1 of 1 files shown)


marvinkruse:
You chose a different way of fixing that compared to what the auditors proposed. Quoting the proposed solution here:

> Update PP_Streaming_v1:supportsInterface() implementation **and** instead of IPP_Streaming_v1 inheriting from the IPaymentProcessor_v1 make PP_Streaming_v1 inherit IPaymentProcessor_v1.

So they propose on top of fixing the `supportsInterface()` to have `PP_Streaming_v1` inherit from `IPaymentProcessor_v1` instead of `IPP_Streaming_v1` inheriting from it. Any thoughts?

FHieser:
IPP_Streaming_v1 is inheriting from IPaymentProcessor_v1 anyway so I dont see a reason to "downgrade" to just the IPaymentProcessor_v1 interface

marvinkruse:
Alright, that's fine! The fix itself looks good to me, let's do it this way then :)

marvinkruse (approved):

PR #494, opened 2024-05-30
[SC-379] Bug(L-7): Interface Getter Functions not set as View

marvinkruse:


FHieser (approved):
Isnt it super weird that this isnt just enforced?

marvinkruse:
> Isnt it super weird that this isnt just enforced?

Yeah it's weird, the compiler lets you know of most of them though, so it is aware, but there might be some edge-cases where having a function non-view might be beneficial probably. But I would also prefer if it was enforced :D

PR #507, opened 2024-05-31
[SC-482] Audit(M-7) staking rounding errors

FHieser:


0xNuggan:
LGTM - Two decimals rounding error shoudl be good enough even for tokens like USDC/T

0xNuggan (approved):

PR #515, opened 2024-05-31
[SC-556] Feat: Implement ProxyAdmin flow for self-governed module upgrades

marvinkruse:
**Status**: ⌛ Done, pending reviews.

<hr> 

**User Story**
As a user, I want to be able to transfer the rights to upgrading my self-governed modules, and it should be easy, not needing to transfer each module one-by-one.

**Additional Context**
* Currently each modules proxy is owned by a immutable address.
* Also the admin of the proxies can't interact with the underlying module itself, due to the nature of the TransparentProxy
* Proposed Solution:
  * We wanna have it be owned by a ProxyAdmin, that is deployed upon workflow creation
  * All modules + orchestrator owned by that one ProxyAdmin

### DIFF
--- a/src/factories/OrchestratorFactory_v1.sol
+++ b/src/factories/OrchestratorFactory_v1.sol
@@ -138,7 +138,7 @@ contract OrchestratorFactory_v1 is
         address proxy;
         //If the workflow should fetch their updates themselves
         if (workflowConfig.independentUpdates) {
-            //Use a InverterTransparentUpgradeableProxy as a proxy
+            //Use an InverterTransparentUpgradeableProxy as a proxy
             proxy = address(
                 new InverterTransparentUpgradeableProxy_v1(
                     beacon, workflowConfig.independentUpdateAdmin, bytes("")
--- a/src/factories/OrchestratorFactory_v1.sol
+++ b/src/factories/OrchestratorFactory_v1.sol
@@ -7,6 +7,7 @@ import {InverterBeaconProxy_v1} from "src/proxies/InverterBeaconProxy_v1.sol";
 import {InverterTransparentUpgradeableProxy_v1} from
     "src/proxies/InverterTransparentUpgradeableProxy_v1.sol";
 import {IInverterBeacon_v1} from "src/proxies/interfaces/IInverterBeacon_v1.sol";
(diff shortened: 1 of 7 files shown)


Zitzak:
I'm seeing that the issue for this got created last Friday. Did this slip through the cracks or is this part of a solution for another issue? What still needs to happen for this?

Also assuming this needs to be done as well before Wednesday? 
@marvinkruse

FHieser:
@Zitzak We noticed for the immutability update that it would be practical if we created a governed Admin for the Modules that needed individual updates instead of using a predefined address that wasnt changeable
We agreed that we would tackle this in another issue as time was pressing. Didnt think we would include this now @marvinkruse

marvinkruse:
I think it would be great to have this one in there, if possible. So I started on it between my other tasks. I'll try to get it done today, otherwise it'll be a later thing for later.

marvinkruse:
@FHieser ready for your review, sir 🫡 
if you have time (which I doubt, but just in case) @Zitzak and @0xNuggan would be great too ofc! but not a must.

src/factories/OrchestratorFactory_v1.sol
```diff
@@ -138,7 +139,14 @@ contract OrchestratorFactory_v1 is
         address proxy;
         //If the workflow should fetch their updates themselves
         if (workflowConfig.independentUpdates) {
-            //Use a InverterTransparentUpgradeableProxy as a proxy
+            //Deploy a proxy admin contract that owns the invidual proxies
+            //Overwriting the independentUpdateAdmin as the ProxyAdmin will
+            //be the actual admin of the proxy
+            workflowConfig.independentUpdateAdmin = address(
```
marvinkruse, line 145: Overwriting this in the config, as it is used down the line by the `createModule` function, thus using the `ProxyAdmin` as the admin, instead of the one that the user has set. The one the user set is used here, so the `ProxyAdmin` is owned by them.

FHieser (changes requested):
LGTM
Good work with the overwrite of the indepententAdmin :+1: 

Was supposed to be an approve :3

FHieser (approved):

marvinkruse:
> LGTM
> 
> Good work with the overwrite of the indepententAdmin :+1: 

That was quick, thank you!

PR #516, opened 2024-05-31
[SC-310] Bug(M-1): Inflated supply when redeeming issuance token on FM_BC

0xNuggan:
## What has been done
- Moved the processProtocolFee some lines down so we re-mint some of the burned tokens to the treasury. This avoids having to worry about allowances/balances


Zitzak (approved):
LGTM

PR #517, opened 2024-05-31
Fix: Make issuance token interface inherit IERC20

0xNuggan:
Currently the interface does not inherit the base IERC20 contract, this can make casting cumbersome when using only the interface. Since the implementation inherits ERC20, I think we can make the interface inherit it too.


marvinkruse (approved):
lgtm - good idea!

PR #520, opened 2024-06-03
[SC-557] Bug: Update state modifier BC functions used to calculate price

Zitzak:
## Why
The function `calculateSaleReturn` and `calculatePurchaseReturn` had state modifiers signalling that a state would be changed. This would force the FE to simulate a call to the function instead of just reading it. This was a bug, as they don't change a state
## What has been done?
- Update state mutability for the functions `calculateSaleReturn` and `calculatePurchaseReturn` as well as all the functions they depended on which were incorrectly not set to view

### DIFF
--- a/src/external/fees/interfaces/IFeeManager_v1.sol
+++ b/src/external/fees/interfaces/IFeeManager_v1.sol
@@ -143,7 +143,7 @@ interface IFeeManager_v1 {
         address workflow,
         address module,
         bytes4 functionSelector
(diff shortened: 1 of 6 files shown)


0xNuggan (approved):

0xNuggan:
Looks good!

Zitzak:
Should this be merged to `dev` as it is directed now, or also into the `macro-fixes` branch? Wdyt @marvinkruse ?

PR #522, opened 2024-06-03
[SC-497] Bug(M-5): Queued stakers can receive rewards from pending assertions

0xNuggan:
## What has been done
- Added a boolean check if there is already a live assertion posted and revert in that case
- Added corresponding error

### DIFF
--- a/src/modules/logicModule/LM_PC_KPIRewarder_v1.sol
+++ b/src/modules/logicModule/LM_PC_KPIRewarder_v1.sol
@@ -61,6 +61,7 @@ contract LM_PC_KPIRewarder_v1 is
     mapping(bytes32 => RewardRoundConfiguration) public assertionConfig;
 
     // Deposit Queue
+    bool public assertionPending;
     address[] public stakingQueue;
     mapping(address => uint) public stakingQueueAmounts;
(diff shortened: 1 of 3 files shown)


src/modules/logicModule/LM_PC_KPIRewarder_v1.sol
```diff
@@ -364,6 +371,9 @@ contract LM_PC_KPIRewarder_v1 is
             assertionData[assertionId].asserter,
             assertionId
         );
+
+        // Independently of the fact that the assertion resolved true or not, new assertions can now be posted.
+        assertionPending = false;
```
FHieser, line 376: @0xNuggan Just out of interest:
is there a usecase where a assertion cant be resolved?

FHieser: Otherwise LGTM

0xNuggan: No, it ends up resolved truthfully or not. It can be disputed though, which delays the process

Zitzak (approved):
LGTM

PR #523, opened 2024-06-03
[Sc-479] Bug(M-6): Asserter can post invalid assertions with minimal penalty

0xNuggan:
## What has been done
- Added the possibility to manually set the required Bond when setting the currency (enforcing it to be higher or equal to the minimum). This way users can set a higher one if they believe the funds at stake merit it.
- Adapted corresponding init functions and tests

### DIFF
--- a/src/modules/logicModule/abstracts/oracleIntegrations/UMA_OptimisticOracleV3/OptimisticOracleIntegrator.sol
+++ b/src/modules/logicModule/abstracts/oracleIntegrations/UMA_OptimisticOracleV3/OptimisticOracleIntegrator.sol
@@ -73,7 +73,7 @@ abstract contract OptimisticOracleIntegrator is
     ) internal onlyInitializing {
         _setOptimisticOracle(ooAddr);
         _setDefaultAssertionLiveness(liveness);
-        _setDefaultCurrency(currencyAddr, bondAmount);
+        _setDefaultCurrencyAndBond(currencyAddr, bondAmount);
     }
 
     //--------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


src/modules/logicModule/abstracts/oracleIntegrations/UMA_OptimisticOracleV3/OptimisticOracleIntegrator.sol
```diff
@@ -96,11 +98,11 @@ abstract contract OptimisticOracleIntegrator is
     // Setter Functions
 
     /// @inheritdoc IOptimisticOracleIntegrator
-    function setDefaultCurrency(address _newCurrency)
+    function setDefaultCurrency(address _newCurrency, uint _newBond)
```
Zitzak: Not strictly necessary but could be changed to reflect additional usecase. `setDefaultCurrencyAndBond`

Zitzak (approved):
LGTM. The one comment is optional I would say

src/modules/logicModule/abstracts/oracleIntegrations/UMA_OptimisticOracleV3/OptimisticOracleIntegrator.sol
0xNuggan: Fair enough, changed it

FHieser (approved):
The implementation seems valid, but the bond basically means min amount of Tokens that will be distributed correct?

0xNuggan:
No, it's an amount that gets taken from the asserter in case of dispute, and as fee

PR #524, opened 2024-06-03
[SC-503] Bug(H-8): Attacker may DoS KPI rewarder by filling out the staking queue

0xNuggan:
## What has been done
- Added a minimum stake requirement to the stake functions. It starts out at zero, but the workflow owner can increase it if they deem it necessary
- Updated tests


FHieser (approved):

0xNuggan:
tagging @Zitzak  since it's high and we want 2 reviews

Zitzak:
Just for my understanding, instead of adding a mechanism to manage the queue (as the auditors suggested) you introduced a cost to staking (a min amount) which should also deter sybil attacks. Did I get that right? @0xNuggan

Zitzak (approved):
LGTM

0xNuggan:
They suggested both options, none of them perfect. Adding a minimum stake mitigates the vector somewhat. The other solution would have required extra work and changes

PR #526, opened 2024-06-03
[SC-476] Bug: Fix grantModuleRole bug

0xNuggan:
## What has been done
- Corrected the gating of grantModuleRole so only the admin of that role can modify it
- Merged DEFAULT_ADMIN_ROLE and ORCHESTRATOR_OWNER_ROLE into one. Now the owner has full admin control, and modules can selectively opt out on a per-role basis. Before this the OWNER had to grant themselves ADMIN role first (which they could do anyway), so the trust assumptions don't change.

### DIFF
--- a/src/modules/authorizer/role/AUT_Roles_v1.sol
+++ b/src/modules/authorizer/role/AUT_Roles_v1.sol
@@ -106,17 +106,7 @@ contract AUT_Roles_v1 is
         // It is defined in the AccessControl contract and identified with bytes32("0x00")
         // Modules can opt out of this on a per-role basis by setting the admin role to "BURN_ADMIN_ROLE".
 
-        // Set up OWNER role structure:
-
-        // -> set OWNER as admin of itself
-        //_setRoleAdmin(DEFAULT_ADMIN_ROLE, DEFAULT_ADMIN_ROLE);
-        // -> set OWNER as admin of DEFAULT_ADMIN_ROLE
-        //_setRoleAdmin(DEFAULT_ADMIN_ROLE, DEFAULT_ADMIN_ROLE);
-
-        // Set up MANAGER role structure:
-        // -> set OWNER as admin of DEFAULT_ADMIN_ROLE
-        _setRoleAdmin(ORCHESTRATOR_MANAGER_ROLE, DEFAULT_ADMIN_ROLE);
-        // grant MANAGER Role to specified address
+        // Grant MANAGER Role to specified address
         _grantRole(ORCHESTRATOR_MANAGER_ROLE, initialManager);
 
         // If there is no initial owner specfied or the initial owner is the same as the deployer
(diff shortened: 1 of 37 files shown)


src/modules/base/Module_v1.sol
```diff
@@ -202,14 +213,14 @@ abstract contract Module_v1 is
 
     function revokeModuleRole(bytes32 role, address target)
         external
-        onlyOrchestratorOwner
```
FHieser, line 205: is onlyOrchestratorOwner still a thing then?
If it is essentially the same as onlyModuleRoleAdmin(role) then shouldnt the modifier be replaced everywhere?
I saw it in a few other modules still

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -50,11 +50,7 @@ contract AUT_Roles_v1 is
 
     //--------------------------------------------------------------------------
     // Storage
-
-    // Stored for easy public reference. Other Modules can assume the following roles to exist
-    bytes32 public constant ORCHESTRATOR_OWNER_ROLE = "0x01";
     bytes32 public constant ORCHESTRATOR_MANAGER_ROLE = "0x02";
```
FHieser, line 53: Also wouldnt the ORCHESTRATOR_MANAGER_ROLE move up to 0x01?

FHieser (changes requested):
See comments

src/modules/authorizer/role/AUT_Roles_v1.sol
0xNuggan: No, because I'm going to remove it in the next PR :)

src/modules/base/Module_v1.sol
0xNuggan: I get your point, but I do think the modifiers serve two different functions: onlyModuleRoleAdmin() checks for the admin of a specific role, which starts out as the owner, but can also be transferred to somebody else or burned entirely. onlyOrchestratorOwner basically defers authorization to the general admin, without defining a specific role or allowing that kind of mechanism. Does that make sense? Do you think it would be clearer if I renamed the modifier to onlyOchestratorAdmin ?

0xNuggan: After discussion in the call I renamed all mentions of OrchestratorOwner into OrchestratorAdmin (except in the scripts, where I think having an user be called "owner" still makes sense)

0xNuggan:
The refactor ended up touching a lot of files, so it may make sense to review commit for commit @marvinkruse @Zitzak

FHieser (approved):

Zitzak:
Ahh I was just to late, apologies. Tried to keep up with the PRs but had to focus on other stuff as well.

I was just going to finish the review with the following message:
_LGTM except for the one comment. Only in the scripts the reference to the Orchestrator Owner still stands. Maybe create an issue to change this after the public audit start? Wdyt?_

@0xNuggan

PR #527, opened 2024-06-04
[SC-434] Chore: Update Readmes

marvinkruse:
This PR updates the readmes of the repo to represent the latest state and be prepared for a public release.
Changes:
* README and CONTRIBUTING revised
* SECURITY added
* Makefile updated to work accordingly (help command wasn't working)
* Update the test file that requires a working Sepolia RPC URL to work with a pre-set fallback URL (the official Sepolia RPC URL for Europe)
  * Goal: new users just wanting to run the tests can just run them without searching for any RPC URLs.
  * We can still locally overwrite it in `.env` files via `SEPOLIA_RPC_URL`, it's just a fallback

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,8 @@
 <img align="right" width="150" height="150" top="100" src="./assets/logo_circle.svg">
 
-# Contributing to the Inverter Network Contracts
+# Contribution Guidelines
 
-Thanks for your interest in improving the Inverter Network contracts! Your contributions play a vital role in enhancing the robustness and usability of the network. Open-source projects like ours thrive on community collaboration, and every contribution, big or small, pushes us closer to our shared vision.
+Thanks for your interest in improving the Inverter Network contracts! Your contributions play a vital role in enhancing the robustness and usability of the protocol. Open-source projects like ours thrive on community collaboration, and every contribution, big or small, pushes us closer to our shared vision.
 
 ## Using GitHub Issues
 GitHub issues are an essential tool for collaboration and tracking in our project. If you're an external contributor and come across something that isn't working as expected or have suggestions, concerns, or any other reasons to reach out:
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -1,8 +1,8 @@
 <img align="right" width="150" height="150" top="100" src="./assets/logo_circle.svg">
 
-# Contributing to the Inverter Network Contracts
+# Contribution Guidelines
```
Zitzak, line 3: Do you think it is a good idea to add a link to  the testing guidelines somewhere? Maybe not rn but soonish

README.md
```diff
@@ -14,9 +14,16 @@ The Inverter Network smart contracts are developed using the [foundry toolchain]
 
 ## Usage
 
-Common tasks are executed through a `Makefile`.
+Common tasks are executed through a `Makefile`. The most common commands are:
+* `make build` to compile the project.
+* `make test` to run the test suite.
+  * Note: _Some of our tests  require a working Sepolia RPC URL, as we test certain contracts via fork testing. We implemented fallbacks in these cases in the code directly, so they should work even without any RPC set in the environment. If that does not work, please set a working one via `export SEPOLIA_RPC_URL=https://rpc-url-here`._
```
Zitzak:
```suggestion
  * Note: _Some of our tests  require a working Sepolia RPC URL, as we test certain contracts via fork testing. We implemented fallbacks for these particular test cases in the code directly, so they will work even without any RPC set in the environment. In the unlikely case the tests do not work without RPC, please set a working RPC via `export SEPOLIA_RPC_URL=https://rpc-url-here`._
```

SECURITY.md
```diff
@@ -0,0 +1,46 @@
...
+
+We are committed to transparency and will make the results of our security reviews and audits available to the public, fostering trust and confidence in the security of our platform.
+
+## Security Guideline
+
+At Inverter Network, we adhere to a comprehensive Security Guideline that outlines our best practices and approach to ensuring the security of our smart contracts. This guideline is a public document that we follow at all times. You can find our Security Guideline [here](https://docs.google.com/document/d/1CZgM9OEuibNrimbNeActve5n9ro3Ydu03OfSnZfRo_s).
+
+We encourage you to review our Security Guideline to understand our commitment to security and the measures we have in place to protect our users and the integrity of our platform.
+
+## Bug Bounty Program
```
Zitzak, line 37: Maybe we can move the bug bounty part a bit more upwards, closer to the actual reporting header. In this way we can leave out 1 of the 2 mentions of "You will be compensated" within the sections _Reporting Security Issues_ & _Responsible Disclosure_, as readers will know to read on to the Bug Bounty sections

Zitzak (changes requested):
2 Small changes, but otherwise looking good

CONTRIBUTING.md
marvinkruse: Yeah, once we polished that doc, I think that's a good idea. Should be fine if we add that within the next 2-3 weeks.

SECURITY.md
marvinkruse: Yes, this will be updated once we have something, as discussed today. Will leave it like this for now.

README.md
marvinkruse: Addressed via next commit.

marvinkruse:
@Zitzak I updated it according to your comment(s), and included tech spec as well as the new docs link.

Zitzak (approved):
LGTM

PR #533, opened 2024-06-05
[SC-564] Bug(H-5): Redo - single failed transfer may block all pp simple

FHieser:


src/modules/paymentProcessor/PP_Simple_v1.sol
```diff
@@ -142,4 +159,54 @@ contract PP_Simple_v1 is Module_v1, IPaymentProcessor_v1 {
...
+        //Delete the field
+        delete unclaimableAmountsForRecipient[client][token][sender];
+
+        //Call has to succeed otherwise no state change
+        IERC20(token).safeTransferFrom(client, paymentReceiver, amount);
+
+        emit TokensReleased(paymentReceiver, address(token), amount);
+
+        //Make sure to let paymentClient know that amount doesnt have to be stored anymore
+        IERC20PaymentClientBase_v1(client).amountPaid(token, amount);
```
0xNuggan: Would moving this up to before the transfer alter functionality? To avoid any weird callback shenanigans with malicious tokens that then call the client.

0xNuggan (approved):

PR #536, opened 2024-06-08
[SC-583] Bug(53): Incorrect Time Check in PP_Streaming_v1

marvinkruse:
**Status**: ⌛ Done, but waiting with merging until finding was confirmed by Hats and has a clear ID.

Finding can be found [here](https://github.com/hats-finance/Inverter-Network-0xe47e52c4fea05e555920f1dcdcc6fb8eca103eeb/issues/53).

----

I chose to get rid of the `type(uint).max` check overall, as I feel it doesn't really serve any purpose. If the goal is to limit users from starting a stream at `type(uint).max` (_as it is "too far into the future"_), they could still start it at `type(uint).max - 1`, so we would need to choose a reasonable limit, which I deem gets us into more trouble finding one. 

So let's just allow users to set this to anything they want, as long as the most important condition is ensured: `start + cliff <= end`. This ensures that (1) the `start` is not after the `end` (but can be at the same time to use Streaming PP for instant payouts), and (2) the `cliff` doesn't bring the whole thing to after the `end`. This is imho all we need, as the rest doesn't matter.

I also added a test that verifies all of this specifically, which should have existed beforehand to catch this in the first place.


FHieser (approved):

marvinkruse:
Thank you for the review. Waiting with merging until we have clear IDs for the findings.

PR #537, opened 2024-06-08
Chore: Add Contract Size Check to CI

marvinkruse:
**Status**: Done, waiting for reviews
---

In order to find out early about contracts being over the size limit, I added a Makefile command to verify it and added it to our CI. Now we will find out early if this is the case.

There are two important things here:
(1) I added two commands:
* `make list-size` (which just calls `forge build --sizes`) and prints it out, for us locally
* `make check-size` (which does the same, but only prints the ones that are over the limit) for the CI, but can be used locally as well ofc

(2) Per default, the forge build output is a list of all contracts. I wanted to only see the ones from `src/`, but there is no command that does only that. So my solution is to rename `test/` and `script/` to `.test/` and `.script/` prior to executing it and then naming them back afterward. This works file and does exactly what we need. Just be aware, that if the compilation takes a while, your IDE might be showing you Git changes, as the folders are removed etc. - just ignore as this will disappear once the command ran through. - I just think that this way it's easier to see which ones are actually over the limit or close, as a lot of Mocks are over the limit but don't matter to us.

Look how clean it is 😁
Screenshot 2024-06-08 at 18 54 37

---

I need people to test this locally, as it works for me on my Mac. @0xNuggan and @FHieser would be great, as we then have a second Mac that tested it and a Linux user (works in the CI, so Linux should be good).


marvinkruse:
Lol, just noticed now that this PR will never not fail in the CI probably. So it failing here is good, trust me! 😂 

Maybe we should wait with merging until we fixed the size issues in a separate PR 😂

0xNuggan:
I runs correctly for me, but `make list-size` exits with Error Code 1 (while `forge build --sizes` doesn't). The output of the command seems correct, though. I'm not sure if this is because some of the contracts are above limit?

FHieser:
works as promised for me
I also get this error msg for the list-size: make: *** [Makefile:38: list-size] Error 1
For the check-size its: make: *** [Makefile:47: check-size] Error 1

@marvinkruse  maybe include it in the make pre-commit too, as that ideally is the prerunner to the CI when were local

marvinkruse:
@FHieser @0xNuggan 
What you describe is expected behavior, `forge build --sizes` is exiting with a code 1 when there is any contract over the limit (like you suspected @0xNuggan). Your command line is just not showing that probably, as it doesn't return an explicit error message - which the Makefile adds per default whenever an underlying command exits with a non-zero code.

Screenshot 2024-06-11 at 14 13 13

So what you describe is just everything working as intended. I just pushed the change that reduces the optimizer runs to 1, which now should have both commands `make check-size` and `make list-size` to exit with a code 0 (no error). **Can you confirm that?**

marvinkruse:
Also added the `make check-size` command to the `make pre-commit` now. So whenever we run it locally, it will now test that prior to doing anything.

~~Additionally, I just added the contract size limit to each script, so we never run the script verification without the correct contract code size limit. We can not add it to the `foundry.toml`, as we have mocks that are over the limit, so the whole thing doesn't compile anymore.~~  This didn't work, removed it again. Adding it to pre-commit should be enough for now.

FHieser (approved):
All as discussed

Zitzak:
@marvinkruse please merge this into the `hats-fixes` branch

PR #540, opened 2024-06-25
[Audit] Macro Fixes - Omega Review

marvinkruse:
This PR will contain the fixes to the audit by [Macro](https://0xmacro.com/), to be reviewed by Team Omega.

### DIFF
--- a/src/external/fees/FeeManager_v1.sol
+++ b/src/external/fees/FeeManager_v1.sol
@@ -70,6 +70,9 @@ contract FeeManager_v1 is ERC165, IFeeManager_v1, Ownable2StepUpgradeable {
     mapping(address => mapping(bytes32 => Fee)) internal workflowIssuanceFees;
     mapping(address => mapping(bytes32 => Fee)) internal workflowCollateralFees;
 
+    // Storage gap for future upgrades
+    uint[50] private __gap;
+
     //--------------------------------------------------------------------------
     // Initialization
 
--- a/src/modules/authorizer/role/AUT_Roles_v1.sol
+++ b/src/modules/authorizer/role/AUT_Roles_v1.sol
@@ -109,6 +109,10 @@ contract AUT_Roles_v1 is
         internal
         onlyInitializing
     {
+        if (initialOwner == address(0)) {
+            revert Module__Authorizer__InvalidInitialOwner();
+        }
+
         // Note about DEFAULT_ADMIN_ROLE: The DEFAULT_ADMIN_ROLE has admin privileges on all roles in the contract. It starts out empty, but we set the orchestrator owners as "admins of the admin role",
         // so they can whitelist an address which then will have full write access to the roles in the system. This is mainly intended for safety/recovery situations,
         // Modules can opt out of this on a per-role basis by setting the admin role to "BURN_ADMIN_ROLE".
@@ -131,11 +135,7 @@ contract AUT_Roles_v1 is
 
         // If there is no initial owner specfied or the initial owner is the same as the deployer
 
-        if (initialOwner != address(0)) {
-            _grantRole(ORCHESTRATOR_OWNER_ROLE, initialOwner);
-        } else {
-            _grantRole(ORCHESTRATOR_OWNER_ROLE, _msgSender());
-        }
+        _grantRole(ORCHESTRATOR_OWNER_ROLE, initialOwner);
     }
 
     //--------------------------------------------------------------------------
--- a/src/modules/authorizer/role/AUT_Roles_v1.sol
+++ b/src/modules/authorizer/role/AUT_Roles_v1.sol
@@ -80,6 +80,14 @@ contract AUT_Roles_v1 is
         _;
     }
 
+    /// @notice Verifies that the owner being added is not the orchestrator
+    modifier noSelfOwner(bytes32 role, address who) {
+        if (role == ORCHESTRATOR_OWNER_ROLE && who == address(orchestrator())) {
+            revert Module__Authorizer__OrchestratorCannotHaveOwnerRole();
+        }
+        _;
+    }
+
     //--------------------------------------------------------------------------
     // Initialization
 
@@ -145,6 +153,20 @@ contract AUT_Roles_v1 is
         returns (bool)
     {
         return super._revokeRole(role, who);
+    }
+
+    /// @notice Overrides {_grantRole} to prevent having the Orchestrator having the OWNER role
+    /// @param role The id number of the role
+    /// @param who The user we want to check on
+    /// @return bool Returns if grant has been succesful
+    function _grantRole(bytes32 role, address who)
+        internal
+        virtual
+        override
+        noSelfOwner(role, who)
+        returns (bool)
+    {
+        return super._grantRole(role, who);
     }
 
     //--------------------------------------------------------------------------
--- a/src/modules/authorizer/role/AUT_Roles_v1.sol
+++ b/src/modules/authorizer/role/AUT_Roles_v1.sol
@@ -135,9 +135,7 @@ contract AUT_Roles_v1 is
         _setRoleAdmin(ORCHESTRATOR_MANAGER_ROLE, ORCHESTRATOR_OWNER_ROLE);
         // grant MANAGER Role to specified address
         _grantRole(ORCHESTRATOR_MANAGER_ROLE, initialManager);
-
-        // If there is no initial owner specfied or the initial owner is the same as the deployer
-
+        // grant OWNER Role to initial owner address
         _grantRole(ORCHESTRATOR_OWNER_ROLE, initialOwner);
     }
 
@@ -159,7 +157,7 @@ contract AUT_Roles_v1 is
     }
 
     /// @notice Overrides {_grantRole} to prevent having the Orchestrator having the OWNER role
-    /// @param role The id number of the role
+    /// @param role The id of the role
     /// @param who The user we want to check on
     /// @return bool Returns if grant has been succesful
     function _grantRole(bytes32 role, address who)
--- a/src/modules/authorizer/role/AUT_Roles_v1.sol
+++ b/src/modules/authorizer/role/AUT_Roles_v1.sol
@@ -58,6 +58,9 @@ contract AUT_Roles_v1 is
     bytes32 public constant BURN_ADMIN_ROLE =
         0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
 
+    // Storage gap for future upgrades
+    uint[50] private __gap;
+
     //--------------------------------------------------------------------------
     // Modifiers
 
--- a/src/modules/authorizer/role/AUT_Roles_v1.sol
+++ b/src/modules/authorizer/role/AUT_Roles_v1.sol
@@ -81,12 +81,7 @@ contract AUT_Roles_v1 is
     }
 
     //--------------------------------------------------------------------------
-    // Constructor and initialization
-
-    constructor() {
-        // make the BURN_ADMIN_ROLE immutable
-        _setRoleAdmin(BURN_ADMIN_ROLE, BURN_ADMIN_ROLE);
-    }
+    // Initialization
 
     /// @inheritdoc Module_v1
     function init(
@@ -109,6 +104,9 @@ contract AUT_Roles_v1 is
         // Note about DEFAULT_ADMIN_ROLE: The DEFAULT_ADMIN_ROLE has admin privileges on all roles in the contract. It starts out empty, but we set the orchestrator owners as "admins of the admin role",
         // so they can whitelist an address which then will have full write access to the roles in the system. This is mainly intended for safety/recovery situations,
         // Modules can opt out of this on a per-role basis by setting the admin role to "BURN_ADMIN_ROLE".
+
+        // make the BURN_ADMIN_ROLE immutable
+        _setRoleAdmin(BURN_ADMIN_ROLE, BURN_ADMIN_ROLE);
 
         // Set up OWNER role structure:
 
--- a/src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
+++ b/src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
@@ -111,6 +111,8 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
     /// @param who The address to grant the role to
     /// @return bool Returns if the role has been granted succesful
     /// @dev Overrides {_grantRole} from AccessControl to enforce interface implementation and threshold existence when role is token-gated
+    /// @dev Please note: current check for validating a valid token is not conclusive and could be
+    ///         circumvented through a callback() function
     function _grantRole(bytes32 role, address who)
         internal
         virtual
--- a/src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
+++ b/src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
@@ -98,8 +98,8 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
     /// @notice Grants a role to an address
     /// @param role The role to grant
     /// @param who The address to grant the role to
-    /// @return bool Returns if the role has been granted succesful
-    /// @dev Overrides {_grantRole} from AccessControl to enforce interface implementation and threshold existence when role is token-gated
+    /// @return bool Returns true if the role has been granted succesfully
+    /// @dev Overrides {_grantRole} from AUT_ROLES_v1 to enforce interface implementation and threshold existence when role is token-gated
     /// @dev Please note: current check for validating a valid token is not conclusive and could be
     ///         circumvented through a callback() function
     function _grantRole(bytes32 role, address who)
--- a/src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
+++ b/src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
@@ -92,7 +92,7 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
     //--------------------------------------------------------------------------
     // Overloaded and overriden functions
 
-    function hasRole(bytes32 roleId, address account)
+    function hasRole(bytes32 roleId, address who)
         public
         view
         virtual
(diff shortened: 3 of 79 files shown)


src/external/fees/FeeManager_v1.sol
```diff
@@ -70,6 +70,9 @@ contract FeeManager_v1 is ERC165, IFeeManager_v1, Ownable2StepUpgradeable {
     mapping(address => mapping(bytes32 => Fee)) internal workflowIssuanceFees;
     mapping(address => mapping(bytes32 => Fee)) internal workflowCollateralFees;
 
+    // Storage gap for future upgrades
+    uint[50] private __gap;
```
omega-audits, line 74: As long as no other contract inherits from FeeManager_v1, it is not necessary to add the storage gap

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -125,11 +138,7 @@ contract AUT_Roles_v1 is
 
         // If there is no initial owner specfied or the initial owner is the same as the deployer
```
omega-audits: comment is outdated

src/external/fees/FeeManager_v1.sol
omega-audits, line 74: If this contract is not inherited by another contract (and it is not), then adding a storage gap is not necessary or useful. This holds here, and for all the other contracts where the storage gap is added.

src/modules/authorizer/role/AUT_Roles_v1.sol
omega-audits: also the line 134 can be removed

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -149,6 +158,20 @@ contract AUT_Roles_v1 is
         return super._revokeRole(role, who);
     }
 
+    /// @notice Overrides {_grantRole} to prevent having the Orchestrator having the OWNER role
+    /// @param role The id number of the role
```
omega-audits:
```suggestion
    /// @param role The id of the role
```

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
```diff
@@ -89,42 +89,89 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
...
     }
 
     /// @notice Grants a role to an address
     /// @param role The role to grant
-    /// @param account The address to grant the role to
+    /// @param who The address to grant the role to
     /// @return bool Returns if the role has been granted succesful
-    /// @dev Overrides {_grantRole} from AccesControl to enforce interface implementation when role is token-gated
-    function _grantRole(bytes32 role, address account)
+    /// @dev Overrides {_grantRole} from AccessControl to enforce interface implementation and threshold existence when role is token-gated
```
omega-audits:
```suggestion
    /// @dev Overrides {_grantRole} from AUT_ROLES_v1 to enforce interface implementation and threshold existence when role is token-gated
```

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
```diff
@@ -89,42 +89,89 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
...
-            return super.hasRole(roleId, account);
+            return super.hasRole(roleId, who);
         }
     }
 
     /// @notice Grants a role to an address
     /// @param role The role to grant
-    /// @param account The address to grant the role to
+    /// @param who The address to grant the role to
     /// @return bool Returns if the role has been granted succesful
```
omega-audits:
```suggestion
    /// @return bool Returns  true if the role has been granted succesfully
```

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
```diff
@@ -89,42 +89,89 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
...
-            catch {
-                revert Module__AUT_TokenGated_Roles__InvalidToken(account);
+            // Make sure that a threshold has been set before granting the role
+            if (getThresholdValue(role, who) == 0) {
+                revert Module__AUT_TokenGated_Roles__TokenRoleMustHaveThreshold(
+                    role, who
+                );
+            }
+
+            // Check that address has code attached
```
omega-audits, line 119: Two remarks here:

1. it is better to do these checks as early in the process as possible, i.e. in `setThreshold`, and not here
2. These checks do not guarantee that `who` is a token, let alone that it is a useful token that can be used for access control. This is not something you can feasably check on-chain. You must trust the admin process here, and not implement these checks which may give a sense of false security. I'd recommend just doing the extcodesize check and do not bother with the duck typing

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
```diff
@@ -89,42 +89,89 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
...
     /// @param role The role to grant
-    /// @param account The address to grant the role to
+    /// @param who The address to grant the role to
     /// @return bool Returns if the role has been granted succesful
-    /// @dev Overrides {_grantRole} from AccesControl to enforce interface implementation when role is token-gated
-    function _grantRole(bytes32 role, address account)
+    /// @dev Overrides {_grantRole} from AccessControl to enforce interface implementation and threshold existence when role is token-gated
+    /// @dev Please note: current check for validating a valid token is not conclusive and could be
+    ///         circumvented through a callback() function
+    function _grantRole(bytes32 role, address who)
```
omega-audits, line 105: A general question here: is it ok/expected if different tokens get granted the same "tokenbased" role? This may be confusing (it is to me) - some of the code seems to expet this, other parts seem to be written with the idea that only a single address has this role

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
```diff
@@ -89,42 +89,89 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
...
+    function hasRole(bytes32 roleId, address who)
         public
         view
         virtual
         override(AccessControlUpgradeable, IAccessControl)
         returns (bool)
     {
         if (isTokenGated[roleId]) {
-            return _hasTokenRole(roleId, account);
+            return _hasTokenRole(roleId, who);
```
omega-audits: So this is very confusing. 

`_hasTokenRole` checks if the `who` address has a balance of tokens in one of the tokens that has the role `roleId`. 

So this new implemention of `hasRole` will give you this beahvior:

```
call `grantRole(myTokenRole, myToken)` with success
now `hasRole(myTokenRole, myToken)` returns false
```
but of course, internally, myToken does have the myTokenRole. You have no way of checking that directly anymore. You could check the threshold for that. Also, some other function, like `_revokeRole` and `_grantRole` in the AccessControl module call `hassRole`, and expect the standard behavior - not the one you implemented. 


The advice here is to have a consistent interface. IF you need the new behavior of `hasRole` that you implented here, then it is more consistent to have a behavior (for example) as this:
- setRole will revert if you try to set the role directly for a tokenGated role (Becase if an accoutn has that role depends on its token balance, so you cannot set it)
- implement a new function `setTokenForTokenGatedRole`

omega-audits: It seems to me that now, calling `revokeRole` here will _not_ remove the role from the roles mapping, which may lead to unexpected behavior later on.

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
```diff
@@ -89,42 +89,89 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
...
+        override
+        returns (bool)
+    {
+        if (isTokenGated[role]) {
+            // Set the threshold to 0 before revoking the role from the token
+            bytes32 thresholdId = keccak256(abi.encodePacked(role, who));
+            thresholdMap[thresholdId] = 0;
+            emit ChangedTokenThreshold(role, who, 0);
+        }
+        return super._revokeRole(role, who);
```
omega-audits, line 160: I do not think this will remove the `who` from the acocunts that have the `role`, because you overwritten `hasRole` which is called here: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/05f218fb6617932e56bf5388c3b389c3028a7b73/contracts/access/AccessControl.sol#L201

src/external/fees/FeeManager_v1.sol
omega-audits: Also even if added it is better to have it at the end of the file to ensure new storage variables are not being written after it

src/modules/paymentProcessor/PP_Simple_v1.sol
```diff
@@ -88,43 +102,112 @@ contract PP_Simple_v1 is Module_v1, IPaymentProcessor_v1 {
...
-
             emit PaymentOrderProcessed(
                 address(client),
                 recipient,
                 amount,
                 orders[i].createdAt,
                 orders[i].dueTo
             );
+
+            (bool success, bytes memory data) = token_.call(
```
omega-audits, line 128: Why do a low-level call here? The safeTransferFrom seemed a better choice here

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -610,6 +619,7 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
         uint _dueTo,
         uint _walletId
     ) internal {
+        ++numVestingWallets[client][_paymentReceiver];
```
omega-audits, line 622: Moving it up here does not seem necessary, and it makes the check in line 152 potentially less useful.

src/modules/logicModule/LM_PC_KPIRewarder_v1.sol
```diff
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 pragma solidity 0.8.23;
 
-import "forge-std/console.sol";
```
omega-audits, line 4: for this file and the staking module, we should probably do a separate review, perhaps after you merged these files

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
Zitzak: @0xNuggan @FHieser

Would love your input on this? What is expected behaviour here?

src/modules/logicModule/LM_PC_KPIRewarder_v1.sol
Zitzak: @marvinkruse how do you want to approach this?

src/external/fees/FeeManager_v1.sol
Zitzak: Is there any downside to having the storage gap here for now? Currently indeed there is no plan to inherit the contract but this would make it future proof.

In relation to the location of the variable, we have it at the end of the state variables which has the same effect right?

marvinkruse: Yeah I see the point, as it's also future-proof without this gap, as adding variables is always possible, even without a gap (it's just needed for inherited contracts). We can then probably remove it (and in the `Governor` I believe), as it does nothing but increase deployment cost. But we'll bring this up internally to quickly discuss tomorrow I think.

I also agree on the point of adding it as the last line of the contract though, just to be safe that we don't accidentally mess something up?

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
Zitzak: Adding these additional checks indeed are no guarantee. We have added this to the Natspec comment as well. We decided to add them here nonetheless, on recommendations of 0xMacro as it can never hurt.

In relation to doing the checks earlier in the process that makes sense

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
Zitzak: @0xNuggan

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
Zitzak: @0xNuggan

src/modules/paymentProcessor/PP_Simple_v1.sol
Zitzak: Purpose being that we don't want the whole Tx to revert, but instead add the amount to the unclaimableAmount in case of a failed transfer

src/external/fees/FeeManager_v1.sol
Zitzak: Assuming here that we will change this in all the contracts right @marvinkruse , moving the `__gap` to the end of the contract?

src/modules/logicModule/LM_PC_KPIRewarder_v1.sol
marvinkruse: Yeah, let's skip it for this review then and tackle it separately. I'll follow up with this on Slack 👍

src/external/fees/FeeManager_v1.sol
marvinkruse: > Assuming here that we will change this in all the contracts right @marvinkruse , moving the `__gap` to the end of the contract?

yeah, if we do this we'd do it in all contracts (and simultaneously - if we agree on that - remove it from the ones that don't need it; but there is probably a point to just leave it in for consistency and ease of maintenance)

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
0xNuggan: yes, hasRole will only return true if the balance is above the threshold, and that means there is no direct way to check if a token has a role or not (as you said, the way would be to go through the threshold). We can add that, it's a clear QoL improvement. On the other point, that's why we enforce there being a threshold that late in the process, so it regverts if somebody tries to set a tokengated role to a user. Keep in mind that token-gated and regular roles can coexist in a single authorizer

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
0xNuggan: Yes, it is expected that different tokens can have the same tokenbased role. Where do you feel that only one address can have it?

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
0xNuggan: hasTWe have this check at this point since several functions end up calling _grantRole. This way we ensure that if the role is token-gated, it will need to have been configured correctly (by going through grantTokenRoleFromModule). This way you prevent granting a  tokenGated to a particular user addreess

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
0xNuggan: It doesn't beacause you cannot revoke the role to a particular user if it is token-gated. You can only revoke it to the token wholesale. Do you think it would be better to be more explicit and revert the call if it is trying to revoke a role to an address without a threshold (and as such is probably either a user or an already removed token?)

src/external/fees/FeeManager_v1.sol
Zitzak: I would go with your decision here @marvinkruse. How should we proceed?

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
Zitzak: Okay so then we would leave it at this location @0xNuggan ?

src/external/fees/FeeManager_v1.sol
marvinkruse: We had a call today on this, and we all agreed that this makes sense, but we'd just prefer to leave the variable(s) at the top of the contract, so we won't change anything here.

We'll also leave the gap in the contracts that are not inherited from, as it's easier to maintain and only we suffer in terms of a slight deployment cost increment, no one else.
@Zitzak

Zitzak: Thank you for the update. Then we leave it as it is

src/modules/paymentProcessor/PP_Streaming_v1.sol
FHieser: It was a fix for Issue I - 10. It basically forces even invalid PaymentOrders to increase the vestingWallet count

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
Zitzak: image (1)

jellegerbrandy: this is the non-tokengated case though :)

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
jellegerbrandy: My point is here that you are breaking expected behavior, as `hasRole(role, account)` does not mean anymore that the account has that role. 
This is bound to give you problems at the certain point, because other functions may expect that hasRole means what it says on the surface. For example, _revokeRole and _grantRole  in the OZ modules do, and wil now not anymore behave as you would expect.

0xNuggan:
Thanks for the comment @jellegerbrandy  ! I removed the hasRole() override and moved the check to hasModuleRole, which is the function called by the Module modifiers to determine if a caller is authorized to do an action. This way when a role is token-gated, hasRole() checks if the token is in the enum, and hasTokenRole() checks if the target has the required balance.

jellegerbrandy:
Ok, looks good!

Zitzak:
This PR has gotten Omega's approval. See comment above, as well [comment in Slack](https://inverternetwork.slack.com/archives/C05UNJBGVNC/p1722362248721499). What would be the next step here @marvinkruse ?

marvinkruse:
Oh, sorry, only saw this now @Zitzak - I now created the PR #640 that applies the two commits that happened here (which are not on `dev` yet) to apply these to dev. I will close this PR once #640 is merged!

PR #542, opened 2024-06-27
Fix: Event cleanup

marvinkruse:
This PR will contain all of the event fixes we need for the new deployment, including:
* Hats fixes regarding events
* General findings that came up by @marvinkruse 

@marvinkruse will cherry-pick the hats stuff into here.

### DIFF
--- a/src/factories/interfaces/IModuleFactory_v1.sol
+++ b/src/factories/interfaces/IModuleFactory_v1.sol
@@ -35,7 +35,7 @@ interface IModuleFactory_v1 {
     /// @param metadata The registered Metadata
     /// @param beacon The registered Beacon
     event MetadataRegistered(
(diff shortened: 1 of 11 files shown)


marvinkruse:
Removed some indexing keywords from event parameters that just didn't make any sense to me (as you wouldn't search for these etc.)

@0xNuggan one question: is the `KpiToUse` something that is intended to be indexed or makes sense that it's indexed? Wasn't sure, you probably know best 👍

0xNuggan:
@marvinkruse  Yeah, KPI to use is something people may look for at some point

marvinkruse:
Merged all of the fixes from hats into this one that were relevant to the events and required for the indexer. My commit is the one that doesn't have an [ID] in front of it, where I remove some unnecessary ones. Check it out, then we can merge this and  deploy.

FHieser (approved):

0xNuggan (approved):

PR #544, opened 2024-06-27
Sc 680 task add feemanager address during init of governor

FHieser:


script/deployment/DeploymentScript.s.sol
```diff
@@ -253,21 +253,31 @@ contract DeploymentScript is Script {
         // Fetch the treasury address
         address treasury = vm.envAddress("TREASURY_ADDRESS");
 
+        console2.log(
+            "-----------------------------------------------------------------------------"
+        );
+
+        console2.log("Deploy Fee Manager \n");
+
+        feeManager = deployFeeManager.createProxy(communityMultisig); //@note owner of the FeeManagerBeacon will be the communityMultisig. Is that alright or should I change it to Governor? Needs more refactoring that way
```
Zitzak, line 262: Sounds right to me

script/external/DeployGovernor_v1.s.sol
```diff
@@ -43,14 +44,20 @@ contract DeployGovernor_v1 is Script {
...
-        return run(communityMultisig, teamMultisig, timelockPeriod);
+        return run(communityMultisig, teamMultisig, timelockPeriod, feeManager);
     }
 
     function run(
         address communityMultisig,
         address teamMultisig,
-        uint timelockPeriod
+        uint timelockPeriod,
+        address initialFeeManager
```
Zitzak, line 60: Why do we actually get the variable from our .env as well as from the function call? I'm seeing that this happens elsewhere also

src/external/governance/Governor_v1.sol
```diff
@@ -174,11 +177,10 @@ contract Governor_v1 is ERC165, IGovernor_v1, AccessControlUpgradeable {
 
     /// @inheritdoc IGovernor_v1
     function setFeeManager(address newFeeManager)
-        external
+        public
```
Zitzak: Why not keep it external?

Zitzak: Is there a need for it to be called from within this contract?

Zitzak (approved):
Only some question, no change required

script/external/DeployGovernor_v1.s.sol
FHieser: This is to test the scripts in the Make pre-commit
We basically mock the other addresses that way

src/external/governance/Governor_v1.sol
FHieser: I did that before the internal function
Will change it

script/external/DeployGovernor_v1.s.sol
Zitzak: Ahh okay, makes sense

PR #548, opened 2024-07-01
[SC-677] Bug(139):  Add validation to setting virtual issuance supply in FM_BC_Bancor

Zitzak:
## What has been done?
- The internal `_setVirtualSupply()`  function has been made virtual to be extended with functionalities in the implementation
- The extended function validates that the to-be-set supply after conversion does not result in 0
- Adapted test


marvinkruse:
@Zitzak The CI is failing, I believe this might be due to the changes introduced here (rejecting too many inputs for the tests, I believe)?

Zitzak:
Learned something new about using `bound` instead of `vm.assume` when not only having a narrow check

Zitzak:
If you have time, could you also take a look at this PR @0xNuggan, thank you

0xNuggan (approved):
LGTM

PR #552, opened 2024-07-02
[Fix: L-9] Add additional checks for token compliance

Zitzak:
Waiting on auditors for response on additional check to be introduced


Zitzak:
Link to Slack message
https://inverternetwork.slack.com/archives/C04S8MKSB5G/p1719924704369239?thread_ts=1719924535.073779&cid=C04S8MKSB5G

marvinkruse:
@Zitzak I asked, and they meant that we should check the return value to be of the right length (i.e. the length of a uint256). Had some further discussions with them and we agreed that doing a `.call()` makes most sense, and then verifying both the `success` bool and the `bytes data` return value, and analyze these. 

I just did this now as I was testing it out, could you check if you think it's good? :)

Good work otherwise! I just reordered your stuff to be above, made a bit more sense to me to first check whether there is code at the address and then calling it. Lmk if that's okay with you!

marvinkruse (approved):
Reviewed your code size check, good stuff!

Zitzak:
LGTM, thank you for following up and the jumping into the code @marvinkruse , will merge

PR #553, opened 2024-07-02
[SC-613] Bug(22): Add revert when rounded down fee is zero

Zitzak:
Ready for review

### DIFF
--- a/src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
+++ b/src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
@@ -323,13 +323,14 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
     }
 
     /// @dev Calculates the propotion of the fees for the given amount and returns them plus the amount minus the fees
+    ///         Reverts when the workflow fee + protocol fee > BPS, and if the fee amounts < 0 when
+    ///         with fee percentages > 0 which points to a rounding down to zero due to interger division
     /// @param _totalAmount The amount from which the fees will be taken
     /// @param _protocolFee The protocol fee percentage in relation to the BPS that will be applied to the totalAmount
     /// @param _workflowFee The workflow fee percentage in relation to the BPS that will be applied to the totalAmount
     /// @return netAmount The total amount minus the combined fee amount
     /// @return protocolFeeAmount The fee amount of the protocol fee
     /// @return workflowFeeAmount The fee amount of the workflow fee
-
     function _calculateNetAndSplitFees(
         uint _totalAmount,
         uint _protocolFee,
@@ -342,8 +343,19 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
         if ((_protocolFee + _workflowFee) > BPS) {
             revert Module__BondingCurveBase__FeeAmountToHigh();
         }
-        protocolFeeAmount = _totalAmount * _protocolFee / BPS;
-        workflowFeeAmount = _totalAmount * _workflowFee / BPS;
+        protocolFeeAmount =
+            _protocolFee > 0 ? _totalAmount * _protocolFee / BPS : 0;
+        workflowFeeAmount =
+            _workflowFee > 0 ? _totalAmount * _workflowFee / BPS : 0;
+
+        // Revert if calculated fee amount rounded down to zero
+        if (_protocolFee > 0 && protocolFeeAmount == 0) {
+            revert Module__BondingCurveBase__BuyAmountToLow();
+        }
+        if (_workflowFee > 0 && workflowFeeAmount == 0) {
+            revert Module__BondingCurveBase__BuyAmountToLow();
+        }
+
         netAmount = _totalAmount - protocolFeeAmount - workflowFeeAmount;
     }
 
--- a/src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
+++ b/src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
@@ -348,18 +348,18 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
         // Calculate protocol fee amount if applicable
         if (_protocolFee > 0) {
             protocolFeeAmount = _totalAmount * _protocolFee / BPS;
+            // Revert if calculated protocol fee amount rounded down to zero
+            if (protocolFeeAmount == 0) {
+                revert Module__BondingCurveBase__TradeAmountToLow();
+            }
         }
         // Calculate workflow fee amount if applicable
         if (_workflowFee > 0) {
             workflowFeeAmount = _totalAmount * _workflowFee / BPS;
-        }
-
-        // Revert if calculated protocol/workflow fee amount rounded down to zero
-        if (_protocolFee > 0 && protocolFeeAmount == 0) {
-            revert Module__BondingCurveBase__BuyAmountToLow();
-        }
-        if (_workflowFee > 0 && workflowFeeAmount == 0) {
-            revert Module__BondingCurveBase__BuyAmountToLow();
+             // Revert if calculated workflow fee amount rounded down to zero
+            if (workflowFeeAmount == 0) {
+                revert Module__BondingCurveBase__TradeAmountToLow();
+            }
         }
 
         netAmount = _totalAmount - protocolFeeAmount - workflowFeeAmount;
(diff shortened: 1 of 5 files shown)


Zitzak:
@0xNuggan @FHieser could you guys check if this would be a valid solution for the problem. Asking this because changing all the tests is quite some work so I would like to have a check before spending the time. Thanks

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -323,13 +323,14 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
     }
 
     /// @dev Calculates the propotion of the fees for the given amount and returns them plus the amount minus the fees
+    ///         Reverts when the workflow fee + protocol fee > BPS, and if the fee amounts < 0 when
```
0xNuggan: could you rewrite this? It's a bit confusing :)

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -342,8 +343,19 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
         if ((_protocolFee + _workflowFee) > BPS) {
             revert Module__BondingCurveBase__FeeAmountToHigh();
         }
-        protocolFeeAmount = _totalAmount * _protocolFee / BPS;
-        workflowFeeAmount = _totalAmount * _workflowFee / BPS;
+        protocolFeeAmount =
+            _protocolFee > 0 ? _totalAmount * _protocolFee / BPS : 0;
```
0xNuggan: I assume this is for gas saving? I'm wondering if it isn't more efficient to wrap everything into if(protocolFee > 0 ) brackets? Just a thouhgt though, the logic looks right

0xNuggan (changes requested):

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
Zitzak: So it was not for gas saving but to set the fee amounts to 0 in case fee percentage is not set. However what you describe is more readable and I didn't think about the fact that the variables are initiated with 0 anyways, making the ternary operators not needed.

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -342,8 +345,23 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
         if ((_protocolFee + _workflowFee) > BPS) {
             revert Module__BondingCurveBase__FeeAmountToHigh();
         }
-        protocolFeeAmount = _totalAmount * _protocolFee / BPS;
-        workflowFeeAmount = _totalAmount * _workflowFee / BPS;
+        // Calculate protocol fee amount if applicable
+        if (_protocolFee > 0) {
+            protocolFeeAmount = _totalAmount * _protocolFee / BPS;
```
FHieser, line 350: Couldnt you just set the protocolFeeAmount == 0 check here?

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -322,14 +322,17 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
             _getFeeManagerIssuanceFeeData(_selector);
     }
 
-    /// @dev Calculates the propotion of the fees for the given amount and returns them plus the amount minus the fees
+    /// @dev    Calculates the propotion of the fees for the given amount and returns them plus the amount minus the fees
+    ///         Reverts under the following two conditions:
+    ///             - if (workflow fee + protocol fee) > BPS
+    ///             - if protocol fee amount or workflow fee amounts < 0 given the fee percentage is not zero. This
```
FHieser: Is this really the way we want to go here?
Just reverting? -> I dont have a hard opinion on this, im just curious

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -342,8 +345,23 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
...
+            protocolFeeAmount = _totalAmount * _protocolFee / BPS;
+        }
+        // Calculate workflow fee amount if applicable
+        if (_workflowFee > 0) {
+            workflowFeeAmount = _totalAmount * _workflowFee / BPS;
+        }
+
+        // Revert if calculated protocol/workflow fee amount rounded down to zero
+        if (_protocolFee > 0 && protocolFeeAmount == 0) {
+            revert Module__BondingCurveBase__BuyAmountToLow();
```
FHieser: What about sell?
Isnt this function going in both directions?

test/modules/fundingManager/bondingCurve/abstracts/RedeemingBondingCurveBase_v1.t.sol
```diff
@@ -537,7 +576,14 @@ contract RedeemingBondingCurveBaseV1Test is ModuleTest {
...
 
-        _depositAmount = bound(_depositAmount, 1, 1e38);
+        _depositAmount = bound(
+            _depositAmount,
+            helperCalculateMinDepositAmount(
+                _collateralFee, _issuanceFee, _collateralFee
+            ),
+            1e38
+        );
+        emit hm(_depositAmount);
```
FHieser: hm

FHieser (commented):
Added a few questions

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
Zitzak: We can currently only go a few routes:
1. Set a minimum fee to ensure that users can't buy dust amounts, resulting in fee being rounded down. However this also have the consiquence where users pay more fee than they should, if they really want to buy a small amount
2. Set a minimum buy amount to ensure that the amount used for calculating the fee is big enough that the fee amount is not rounded down. This number however needs to be somewhat dynamic to ensure this, and forces the user to commit a certain amount.
3. Revert if the buy/sell amount has been so low that that taking a percentage fee is being rounded down to zero. This would at least protect agains the vulnerbility of repeated dust amount buys with no fee paid

So @0xNuggan and I agreed on going with the last route

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
Zitzak: You're right. Changing it

test/modules/fundingManager/bondingCurve/abstracts/RedeemingBondingCurveBase_v1.t.sol
Zitzak: Yeah so testing is still WIP, and will prob need that event still there. Hope you didn't spend to much time looking at the tests. Added in the PR description that those are still todos

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
Zitzak: Assuming you meant moving the check into the if statement codeblock

FHieser: yes

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -322,14 +322,17 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
             _getFeeManagerIssuanceFeeData(_selector);
     }
 
-    /// @dev Calculates the propotion of the fees for the given amount and returns them plus the amount minus the fees
+    /// @dev    Calculates the propotion of the fees for the given amount and returns them plus the amount minus the fees
```
marvinkruse: `propotion` -> `proportion`
(i know you didnt do this, but lets still fix that I think

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
marvinkruse: `< 0` - is this a typo for `==` or `>` or how can it be negative?

marvinkruse (approved):
all good besides the two comments about comments

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
Zitzak: Yes it is a typo indeed. Thanks

marvinkruse (approved):
thank you, good stuff!

PR #554, opened 2024-07-02
[SC-637] Bug(59): Convert values to prevent precision loss in static calculation

Zitzak:


FHieser (approved):
Yeah im not sure on the math here, but if I understood the newly added function call you do exactly what they mentioned in the issue.
As the tests are running through here is my approve

Zitzak:
Yeah I basically did the conversion similar as we did in other parts of the code, but haven't done here @FHieser

PR #557, opened 2024-07-03
[Fix: H-8] Remove staking queue

0xNuggan:
Since the staking queue was opening up too many attack vectors, this PR removes it completely, and in exchanges disables staking when there is an unresolved assertion. 
**NOTE:** That depending on the assertion liveness and dispute  windows, this could leave the contract closed for nontrivial amounts of time.

In addition to this change, the minimum assertion length should be lowered to two hours. This will be done in a separate PR for cleanliness


FHieser (approved):
LGTM

Zitzak:
Please cherry pick this commit over to the `hats-finance` branch after merging @marvinkruse

marvinkruse (approved):

PR #562, opened 2024-07-04
[SC-631] Bug(50):  any module can drain the fundingmanager of all funding

Zitzak:
## What has been done?
- Changed modifier on `transferOrchestratorToken` from `onlyOrchestratorAdmin` to `onlyPaymentClient`
- Removed all reference of `executeTx`
- Removed all reference of `executeTxFromModule`
- Adapted tests


test/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.t.sol
```diff
@@ -1555,25 +1558,54 @@ contract FM_BC_Bancor_Redeeming_VirtualSupplyV1Test is ModuleTest {
...
+            └── and the PaymentClient module is registered in the Orchestrator
+                └── then is should send the funds to the specified address
+                    └── and it should emit an event
     */
 
-    function testTransferOrchestratorToken(address to, uint amount)
-        public
-        virtual
-    {
+    function testTransferOrchestratorToken_FailsGivenCallerIsNotPaymentClient(
```
FHieser: You dont test this here, but in the module directly.
In here you would just make sure that the Modifier is in the correct place

test/modules/fundingManager/rebasing/FM_Rebasing_v1.t.sol
```diff
@@ -409,37 +413,51 @@ contract FM_RebasingV1Test is ModuleTest {
     //--------------------------------------------------------------------------
     // Tests: OnlyOrchestrator Mutating Functions
 
-    function testTransferOrchestratorToken(address to, uint amount) public {}
+    function testTransferOrchestratorToken_FailsGivenCallerIsNotPaymentClient(
```
FHieser: See comment above

test/modules/fundingManager/rebasing/FM_Rebasing_v1.t.sol
```diff
@@ -409,37 +413,51 @@ contract FM_RebasingV1Test is ModuleTest {
...
+        vm.expectRevert(IModule_v1.Module__OnlyCallableByPaymentClient.selector);
+        fundingManager.transferOrchestratorToken(to, amount);
+    }
 
-    function testTransferOrchestratorTokenFails(address caller, address to)
-        public
-    {
-        vm.assume(to != address(0) && to != address(fundingManager));
+    function testTransferOrchestratorToken_FailsGivenCallerIsPaymentClientButNotRegisteredModule(
+        address to,
```
FHieser, line 417: see comment above

test/modules/fundingManager/rebasing/FM_Rebasing_v1.t.sol
```diff
@@ -409,37 +413,51 @@ contract FM_RebasingV1Test is ModuleTest {
...
 
-        if (caller != address(_orchestrator)) {
-            vm.expectRevert(
-                IModule_v1.Module__OnlyCallableByOrchestrator.selector
-            );
-        }
-        vm.prank(caller);
-        fundingManager.transferOrchestratorToken(address(0xBEEF), 1);
+    function testTransferOrchestratorToken_WorksGivenCallerIsPaymentClientAndRegisteredModule(
+        address to,
```
FHieser, line 428: see comment above

test/modules/fundingManager/bondingCurve/FM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1.t.sol
```diff
@@ -104,15 +106,17 @@ contract FM_BC_Restricted_Bancor_Redeeming_VirtualSupplyV1UpstreamTests is
     }
 
     // Override to test deactivation
-    function testTransferOrchestratorToken(address to, uint amount)
-        public
-        override
-    {
+    function testTransferOrchestratorToken_WorksGivenCallerIsPaymentClientAndRegisteredModule(
+        address to,
```
FHieser, line 110: see comment above

FHieser (changes requested):
I would move the onlyPaymentclient test to the base module test
Seems good otherwise

test/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.t.sol
Zitzak: This probably feeds into this comment right @FHieser ?
https://inverternetwork.slack.com/archives/C04S8MKSB5G/p1719839850367339

And I am testing that the module is in place like this. I know we spook about this before but could you give me an example of how you envision this?

FHieser: Yep
The modifier that you test here is originally located in the Module directly, but you still test it here/ in the other fundingManagers (which means you test 3 times the same thing essentially)
Create a testModifier test in the Module_v1.t.sol (maybe via a new function you create in the ModuleV1Mock) and test it instead there. Then in the fundingManagers you only have to test if the modifier is in position in the first place.
Use the LM_PC_Bounties_v1.t.sol file I created as an example and search for 1. the modifier section (this will land in the Module.t.sol) and then 2. for on  ModifierInPosition test (this will land in the fundingmanagers). 
Does that make it clearer? @Zitzak

Zitzak: Yes will have a look at it, and adapt accordingly after our calls. Thank you @FHieser

0xNuggan (approved):
Looks good to me. This should bridge us over until we rethink the whole PaymentOrder flow

marvinkruse (approved):
good stuff!

PR #563, opened 2024-07-04
[SC-690] feat: remove transferOrchestratorToken override

Zitzak:
## What has been done?
- We have restructured the flow of how tokens get drawn from the FM. This resulted in that there is no need for the restriction anymore
## Issue and PR
- https://linear.app/inverter/issue/SC-631/bug50-any-module-can-drain-the-fundingmanager-of-all-funding-tokens
- https://github.com/InverterNetwork/contracts/pull/562


0xNuggan (changes requested):
You can also remove the error defined in line 44. Apart from that, it looks good to me!

Zitzak:
That error is still being used by the override of function `mintIssuanceTokenTo` @0xNuggan :)
https://github.com/InverterNetwork/contracts/blob/4baafa12ce39696baa20be8c6977a6a8fa32e1cd/src/modules/fundingManager/bondingCurve/FM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1.sol#L87

0xNuggan (approved):
You are absolutely right, I don't know how I missed that :)

marvinkruse (approved):

PR #564, opened 2024-07-05
[SC-643] Bug(70): Staking_v1 can be used for pyramid scheme

FHieser:


src/factories/ModuleFactory_v1.sol
```diff
@@ -141,12 +141,25 @@ contract ModuleFactory_v1 is
...
+        IModule_v1(proxy).init(orchestrator, metadata, configData);
+
+        return proxy;
+    }
+
+    function createModuleProxy(
+        IModule_v1.Metadata memory metadata,
+        IOrchestrator_v1 orchestrator,
+        IOrchestratorFactory_v1.WorkflowConfig memory workflowConfig
+    ) public returns (address) {
```
marvinkruse, line 164: How are these changes relevant to the issue?

marvinkruse: Just to resolve this: It's necessary as we need to know about the funding manager token in the staking contract (module), so a dependency injection issue resolved here. Thank you for explaining (via Huddle) @FHieser - closing this comment

marvinkruse (approved):

PR #565, opened 2024-07-08
[SC-696] (Bug148) DataAssertionResolved(...) event emitted even when assertion is not resolved truthfully

0xNuggan:
Moved the code for the event emission and adapted test

### DIFF
--- a/src/modules/logicModule/abstracts/oracleIntegrations/UMA_OptimisticOracleV3/OptimisticOracleIntegrator.sol
+++ b/src/modules/logicModule/abstracts/oracleIntegrations/UMA_OptimisticOracleV3/OptimisticOracleIntegrator.sol
@@ -233,17 +233,17 @@ abstract contract OptimisticOracleIntegrator is
 
         DataAssertion memory dataAssertion = assertionData[assertionId];
 
-        emit DataAssertionResolved(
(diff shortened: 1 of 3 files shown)


src/modules/logicModule/abstracts/oracleIntegrations/UMA_OptimisticOracleV3/OptimisticOracleIntegrator.sol
```diff
@@ -233,17 +233,17 @@ abstract contract OptimisticOracleIntegrator is
...
-            dataAssertion.asserter,
-            assertionId
-        );
-
-        // If the assertion was true, then the data assertion is resolved.
+        // If the assertion was true, then the data assertion is resolved and an event is emitted.
         if (assertedTruthfully) {
             assertionData[assertionId].resolved = true;
+
+            emit DataAssertionResolved(
```
FHieser, line 240: hm
should we have an event then if its not asserted?

0xNuggan: Upon further discussion, this finding is invalid. I'll update the linear issue

PR #568, opened 2024-07-08
[SC-641] (Bug 65) LM_PC_KPIRewarder_v1 can be set as a callback address to another assertion in order to set assertionPending = false

0xNuggan:
## What has been done
- assertionResolvedCallback() now reverts if the assertionId doesn't exist locally.
- Added check to test suite

**Note:** I thought about including the check in the OptimisiticOracleIntegrator instead of the KPIRewarder, but ended up deciding against it, since the vulnerability arises from changing the state of assertionResolved. One could imagine an OptimisiticOracleIntegrator Module that is actually designed to receive callbacks it didn't create itself.

### DIFF
--- a/test/modules/logicModule/LM_PC_KPIRewarder_v1.t.sol
+++ b/test/modules/logicModule/LM_PC_KPIRewarder_v1.t.sol
@@ -5,6 +5,9 @@ import "forge-std/console.sol";
 
 // External Libraries
 import {Clones} from "@oz/proxy/Clones.sol";
+import {ClaimData} from
+    "@lm/abstracts/oracleIntegrations/UMA_OptimisticOracleV3/optimistic-oracle-v3/ClaimData.sol";
+import {IERC20} from "@oz/token/ERC20/IERC20.sol";
 
 // Internal Dependencies
 import {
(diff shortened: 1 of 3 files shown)


test/modules/logicModule/LM_PC_KPIRewarder_v1.t.sol
```diff
@@ -1184,6 +1189,64 @@ contract LM_PC_KPIRewarder_v1_assertionresolvedCallbackTest is
         );
         kpiManager.assertionResolvedCallback(createdID, true);
     }
+
+    function test_RevertWhen_TheAssertionIdDoesNotExist(
```
Zitzak, line 1193: Not critical, especially given the time constraints but we should start writing our test function prototypes as defined by the tesing handbook, i.e. `testAssertionResolvedCallback_FailsGivenCondition...`

Zitzak (approved):
LGTM

test/modules/logicModule/LM_PC_KPIRewarder_v1.t.sol
Zitzak: Ehhmm apologies, but looking at the test handbook again, maybe I'm writing my test function prototypes wrong, not you? 😅  I have actually looked it up multiple times but now doubting if I misunderstood what was written there all along. Bit confused

PR #574, opened 2024-07-08
[SC-674] Bug(133): Add warning to natspec FM_Rebasing_v1

Zitzak:


FHieser (approved):
Although its a bit weird, that we dont have a proper Interface here

Zitzak:
I agree @FHieser, created an issue for refactoring

PR #575, opened 2024-07-09
[SC-698] Bug(155): Enforce closed curve during virtual supply change

Zitzak:


0xNuggan (approved):
LGTM. When we check coverage the script may complain that we don't check failsWithOnlyBuyOpen and failsWithOnlySellOpen explicitly, but given the time constraints right now I think this is good to go

PR #577, opened 2024-07-09
[SC-665] Bug(117): Hash motionId in AUT_EXT_VotingRoles

Zitzak:


src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
```diff
@@ -330,12 +331,12 @@ contract AUT_EXT_VotingRoles_v1 is IAUT_EXT_VotingRoles_v1, Module_v1 {
...
     }
 
-    function executeMotion(uint motionId) external {
+    function executeMotion(bytes32 motionId) external {
         // Get pointer to the motion.
         Motion storage motion_ = motions[motionId];
 
         // Revert if motionId invalid.
-        if (motionId >= motionCount) {
+        if (motion_.startTimestamp == 0) {
```
Zitzak, line 339: These checks are the only once I replaced from the main logic. @FHieser jfyi. I couldn't find a way to change the startTime so I went with this to check validity

0xNuggan (approved):

PR #578, opened 2024-07-09
[SC-609] Bug(16): ensureValidPaymentOrder missing start and end range

FHieser:


src/modules/paymentProcessor/PP_Simple_v1.sol
```diff
@@ -184,6 +184,17 @@ contract PP_Simple_v1 is Module_v1, IPaymentProcessor_v1 {
         _claimPreviouslyUnclaimable(client, token, receiver);
     }
 
+    /// @inheritdoc IPaymentProcessor_v1
+    function validPaymentOrder(
+        IERC20PaymentClientBase_v1.PaymentOrder memory order
+    ) external view returns (bool valid) {
```
marvinkruse: You are initializing a variable `valid` here via the return statement, which we usually only do if we actually use it. The other function you added doesn't do this, so just to be consistent I think this should just be `returns (bool)`?

src/modules/paymentProcessor/IPaymentProcessor_v1.sol
```diff
@@ -102,4 +103,11 @@ interface IPaymentProcessor_v1 {
         address token,
         address receiver
     ) external;
+
+    /// @notice Function that checks if the given PaymentOrder was valid
+    /// @param order The IERC20PaymentClientBase_v1 Order that needs to be checked
+    /// @return valid Bool if the Payment Order is valid
+    function validPaymentOrder(
+        IERC20PaymentClientBase_v1.PaymentOrder memory order
+    ) external view returns (bool valid);
```
marvinkruse: Same as the comment I left in `PP_Simple_v1`

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -406,6 +384,15 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
         return paymentReceiverStreams;
     }
 
+    /// @inheritdoc IPaymentProcessor_v1
+    function validPaymentOrder(
+        IERC20PaymentClientBase_v1.PaymentOrder memory order
+    ) external view returns (bool valid) {
```
marvinkruse: Same as the comment I left in `PP_Simple_v1`

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -629,56 +616,49 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
-        uint _cliff,
-        uint _end
+        address _client,
+        IERC20PaymentClientBase_v1.PaymentOrder memory _order,
+        uint _streamId
     ) internal {
-        ++numStreams[client][_paymentReceiver];
+        ++numStreams[_client][_order.recipient];
+
+        //@todo
```
marvinkruse: Left a `@todo` here

marvinkruse (changes requested):
Just a little comment, otherwise great!

src/modules/paymentProcessor/PP_Streaming_v1.sol
FHieser: nooooooooooooo

marvinkruse (approved):

PR #581, opened 2024-07-09
[SC-650] Bug(84): Redo - Update contract dependencies to upgradeable versions

marvinkruse:
Redoing PR #571 because of some issues with the ElasticReceiptToken. Anything regarding the ElasticReceiptToken will be handled in a separate PR.

### DIFF
--- a/src/external/forwarder/TransactionForwarder_v1.sol
+++ b/src/external/forwarder/TransactionForwarder_v1.sol
@@ -19,7 +19,7 @@ import {Context} from "@oz/utils/Context.sol";
  *          transactions (multi-call), facilitating complex, multi-step interactions within a single
  *          transaction.
  *
- * @dev     Integrates {ERC2771Forwarder} and Context to manage and relay meta transactions.
+ * @dev     Integrates {ERC2771Forwarder} and {Context} to manage and relay meta transactions.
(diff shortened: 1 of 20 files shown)


src/external/forwarder/TransactionForwarder_v1.sol
```diff
@@ -34,9 +34,6 @@ contract TransactionForwarder_v1 is
     ERC2771Forwarder,
     Context
 {
```
Zitzak, line 36: We are we removing the `__gap` here. Don't we need it to for the possible upgrade?

Zitzak (commented):
LGTM, only unsure why the gap was removed

marvinkruse:
> LGTM, only unsure why the gap was removed

I removed it there as it's not an upgradeable contract, and thus doesn't serve any purpose there.

src/external/forwarder/TransactionForwarder_v1.sol
marvinkruse: It's a non upgradable contract, at least thats how we built it

Zitzak: Thank you, I'll approve

Zitzak (approved):

PR #583, opened 2024-07-09
Chore: Emit collateral token 

fabianschu:


marvinkruse (approved):

marvinkruse:
looks good to me, when merging just make sure to squash it as "Chore: Emit collateral token" not "[SC-702] Chore: Emit collateral token", just so it's cleaner 🙏 The linear issue id is just not very useful for anyone else but us (and once the PR is merged we don't need it anymore, at least not in the title)

PR #585, opened 2024-07-10
[SC-597] Feat: Add Inverter Reverter to Upgrade Mechanism

FHieser:


src/factories/ModuleFactory_v1.sol
```diff
@@ -93,6 +94,8 @@ contract ModuleFactory_v1 is
     //--------------------------------------------------------------------------
     // Storage
 
+    address public immutable reverter;
```
marvinkruse, line 98: Currently, there is no real benefit of having it in this contract, as it's just stored. 

What we should do is to ensure that this `reverter` matches the one of the beacon whenever we `registerMetadata`, right? Otherwise, the user can never be sure we didn't just add a different one in the beacon without verifying themselves. If we check in `registerMetadata`, the user can be sure every future module will have the same `reverter`.

marvinkruse (changes requested):

src/factories/ModuleFactory_v1.sol
FHieser: Yeah I straight up forgot about it during coding.

src/factories/ModuleFactory_v1.sol
```diff
@@ -12,6 +12,7 @@ import {IOrchestratorFactory_v1} from
 import {IInverterBeacon_v1} from "src/proxies/interfaces/IInverterBeacon_v1.sol";
 
 // Internal Dependencies
+
```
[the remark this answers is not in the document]
FHieser: The linebreak here?
What triggers you here @marvinkruse ? xD

src/factories/ModuleFactory_v1.sol
marvinkruse: Thank you!

marvinkruse:
do not merge, I need to fix something here first and rebase `hats-fixes`.

marvinkruse:
Closing this one, will be done in #597

PR #591, opened 2024-07-11
[101] bug: add check to ensure enough projectCollateralFee in FM

Zitzak:


src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -353,6 +353,16 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
         virtual
         onlyPaymentClient
     {
+        if (
+            amount
+                > __Module_orchestrator.fundingManager().token().balanceOf(
```
marvinkruse: Maybe I'm just blind but why do we need to go the route of `__Module_orchestrator.fundingManager().token()` if this contract is literally the funding manager, can't we just call `token()` directly to save some gas?

It's not something you introduced, it's also done in line 366, but I'm just confused about this now that I see it.

marvinkruse (commented):
Otherwise good!

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
0xNuggan: Yes, I agree. We can also just do token() in the call below IMO. I think it may be a leftover from the time the token was stored in the orchestrator

Zitzak: Will change, thank you

marvinkruse (approved):
thank you good sir 🙏

PR #592, opened 2024-07-11
[SC-628] Bug(47) Caps on the Rebasing FM and ElasticReceiptToken

marvinkruse:
WIth the limited time that I had, I validated that the math is still fine if we go from 1 billion (1e9) for the max supply to double that (1e18), to allow of low value tokens and high decimal tokens to also be used properly.

I kept them constant, as messing with them can lead to invalid states that we can't control, so this way is the safest - and it's only a receipt token anyways.

Also left the deposit cap to be one power of ten lower, as otherwise it didn't work out anymore with the tests and maths.

### DIFF
--- a/src/modules/fundingManager/rebasing/FM_Rebasing_v1.sol
+++ b/src/modules/fundingManager/rebasing/FM_Rebasing_v1.sol
@@ -77,7 +77,9 @@ contract FM_Rebasing_v1 is
     //--------------------------------------------------------------------------
     // Constants
 
-    uint internal constant DEPOSIT_CAP = 100_000_000e18;
+    // This cap is one power of ten lower than the MAX_SUPPLY of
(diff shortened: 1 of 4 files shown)


marvinkruse:
Last commit was to adapt a test. Because of the limit that essentially went from `1_000_000_000` Tokens (at 18 decimals) to `1_000_000_000_000_000_000` Tokens for the max supply, we had a rounding difference of 1 wei in the tests. I'd say that's acceptable, given that the increased limit was bound to reduce the accuracy by a tiny bit for a huge reduction of the limitation.

The benefit is that now even tokens that are only worth $0.00000000001 would have a deposit limit of $1 million, which is perfectly fine imho.

0xNuggan (approved):
Yes, seems like an acceptable compromise

PR #594, opened 2024-07-11
[SC-610] Bug(17) Safely check signature via OZ

marvinkruse:


0xNuggan (approved):
Not an expert on this, but seems ok. As I understand, you removed the check for address 0 because OZ will revert if it fails to recover the address?

marvinkruse:
> Not an expert on this, but seems ok. As I understand, you removed the check for address 0 because OZ will revert if it fails to recover the address?

Yes, correct, amongst some other useful checks this is basically what OZ is doing [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9e3f4d60c581010c4a3979480e07cc7752f124cc/contracts/utils/cryptography/ECDSA.sol#L140)
```solidity
address signer = ecrecover(hash, v, r, s);
if (signer == address(0)) {
    return (address(0), RecoverError.InvalidSignature, bytes32(0));
}
```

Thanks for the review 🙏

PR #595, opened 2024-07-12
[SC-611] Bug(18) Reorg Issues in Factories Lead to Loss of Funds

marvinkruse:
Switched to a create2-based workflow deployment, to prevent the following attack:
* Alice deploys a workflow
* Reorg happens, so the chain rolls back to before this happened
* Bob deploys a workflow with the same modules but him as the owner
 * The address of the workflow/orchestrator/modules would be the same
* If Alice already shared the address for deposits, users could lose funds
* If Alice redeploys now with the same config, the address of the workflow/orchestrator/modules is different

Now with this change, we base the generation of the workflow/orchestrator/modules address on the message sender and an internal nonce. This way, Bobs deployment does have a different address and Alice can (at any time) still come and redeploy with the same inputs to get the address she had before the reorg.

### DIFF
--- a/src/factories/ModuleFactory_v1.sol
+++ b/src/factories/ModuleFactory_v1.sol
@@ -103,6 +103,10 @@ contract ModuleFactory_v1 is
     /// @dev Mapping of proxy address to orchestrator address.
     mapping(address => address) private _orchestratorOfProxy;
 
+    /// @dev Maps a users address to a nonce
+    ///      Used for the create2-based deployment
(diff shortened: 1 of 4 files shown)


0xNuggan (approved):
Looks good to me! Love learning new stuff when reviewing PRs :grin:

Zitzak (approved):
LGTM. Also helped that you described the purpose of the fix in the PR description. Thank you

PR #596, opened 2024-07-12
[SC-651] Bug(85) Add disclaimer to PP_Streaming

Zitzak:


src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -33,6 +33,16 @@ import {SafeERC20} from "@oz/token/ERC20/utils/SafeERC20.sol";
  *          multiple clients and recipients, integrated with error handling for
  *          payments and managing active streaming schedules and their cancellations.
  *
+ *          DISCLAIMER: Known Limitations
```
0xNuggan, line 36: Maybe explain in the note that this happens for paymentClients where users can trigger payment streams by themselves (like the stakingManager) and specifically warn of using this processor in that type of setup?

marvinkruse:
The [fix address collition in test](https://github.com/InverterNetwork/contracts/pull/596/commits/3b4ddb906e0ae28bf0b647bfb434bf4c24d1a8e2) commit is what needs to be fixed in a different commit within the hats findings PR. That's the one i referenced, which I'll handle once everything else is in, to prevent the force pushing annoyance. So this shouldn't be in here, even if I understand why you did it, right?

Zitzak:
Ahh I didn't know that. However does that mean that all hats fixes PRs, merging into `hats-fixes` will be merged with a failing CI? Because mine was failing here which is why I added it. @marvinkruse

marvinkruse:
Yeah, the `hats-fixes` CI is not running anyways for some reason (I'll check if it works again after the rebase later and fix there), and rerunning the CI usually solves it after 1-2 tries thankfully :D It's not ideal, but that way at least we don't have mixed up things in commits? @Zitzak

Zitzak:
> Yeah, the hats-fixes CI is not running anyways for some reason (I'll check if it works again after the rebase later and fix there), and rerunning the CI usually solves it after 1-2 tries thankfully :D It's not ideal, but that way at least we don't have mixed up things in commits? @Zitzak

Okay then I'll remove the commit for now @marvinkruse

marvinkruse:
Will push a fix commit to hats in a sec, then this one should run fine again

marvinkruse:
THe fix i pushed is already there and this passes again, so this one should be fine now, right?

Zitzak:
Could you give it one more approve @marvinkruse, thank you

marvinkruse (approved):

PR #586, opened 2024-07-10
[SC-633] Bug(55) TransparentProxy can't be paused

FHieser:
Is based on the #585
I would either wait until that PR is resolved to review this or to to look at the latest commit, as that one holds all the relevant changes

PR #599, opened 2024-07-12
[SC-633] Bug(55) TransparentProxy can't be pause

marvinkruse:
Redoing #586 because it was based on a PR that went into a different branch


marvinkruse:
Basically just using the `implementation()` which always points the currently used one by the `beacon`, which can be the `Reverter` contract -> this allows them to pause the contract if we paused the beacon of the module. The old way of `getImplementationAddress` would just get the actual implementation, and wouldn't pass the Reverter contract through.

0xNuggan (approved):

Zitzak (approved):
LGTM

PR #600, opened 2024-07-12
[DONT MERGE YET] Sanity check paymentToken when adding paymentOrder 

0xNuggan:
This PR started as a fix for [SC-666] (😏), but that issue turned out to be invalid. Nontheless, we still need to address the case the paymentToken isn't a correct ERC20 when adding the payment order, since if not, the call to balanceOf() in _ensureTokenBalance() would cause the processor to become stuck. This PR does that.

The branch is based on hats-fixes, but maybe it should be merged into dev?

### DIFF
--- a/src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
+++ b/src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
@@ -231,7 +231,7 @@ abstract contract ERC20PaymentClientBase_v1 is
         }
     }
 
-    function _ensureValidPaymentOrder(PaymentOrder memory order) private view {
+    function _ensureValidPaymentOrder(PaymentOrder memory order) private {
         if (!(orchestrator().paymentProcessor().validPaymentOrder(order))) {
             revert Module__ERC20PaymentClientBase__InvalidPaymentOrder();
         }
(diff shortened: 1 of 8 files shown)


marvinkruse:
(I'll check this out after dinner)

marvinkruse:
After checking, I believe this should be in `dev`. However, you based it on `hats-fixes` work that happened (where Felix changed the way `paymentOrders` are validated. I think it doesn't make sense to redo this based on dev now, so my proposal would be:
* We wait until `hats-fixes` are in `dev` (we'll likely merge them on Monday or Tuesday)
* We (I can also do that) rebase this one so it's based on `dev`, which at that point will be easier, as `dev` contains the required work

Does that make sense?

0xNuggan:
Makes total sense! Let's wait then

Zitzak:
@marvinkruse what is the plan for merging `hats-finance` branch into `dev`? After that we can merge this PR :)

marvinkruse:
> @marvinkruse what is the plan for merging `hats-finance` branch into `dev`? After that we can merge this PR :)

Once all fixes are ready, PR #607 will be merged with the hats findings to `dev`, then we can re-do this one, I'd say?

Zitzak:
@marvinkruse thank you for the clarification

PR #601, opened 2024-07-12
[SC-642] Bug(67) Complete Takeover of VotingRoles possible due to threshold logic

marvinkruse:
I just took this one over, hope you don't mind!

What I did:
* Established the rules of...
 * `threshold` can't be higher than voters count
 * `threshold` needs to be at least 2 if at least 3 voters exist
 * `threshold` can be 1 in the case of 1 or 2 voters existing
* Added a function to add/remove voters while setting a new threshold
* Adapted the tests 

Happy for feedback :)

### DIFF
--- a/src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
+++ b/src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
@@ -128,11 +128,8 @@ contract AUT_EXT_VotingRoles_v1 is IAUT_EXT_VotingRoles_v1, Module_v1 {
             revert Module__VotingRoleManager__EmptyVoters();
         }
 
-        // Revert if threshold higher than number of voters, i.e. threshold being
-        // unreachable.
-        if (threshold_ > votersLen) {
-            revert Module__VotingRoleManager__UnreachableThreshold();
-        }
+        // Revert if the threshold is set incorrectly
+        validateThreshold(votersLen, threshold_);
 
         // Revert if votingDuration outside of bounds.
         if (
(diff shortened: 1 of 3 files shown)


src/modules/authorizer/role/interfaces/IAUT_EXT_VotingRoles_v1.sol
```diff
@@ -39,7 +39,8 @@ interface IAUT_EXT_VotingRoles_v1 {
     error Module__VotingRoleManager__InvalidVoterAddress();
 
     /// @notice The threshold cannot exceed the amount of voters
-    error Module__VotingRoleManager__UnreachableThreshold();
+    ///         or be too low to be considered safe
```
Zitzak:
```suggestion
    ///         or be too low to be considered safe.
```

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
```diff
@@ -190,14 +187,9 @@ contract AUT_EXT_VotingRoles_v1 is IAUT_EXT_VotingRoles_v1, Module_v1 {
...
 
-    function setThreshold(uint newThreshold) external onlySelf {
-        // Revert if newThreshold higher than number of voters.
-        if (newThreshold > voterCount) {
-            revert Module__VotingRoleManager__UnreachableThreshold();
-        }
-
-        // Note that a threshold of zero is valid because a orchestrator can only be
-        // created by a voter anyway.
+    function setThreshold(uint newThreshold) public onlySelf {
```
Zitzak, line 190: Curious why this function and `addVoter` visibilities have been set to public instead of external

test/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.t.sol
```diff
@@ -1007,10 +1015,21 @@ contract AUT_EXT_VotingRoles_v1Test is ModuleTest {
 
         vm.startPrank(address(_governor));
         for (uint i; i < users.length; ++i) {
+            uint before = _governor.threshold();
```
Zitzak, line 1018: I find the use of the word `governor` pretty confusing here. Made me really wonder why we get the threshold for the Voting Roles Authorizer from our Governor before realising that this is prob still a left over from before the `Governor` contract was written.

Maybe a good idea to do a quick rename here? Wdyt?

Zitzak (approved):
LGTM. Only 1 nitpicky find is missing a `.`, and one questions around refactoring something for readability. Will approve already. Thank you for picking up the issue

test/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.t.sol
marvinkruse: yeah thats a good idea actually

marvinkruse: probably makes sense to do in a separate PR to `dev` though, just to keep this separate from the actual fix. will do that.

marvinkruse: Handled in #605

PR #602, opened 2024-07-12
[Audit] Merge Macro Review Fixes

marvinkruse:
Cherry-picked the four commits where we addressed comments of our fixes from Macro:
* #549
* #550 
* #552 

I did **not** include the following, which is a fix that we deployed, as I believe that it is not right anymore
* #549 
(_Idea was good, but doesn't work anymore as on `dev` we have `amountPaid` to not only take the amount but also a token address, which then turns this improvement into a problem. If we only do it at the end, we can't credit the individual amounts to the right token, so we shouldn't change anything. Was just as gas improvement anyways. So we just won't do this. I'll communicate that to Macro accordingly_).


Zitzak (approved):

marvinkruse:
@FHieser can you check whether dropping #549 was smart or not? to me it seemed like it's not a good idea anymore, as the `amountPaid` also takes the address now, and the tokens in the `for` look could have different addresses technically?

PR #603, opened 2024-07-15
[157] Bug: set minAmountOut to bigger than 0

Zitzak:


test/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.t.sol
```diff
@@ -1903,4 +1887,45 @@ contract FM_BC_Bancor_Redeeming_VirtualSupplyV1Test is ModuleTest {
...
+            ? (_bps * _bps)
+                / (_issuanceFee * (_bps - _collateralFee - _workflowFee)) + 1
+            : 1;
+
+        // Use the larger of the two minimum amounts
+        minAmount = minAmountFirstCall > minAmountIssuance
+            ? minAmountFirstCall
+            : minAmountIssuance;
+
+        // // Ensure the final amount after all fee deductions is at least 1
```
0xNuggan: outcommented

test/external/TransactionForwarder_v1.t.sol
```diff
@@ -34,7 +34,7 @@ contract TransactionForwarderV1Test is Test {
         uint signerPrivateKey
     ) public {
         // Restrict the signerKey to a space where it still should work
-        signerPrivateKey = bound(signerPrivateKey, 1, 2 ^ 128);
+        signerPrivateKey = bound(signerPrivateKey, 1, 2 ** 128);
```
0xNuggan: why this change? was it the formatter?

0xNuggan (changes requested):

test/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.t.sol
```diff
@@ -59,10 +59,10 @@ contract FM_BC_Bancor_Redeeming_VirtualSupplyV1Test is ModuleTest {
...
     uint internal constant MAX_SUPPLY = type(uint).max;
 
-    uint internal constant INITIAL_ISSUANCE_SUPPLY = 1;
-    uint internal constant INITIAL_COLLATERAL_SUPPLY = 3;
-    uint32 internal constant RESERVE_RATIO_FOR_BUYING = 333_333;
-    uint32 internal constant RESERVE_RATIO_FOR_SELLING = 333_333;
+    uint internal constant INITIAL_ISSUANCE_SUPPLY = 195_642_169e16;
+    uint internal constant INITIAL_COLLATERAL_SUPPLY = 39_097_931e16;
+    uint32 internal constant RESERVE_RATIO_FOR_BUYING = 199_800;
+    uint32 internal constant RESERVE_RATIO_FOR_SELLING = 199_800;
```
marvinkruse, line 65: Are these the new values that produce better initial pricing values? Is this supposed to be in this PR?

Zitzak: We have not tested with real world values before, resulting in (almost all) tests being run with `minAmountOut == 0`.

Initially I tried to come up with a function which would calculate a dynamic `depositAmount` such that the `minAmountOut != 0`. This took which I concluded that the time spend to make unrealistic starting parameters work is the actual issue.

The values here came from tests done by Rohan, after which the tests immediately started to succeed again.

So this a long answer to say, no they should not perse be in this PR, but going the other route ended up costing to much time, to make unrealistic values work :D

0xNuggan: They are values from the TEC launch; "real world values"

marvinkruse: Oh yeah, the connection to the `minAmountOut` wasn't immediately apparent to me, so this makes sense. Thank you for the explainer @Zitzak 🙏 (and @0xNuggan). Good stuff!

test/external/TransactionForwarder_v1.t.sol
Zitzak: So within [this PR](https://github.com/InverterNetwork/contracts/pull/553/files), while debugging I noticed that the `^` symbol is used as a "power to" symbol, but solidity actually sees it as a XOR operator.

I changed it within those tests but recognized it here as well, which is why I changed it.

However thinking about it now, this should prob not be within this PR (hats finance)

marvinkruse: yeah, it's a good find (even if it's just within a test). could you put that into its own PR if possible? just to keep it clean. I wouldn't be too annoyed though if you'd prefer to just keep it here :D

Zitzak: Tagging @FHieser here as well, as I think he added the symbol :)

Zitzak: Removed @marvinkruse

marvinkruse (approved):

0xNuggan (approved):

PR #604, opened 2024-07-15
Feat: Make ERC20Issuance minting rights whitelist-based

0xNuggan:
## What has been done
- Substituted `address minter` with a mapping, so more than one address can mint.
- Expanded tests


marvinkruse (approved):
lgtm

src/modules/fundingManager/bondingCurve/tokens/ERC20Issuance_v1.sol
```diff
@@ -92,9 +93,8 @@ contract ERC20Issuance_v1 is IERC20Issuance_v1, ERC20, Ownable {
 
     /// @notice Sets the address of the minter.
```
Zitzak: This function now serves for setting and removing. Maybe adjust the comment to reflect that change?

Zitzak: Seeing that you also did that in the interface

test/external/ERC20Issuance.t.sol
```diff
@@ -25,12 +27,22 @@ contract ERC20IssuanceTest is Test {
...
+        assertEq(token.allowedMinters(address(this)), true);
+    }
+
     /*
     test setMinter
     ├── When the caller is not the Admin
     │   └── It should revert
     └── When the caller is the Admin
-    └── It should set the new minter address
+    └── It should set the new minter address rights
```
Zitzak:
```suggestion
        └── It should set the new minter address rights
```

test/external/ERC20Issuance.t.sol
```diff
@@ -43,14 +55,42 @@ contract ERC20IssuanceTest is Test {
...
+    }
+
+    function test_setMinter_Idempotence(address minter, bool allowed) public {
+        // Sometimes we the initial state is that the address is allowed
+        if (uint(uint160(minter)) % 2 == 0) {
+            token.setMinter(minter, true);
+        }
+
+        // state before
+        bool allowedBefore = token.allowedMinters(minter);
```
Zitzak: Unused variable

Zitzak (changes requested):
Reviewed this PR so I would understand the other one about the removing the deployment fromt he FM_BC init. Found some very minor stuff

Zitzak:
I know in the case of paribu that they want to have the ability to mint new tokens directly, hence the added minter and the ownership of the token. However for other projects this might open the door to add new minters through a malicious owner of the contract without actually bricking the system, which would happen with other exploits where the owner has power over lets say changing a module.

So here are my questions:
1. Will we have an interface within our Control Room (or within the meme coin launcher for that matter) to deploy the issuance token such that we can pass the address as argument during FM_BC deployment. For Paribu I'm assuming this will just be a manual process right.
2. Might it make sense to suggest (either manually or within a Control Room flow) to pass ownership of the token to a multisig? 

@0xNuggan

Zitzak (approved):
LGTM. smart to make use of ERC20Capped as well as moving it to a different folder

Zitzak:
Apologies @marvinkruse, intended not to press the re-review button here

marvinkruse:
> Apologies @marvinkruse, intended not to press the re-review button here

all good, will just check it out anyways as I'm here :D

marvinkruse (approved):
lgtm, under the condition that the tests are not failing anymore :D

PR #606, opened 2024-07-16
[SC-623] Bug(38) Users may lose assets by depositing to Rebasing FM

marvinkruse:
**Why?**
See the [finding](https://github.com/hats-finance/Inverter-Network-0xe47e52c4fea05e555920f1dcdcc6fb8eca103eeb/issues/38). TL;DR: If the `supplyTarget` is zero, the contract can’t/doesn’t rebase due to a division-by-zero. This way, the `bitsPerToken` stays and if someone deposits something, the old user(s) can withdraw some of the new tokens, even if they shouldn’t. To fix this, the rebase always needs to happen, so usually 1 Wei needs to be left in the contract.

**What did I do?**
To fix this I make use of a virtual Wei, basically just telling the system that there is 1 Wei whenever a rebase would not happen otherwise. In detail, this is how the rebasing calculation is done prior to this:
```Solidity
_bitsPerToken = _activeBits() / supplyTarget;
```
and this is how it's done now
```Solidity
_bitsPerToken = supplyTarget == 0 ? activeBits : activeBits / supplyTarget;
```
Basically: if the supply target is zero, don't divide (_which is the same as dividing by 1 Wei_). That's the "virtual" wei.

Users will still have some `activeBits`, which is fine, we just don't want to confuse them, so I also adapted the `balanceOf` to return `zero` if there is nothing:
```Solidity
function balanceOf(address who) public view returns (uint) {
    // If the supply target is zero (i.e. no deposited tokens),
    // we hard return zero as the balance, as the user might have
    // dust bits (which happens if the funds are used via the workflow
    // and not withdrawn normally) which they can not use though.
    if (_supplyTarget() == 0) {
        return 0;
    }
    return _accountBits[who] / _bitsPerToken;
}
```

**Overall**
All the tests work, as there are no extra bits or wei that is left in the contract, everything can be used and withdrawn. So for the user experience this should be the best solution? **Please give me some feedback 😁🙏**

### DIFF
--- a/src/modules/fundingManager/rebasing/abstracts/ElasticReceiptTokenBase_v1.sol
+++ b/src/modules/fundingManager/rebasing/abstracts/ElasticReceiptTokenBase_v1.sol
@@ -210,7 +210,7 @@ abstract contract ElasticReceiptTokenBase_v1 is IRebasingERC20, ERC165 {
     /// @dev The supply target MUST never be zero or higher than MAX_SUPPLY,
     ///      otherwise no supply adjustment can be executed.
     /// @dev Has to be implemented in downstream contract.
-    function _supplyTarget() internal virtual returns (uint);
+    function _supplyTarget() internal view virtual returns (uint);
 
     //--------------------------------------------------------------------------
     // Public ERC20-like Mutating Functions
@@ -419,6 +419,13 @@ abstract contract ElasticReceiptTokenBase_v1 is IRebasingERC20, ERC165 {
 
     /// @inheritdoc IERC20
     function balanceOf(address who) public view returns (uint) {
+        // If the supply target is zero (i.e. no deposited tokens),
+        // we hard return zero as the balance, as the user might have
+        // dust bits (which happens if the funds are used via the workflow
+        // and not withdrawn normally) which they can not use though.
+        if (_supplyTarget() == 0) {
+            return 0;
+        }
         return _accountBits[who] / _bitsPerToken;
     }
 
@@ -550,15 +557,21 @@ abstract contract ElasticReceiptTokenBase_v1 is IRebasingERC20, ERC165 {
     ///      updates the bit-tokens conversion rate and the total token supply.
     function _rebase() private {
         uint supplyTarget = _supplyTarget();
+        uint activeBits = _activeBits();
 
         // Do not adjust supply if target is outside of valid supply range.
-        // Note to not revert as this would make transfer's impossible.
-        if (supplyTarget == 0 || supplyTarget > MAX_SUPPLY) {
+        // Note: to not revert as this would make transfer's impossible.
+        if ((supplyTarget == 0 && activeBits == 0) || supplyTarget > MAX_SUPPLY)
+        {
             return;
         }
 
         // Adjust conversion rate and total token supply.
-        _bitsPerToken = _activeBits() / supplyTarget;
+        // Note: If the supply target is zero and active bits exist (which means
+        //       that the deposited funds are used up), we use a supply target of
+        //       1, i.e. just use the active bits here.
+        _bitsPerToken =
+            supplyTarget == 0 ? activeBits : activeBits / supplyTarget;
         _totalTokenSupply = supplyTarget;
 
         // Notify about new rebase.


src/modules/fundingManager/rebasing/abstracts/ElasticReceiptTokenBase_v1.sol
```diff
@@ -210,7 +210,7 @@ abstract contract ElasticReceiptTokenBase_v1 is IRebasingERC20, ERC165 {
     /// @dev The supply target MUST never be zero or higher than MAX_SUPPLY,
     ///      otherwise no supply adjustment can be executed.
     /// @dev Has to be implemented in downstream contract.
-    function _supplyTarget() internal virtual returns (uint);
+    function _supplyTarget() internal view virtual returns (uint);
```
marvinkruse, line 213: This change was needed as `balanceOf()` was complaining otherwise (as it's a `view` function). Should be a `view` function anyways, so this is fine.

src/modules/fundingManager/rebasing/abstracts/ElasticReceiptTokenBase_v1.sol
```diff
@@ -550,15 +557,21 @@ abstract contract ElasticReceiptTokenBase_v1 is IRebasingERC20, ERC165 {
     ///      updates the bit-tokens conversion rate and the total token supply.
     function _rebase() private {
         uint supplyTarget = _supplyTarget();
+        uint activeBits = _activeBits();
 
         // Do not adjust supply if target is outside of valid supply range.
-        // Note to not revert as this would make transfer's impossible.
-        if (supplyTarget == 0 || supplyTarget > MAX_SUPPLY) {
+        // Note: to not revert as this would make transfer's impossible.
+        if ((supplyTarget == 0 && activeBits == 0) || supplyTarget > MAX_SUPPLY)
```
marvinkruse, line 564: Added `activeBits == 0` as an additional condition where this should `return` (i.e. not do anything), which is only the case if no deposits happened. If the funds were used by the workflow, `activeBits` is greater than zero. Was necessary to allow the `supplyTarget` to be zero.

src/modules/fundingManager/rebasing/abstracts/ElasticReceiptTokenBase_v1.sol
```diff
@@ -419,6 +419,13 @@ abstract contract ElasticReceiptTokenBase_v1 is IRebasingERC20, ERC165 {
 
     /// @inheritdoc IERC20
     function balanceOf(address who) public view returns (uint) {
+        // If the supply target is zero (i.e. no deposited tokens),
+        // we hard return zero as the balance, as the user might have
+        // dust bits (which happens if the funds are used via the workflow
+        // and not withdrawn normally) which they can not use though.
+        if (_supplyTarget() == 0) {
```
FHieser, line 426: This can be tested right?

src/modules/fundingManager/rebasing/abstracts/ElasticReceiptTokenBase_v1.sol
```diff
@@ -550,15 +557,21 @@ abstract contract ElasticReceiptTokenBase_v1 is IRebasingERC20, ERC165 {
...
             return;
         }
 
         // Adjust conversion rate and total token supply.
-        _bitsPerToken = _activeBits() / supplyTarget;
+        // Note: If the supply target is zero and active bits exist (which means
+        //       that the deposited funds are used up), we use a supply target of
+        //       1, i.e. just use the active bits here.
+        _bitsPerToken =
+            supplyTarget == 0 ? activeBits : activeBits / supplyTarget;
```
FHieser, line 574: Cant we test this somewhere?

FHieser (commented):
I really like this solution :+1: 
But what bugs me is that the tests are running through in the first place here.
That means that they were not set up tight enough.
Do you think we should do anything about that? @marvinkruse

0xNuggan:
I like the solution too. 
I think the reason the tests weren't failing is because the only test that "empties" the funding manager is locking in 1 wei at the beginning (line 282 in the test). If we comment that line & l.285 the test fails because during the burn (line 542 in ElasticReceiptTokenBase_v1) we divide by 0 again. I think this needs some extra testing

marvinkruse:
Thank you @FHieser and @0xNuggan - I'll work on some tests to cover this and also remove the 1wei deposits in the tests, as these are not needed and things should be handled differently there.

Will tag you again once that's ready! 🙏

marvinkruse:
Closing this one, the approach doesn't solve another underlying problem that we have: the ever-inflating bitsPerToken variable, because tokens are not withdrawn but used.

PR #607, opened 2024-07-16
[Audit] Merging Hats Fixes into Dev

marvinkruse:
Applied every fix from `hats-fixes` up until now (4pm, 17th of July). Will apply the rest once it's ready.

Individually checked each one and adapted it if necessary to work with the changes that happened since we diverged from `dev` (which was quite a bit).


0xNuggan:
Thinking about it, I wouldn't worry too much about the KPI Rewarder: The macro fixes have significant deletions which should bring it under the limit again. Should we push the contract size reduction until afterward?

marvinkruse:
> Thinking about it, I wouldn't worry too much about the KPI Rewarder: The macro fixes have significant deletions which should bring it under the limit again. Should we push the contract size reduction until afterward?

just to update here as there was some discussion on Slack: yes, the macro fixes, once merged, solve this ✅ 

edit: all good with the latest changes, size is good!

marvinkruse:
Mainly waiting for #606 (which I am working on), then we can merge this.

0xNuggan (approved):

PR #608, opened 2024-07-16
CI: Move checks into own steps

marvinkruse:
Moves the individual steps into their own task so that
* it's easier to see what exactly failed (currently it's just "CI", but could be scripts or tests or size)
* the rest still runs even if one fails (so we know whether the tests pass, even if the size is failing)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,24 +1,27 @@
 name: Lint
 
 on:
-  push:
-    branches:
-      - main
-      - dev
-  pull_request:
-    branches:
-      - '*'
+    push:
+        branches:
+            - main
+            - dev
+    pull_request:
+        branches:
+            - "*"
+
+env:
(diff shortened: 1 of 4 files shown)


Zitzak (commented):
LGTM, although not to much experience with it. To that extend, how can I test this setup?

Would be best for others to also review, which is why I don't make my approval explicit

marvinkruse:
> LGTM, although not to much experience with it. To that extend, how can I test this setup?
> 
> Would be best for others to also review, which is why I don't make my approval explicit

Yeah the only real way to test is to look at the GitHub actions, their files, and the output. Which to me all looks right, but please check. There is no other way though, this is nothing we can check locally.

marvinkruse:
If either @FHieser or @0xNuggan could briefly check this out and give me an approve if you're happy that would be awesome 🙏

FHieser (approved):

PR #612, opened 2024-07-16
Chore: Ensure events for state changes [SC-406]

fabianschu:
DONE:
* added a bunch of events and internal setters
* added back metadata as param for ModuleInitialized cause I realized it's needed to be able to compute the module id

DID NOT DO:
* ProtocolFeeCollected event for buy and sell orders (to save gas for end users)
* no events for modifications on `_bitsPerToken` and `_accountBits` on Elastic Receipt Token

### DIFF
--- a/src/external/fees/FeeManager_v1.sol
+++ b/src/external/fees/FeeManager_v1.sol
@@ -114,15 +114,15 @@ contract FeeManager_v1 is
         __Ownable_init(owner);
 
         // initial max fee is 10%
-        maxFee = 1000;
+        _setMaxFee(1000);
 
         if (_defaultCollateralFee > maxFee || _defaultIssuanceFee > maxFee) {
             revert FeeManager__InvalidFee();
         }
 
-        defaultProtocolTreasury = _defaultProtocolTreasury;
-        defaultCollateralFee = _defaultCollateralFee;
-        defaultIssuanceFee = _defaultIssuanceFee;
+        _setDefaultProtocolTreasury(_defaultProtocolTreasury);
+        _setDefaultCollateralFee(_defaultCollateralFee);
+        _setDefaultIssuanceFee(_defaultIssuanceFee);
     }
 
     //--------------------------------------------------------------------------
@@ -175,7 +175,8 @@ contract FeeManager_v1 is
         // In case workflow fee is set return it
         if (workflowCollateralFees[workflow][moduleFunctionHash].set) {
             return workflowCollateralFees[workflow][moduleFunctionHash].value;
-        } // otherwise return default fee
+        }
+        // otherwise return default fee
         else {
             return defaultCollateralFee;
         }
@@ -193,7 +194,8 @@ contract FeeManager_v1 is
         // In case workflow fee is set return it
         if (workflowIssuanceFees[workflow][moduleFunctionHash].set) {
             return workflowIssuanceFees[workflow][moduleFunctionHash].value;
-        } // otherwise return default fee
+        }
+        // otherwise return default fee
         else {
             return defaultIssuanceFee;
         }
@@ -231,8 +233,7 @@ contract FeeManager_v1 is
 
     /// @inheritdoc IFeeManager_v1
     function setMaxFee(uint _maxFee) external onlyOwner validMaxFee(_maxFee) {
-        maxFee = _maxFee;
-        emit MaxFeeSet(_maxFee);
+        _setMaxFee(_maxFee);
     }
 
     //---------------------------
@@ -265,20 +266,16 @@ contract FeeManager_v1 is
     function setDefaultCollateralFee(uint _defaultCollateralFee)
         external
         onlyOwner
-        validFee(_defaultCollateralFee)
     {
-        defaultCollateralFee = _defaultCollateralFee;
-        emit DefaultCollateralFeeSet(_defaultCollateralFee);
+        _setDefaultCollateralFee(_defaultCollateralFee);
     }
 
     /// @inheritdoc IFeeManager_v1
     function setDefaultIssuanceFee(uint _defaultIssuanceFee)
         external
         onlyOwner
-        validFee(_defaultIssuanceFee)
     {
-        defaultIssuanceFee = _defaultIssuanceFee;
-        emit DefaultIssuanceFeeSet(_defaultIssuanceFee);
+        _setDefaultIssuanceFee(_defaultIssuanceFee);
     }
 
     /// @inheritdoc IFeeManager_v1
@@ -288,7 +285,71 @@ contract FeeManager_v1 is
         bytes4 functionSelector,
         bool set,
         uint fee
-    ) external onlyOwner validFee(fee) {
+    ) external onlyOwner {
+        _setCollateralWorkflowFee(workflow, module, functionSelector, set, fee);
+    }
+
+    /// @inheritdoc IFeeManager_v1
+    function setIssuanceWorkflowFee(
+        address workflow,
+        address module,
+        bytes4 functionSelector,
+        bool set,
+        uint fee
+    ) external onlyOwner {
+        _setIssuanceWorkflowFee(workflow, module, functionSelector, set, fee);
+    }
+
+    //--------------------------------------------------------------------------
+    // Internal Functions
+
+    function getModuleFunctionHash(address module, bytes4 functionSelector)
+        internal
+        pure
+        returns (bytes32)
+    {
+        return keccak256(abi.encodePacked(module, functionSelector));
+    }
+
+    function _setDefaultProtocolTreasury(address _defaultProtocolTreasury)
+        internal
+        validAddress(_defaultProtocolTreasury)
+    {
+        defaultProtocolTreasury = _defaultProtocolTreasury;
+        emit DefaultProtocolTreasurySet(_defaultProtocolTreasury);
+    }
+
+    function _setWorkflowTreasury(address workflow, address treasury)
+        internal
+        validAddress(treasury)
+    {
+        workflowTreasuries[workflow] = treasury;
+        emit WorkflowTreasurySet(workflow, treasury);
+    }
+
+    function _setDefaultCollateralFee(uint _defaultCollateralFee)
+        internal
+        validFee(_defaultCollateralFee)
+    {
+        defaultCollateralFee = _defaultCollateralFee;
+        emit DefaultCollateralFeeSet(_defaultCollateralFee);
+    }
+
+    function _setDefaultIssuanceFee(uint _defaultIssuanceFee)
+        internal
+        validFee(_defaultIssuanceFee)
+    {
+        defaultIssuanceFee = _defaultIssuanceFee;
+        emit DefaultIssuanceFeeSet(_defaultIssuanceFee);
+    }
+
+    function _setCollateralWorkflowFee(
+        address workflow,
+        address module,
+        bytes4 functionSelector,
+        bool set,
+        uint fee
+    ) internal validFee(fee) {
         bytes32 moduleFunctionHash =
             getModuleFunctionHash(module, functionSelector);
 
@@ -301,14 +362,13 @@ contract FeeManager_v1 is
         );
     }
 
-    /// @inheritdoc IFeeManager_v1
-    function setIssuanceWorkflowFee(
+    function _setIssuanceWorkflowFee(
         address workflow,
         address module,
         bytes4 functionSelector,
         bool set,
         uint fee
-    ) external onlyOwner validFee(fee) {
+    ) internal validFee(fee) {
         bytes32 moduleFunctionHash =
             getModuleFunctionHash(module, functionSelector);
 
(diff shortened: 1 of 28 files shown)


marvinkruse:
A general thing I noticed is that some of the internal functions you added had an underscore in their naming (i.e. `_setDefaultProtocolTreasury`) and some didn't (i.e. `getModuleFunctionHash`) - was there are reason to this?

Also the variables are named sometimes with and without underscores, like these examples:
```Solidity
function _setDefaultProtocolTreasury(address _defaultProtocolTreasury) internal validAddress(_defaultProtocolTreasury) {
   // ...
}

function _setWorkflowTreasury(address workflow, address treasury) internal validAddress(treasury) {
   // ...
}
```

I don't wanna say that this is on you, as we don't have proper guidelines in place yet and we all probably just didn't do a good job so far ourselves, so this might be completely on us. But let's aim to no introduce any more unnecessary inconsistencies at least 😁

Could you double check and make sure of that? That would be amazing 🙏

fabianschu:
Where I added internal setters I prefixed them with an underscore. As for the parameter names, I just copied them from the external functions (didn't rename anything). This seemed the the most consistent approach to me that doesn't involve cleaning up existing inconsistencies. The examples you are stating stem from these existing inconsistencies. This is how `dev` looks like currently for example :
Screenshot 2024-07-17 at 11 02 32


Now I agree that this annoying and I could clean this up. The question is though, how clean do we want to go. I could either
1. clean up the functions that I touched
2. clean up the codebase

If we wanna go with 2. I suggest to do it in a new PR. What do you think? Anyways, looking at the codebase it is not clear to me where and how we want to use underscores, or phrased differently which convention we want to follow more generally. So if I clean up stuff, which conventions should I follow? @marvinkruse

marvinkruse:
Yeah that makes sense @fabianschu - then let's just not do this here, and just add your changes so they conform to whatever you find the contracts (which you did), otherwise it just gets too messy. So all good 🙏

marvinkruse:
I merged `dev` into this one as there were quite a few changes because of audit fix merges from Macro. You can see that in [this commit in this PR](https://github.com/InverterNetwork/contracts/pull/612/commits/fa44bc2a5f1d4178d7501556a15a7e3adb93f2cd) the `MinimumStakeSet` was removed again, which is because the whole queue concept of the `KPIRewarder` has been removed, and thus there is no concept of a minimum stake anymore. 

I then added another commit [here](https://github.com/InverterNetwork/contracts/pull/612/commits/4beeb617e430205a5ef45e7dd18028a849a125e3), to remove this event in the first place, otherwise it wouldn't be used.

Does that make sense? Hope that is okay with you?

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -181,7 +181,7 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
         // Set selling functionality to open if true. By default selling is false
         sellIsOpen = bondingCurveProperties.sellIsOpen;
 
-        emit CollateralTokenSet(_acceptedToken, collateralTokenDecimals);
+        emit ReserveTokenSet(_acceptedToken, collateralTokenDecimals);
```
marvinkruse: Is this naming something we all agree on? I know that we call it collateral token everywhere else, so this looks a bit "off" to me.

src/modules/logicModule/LM_PC_KPIRewarder_v1.sol
```diff
@@ -187,6 +187,7 @@ contract LM_PC_KPIRewarder_v1 is
 
         assertionPending = true;
 
+        // TODO
```
marvinkruse: Left a `TODO` here

marvinkruse (approved):

marvinkruse (dismissed):
Sorry, didn't mean to approve already, just left these two comments, nothing big

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
fabianschu: Yea, I was not sure about the naming. I just wanted to use the same event for both BondingCurve and Rebasing_FM when it comes to the token that both modules. My understanding is that the token that both modules take can be considered a reserve token, I wasn't so sure if the token can also be called a collateral token for the rebasing fm, but if so I'd also prefer to call it CollateralToken.

src/external/fees/FeeManager_v1.sol
```diff
@@ -265,30 +266,90 @@ contract FeeManager_v1 is
...
     /// @inheritdoc IFeeManager_v1
     function setDefaultIssuanceFee(uint _defaultIssuanceFee)
         external
         onlyOwner
+    {
+        _setDefaultIssuanceFee(_defaultIssuanceFee);
+    }
+
+    /// @inheritdoc IFeeManager_v1
+    function setCollateralWorkflowFee(
```
FHieser: Why exactly are you adding an internal function here?

src/factories/OrchestratorFactory_v1.sol
```diff
@@ -133,9 +133,9 @@ contract OrchestratorFactory_v1 is
         ) {
             revert OrchestratorFactory__InvalidBeacon();
         }
-
         beacon = beacon_;
         moduleFactory = moduleFactory_;
+        emit OrchestratorFactoryInitialized(address(beacon_), moduleFactory_);
```
FHieser, line 138: Where test?

src/factories/ModuleFactory_v1.sol
```diff
@@ -147,6 +147,8 @@ contract ModuleFactory_v1 is
                 initialMetadataRegistration[i], initialBeaconRegistration[i]
             );
         }
+
+        emit GovernorSet(_governor);
```
FHieser, line 151: Where test?

src/modules/base/IModule_v1.sol
```diff
@@ -19,9 +19,9 @@ interface IModule_v1 {
 
     /// @notice Module has been initialized.
     /// @param parentOrchestrator The address of the orchestrator the module is linked to.
-    /// @param moduleTitle The title of the module.
+    /// @param metadata The metadata of the module.
     event ModuleInitialized(
-        address indexed parentOrchestrator, string moduleTitle
+        address indexed parentOrchestrator, Metadata metadata
```
FHieser, line 24: Structs cant really be represented in the events as far as I know.
If you want to emit the whole metadata struct you have to extend the event to contain every field of it

src/modules/fundingManager/IFundingManager_v1.sol
```diff
@@ -37,6 +37,10 @@ interface IFundingManager_v1 {
     /// @param _amount The amount of underlying tokens transfered.
     event TransferOrchestratorToken(address indexed _to, uint _amount);
 
+    /// @notice Event emitted when collateral token has been set
+    /// @param token The token that serves as collateral token making up the curve's reserve
```
FHieser, line 41: Not every fundingManager is made up of a curve. With that the comment here is incorrect

FHieser: In general I dont know if I would use the word collateral here.
"Collateral is an item of value pledged to secure a loan." Is the first sentence on google that pops up for the word and I dont think thats the case for a rebasing or Deposivault FundingManager.
Maybe OrchestratorToken? (Dont know about that one though)

src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
```diff
@@ -116,12 +116,15 @@ contract LM_PC_RecurringPayments_v1 is
...
-        epochLength = abi.decode(configData, (uint));
+        uint newEpochLength = abi.decode(configData, (uint));
+        epochLength = newEpochLength;
 
         // revert if not at least 1 week and at most a year
         if (epochLength < 1 weeks || epochLength > 52 weeks) {
             revert Module__LM_PC_RecurringPayments__InvalidEpochLength();
         }
+
+        emit EpochLengthSet(newEpochLength);
```
FHieser, line 127: Where test?

src/modules/logicModule/LM_PC_Staking_v1.sol
```diff
@@ -239,16 +239,21 @@ contract LM_PC_Staking_v1 is
...
 
         // If trigger address is 0 then its not a user
+        uint earned;
         if (triggerAddress != address(0)) {
-            rewards[triggerAddress] = _earned(triggerAddress, rewardValue);
+            earned = _earned(triggerAddress, rewardValue);
+            rewards[triggerAddress] = earned;
             userRewardValue[triggerAddress] = rewardValue;
         }
+        emit Updated(triggerAddress, newRewardValue, newLastUpdate, earned);
```
FHieser, line 256: Where test?

src/modules/logicModule/LM_PC_Staking_v1.sol
```diff
@@ -262,9 +267,11 @@ contract LM_PC_Staking_v1 is
             return rewardValue;
         }
 
-        return (_getRewardDistributionTimestamp() - lastUpdate) // Get the time difference between the last time it was updated and now (or in case the reward period ended the rewardEnd timestamp)
```
FHieser: I explicitly structured the comments here in a way to explain this complex interaction in a easier way.
You can set brackets here, but please keep it in the original structure

src/modules/logicModule/LM_PC_Staking_v1.sol
```diff
@@ -354,5 +361,6 @@ contract LM_PC_Staking_v1 is
             revert Module__LM_PC_Staking_v1__InvalidStakingToken();
         }
         stakingToken = _token;
+        emit StakingTokenSet(_token);
```
FHieser, line 362: Where test

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -625,6 +625,8 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
 
         // pop the last element of the array
         activePaymentReceivers[client].pop();
+
+        emit PaymentReceiverRemoved(client, paymentReceiver);
```
FHieser, line 629: test

FHieser (changes requested):
See comments

src/modules/base/IModule_v1.sol
fabianschu: No, they are emitted correctly, I'm using this event in the indexer :P

src/external/fees/FeeManager_v1.sol
fabianschu: You're right, this one is redundant, because it is not used in the `init`. I'll remove it

Zitzak:
Given we talked about the Omega audit yesterday, and the need for them to audit additional contracts, do you think some of these code changes should be included? @FHieser

FHieser:
> Given we talked about the Omega audit yesterday, and the need for them to audit additional contracts, do you think some of these code changes should be included? @FHieser

@Zitzak  You mean in the audit itself or that this PR is needed before the launch?
In the first case no, I think the changes here as so small that they dont need to be audited
In the second case, comes down to @fabianschu here how much this is needed from the frontend

Zitzak:
Yeah I meant the audit itself, thank you for answering this. In the case of the code being needed for launch I would say that is a yes, as we need all those events etc for the indexer
@FHieser

FHieser (approved):

marvinkruse:
Personally felt like `CollateralToken` was better then `OrchestratorToken`, does everyone else think so or do all of you like the "new" way of calling it `OrchestratorToken`? I was happy to see that old naming go and be replaced with `CollateralToken` when we did that back then :D @0xNuggan @FHieser @Zitzak  

**Edit**: Resolved, see blow.

marvinkruse (commented):
Besides my comment above, it looks great!

FHieser:
Lol,
when Fabi and I talked about it we joked that this would come up xD
The problem I raised is that Collateral Token does work in the context of issuance, but not in the direction of  general workflow token.
The token has a double role here and I think collateral Token doesnt fit here. I fully understand that OrchestratorToken kinda feels bad, but in my opinion at last it fits better as a "general" naming of the token.
I think we are all open to a better naming, if you have any good idea im happy too if we take that over.
@marvinkruse

marvinkruse:
> Lol, when Fabi and I talked about it we joked that this would come up xD 

Haha, I was fearing this 😂 

> The problem I raised is that Collateral Token does work in the context of issuance, but not in the direction of general workflow token. The token has a double role here and I think collateral Token doesnt fit here. I fully understand that OrchestratorToken kinda feels bad, but in my opinion at last it fits better as a "general" naming of the token. I think we are all open to a better naming, if you have any good idea im happy too if we take that over. @marvinkruse

Makes sense, yeah. The `OrchestratorToken` name is crap, but I can't come up with a better naming right now, so let's just keep it until we have a better way. I'll bring this up again in a different place. It's fine here!

marvinkruse (approved):
looks good, ser!

PR #613, opened 2024-07-17
[SC-700] Remove ERC20Issuance token deployment from FM_BC... contracts

0xNuggan:
## What has been done
-Updated the Bonding Curve contracts to receive an existing ERC20 Issuance token instead of deploying it during initialization.
-Also updated tests and scripts to implement this new flow

*Note*: This PR builds on the changes made in #604 , and as such merges into it (so they can both be merged into dev afterwards).  I've kept them separate for easy review.


marvinkruse (approved):

Zitzak (approved):
LGTM

PR #617, opened 2024-07-19
[SC-623] Bug(38) Users may lose assets by depositing to Rebasing FM

marvinkruse:
Removed the deployment script(s) and notions of the Rebasing FM.

Status: ⌛ `SetupToyOrchestratorScript.s.sol` needs to be updated


marvinkruse:
@0xNuggan 
The `SetupToyOrchestratorScript.s.sol` script is relying on the Rebasing FM, could you change this so it's using a bonding curve version? 🙏

0xNuggan (approved):

PR #619, opened 2024-07-19
[Audit] Hats Fixes - Omega Review

marvinkruse:
This PR will contain the fixes to the audit by [Hats Finance](https://hats.finance), to be reviewed by Team Omega.


src/modules/logicModule/LM_PC_KPIRewarder_v1.sol
PlamenTSV, line 1: Reverting inside the try block is completely fine imo, however if an assertion can be executed, instead of doing:
settleAssertion -> success > revert
why not just let the assertion settle. It's config, data and pending state will get reset anyway and we wouldn't need 2 transactions to execute an assertion that's not actually stuck

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -236,9 +236,9 @@ abstract contract ERC20PaymentClientBase_v1 is
     }
 
     function _ensureValidPaymentOrder(PaymentOrder memory order) private view {
-        _ensureValidRecipient(order.recipient);
-        _ensureValidToken(order.paymentToken);
-        _ensureValidAmount(order.amount);
+        if (!(orchestrator().paymentProcessor().validPaymentOrder(order))) {
```
ben-kaufman, line 235: The problem here is if for example the payment processor is the Simple, then orders are added, but then the payment processor of the Orchestrator is replaced, the time range will still not be validated.

In general, I think it is important to process all payments before changing a payment processor.

src/modules/base/Module_v1.sol
```diff
@@ -296,6 +290,15 @@ abstract contract Module_v1 is
         }
     }
 
+    function _onlyPaymentClientModifier() internal view {
```
ben-kaufman, line 294: This is still quite permissive as any module which returns true for supportsInterface for the PaymentClient inteface will still be able to spend all tokens. But without introducing a system for limiting use of funds of each module there's not much more to do really

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -128,7 +128,8 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
         virtual
         returns (uint mintAmount)
     {
-        _validateDepositAmount(_depositAmount);
```
ben-kaufman, line 131: One thing to ensure amount really never is 0 would be to check also in _calculateNetAndSplitFees that the total fees are not greater than or equal to BPS (it currently allows it to be equal which will result in amount 0)

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -342,8 +345,23 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
         if ((_protocolFee + _workflowFee) > BPS) {
             revert Module__BondingCurveBase__FeeAmountToHigh();
         }
-        protocolFeeAmount = (_totalAmount * _protocolFee) / BPS;
-        workflowFeeAmount = (_totalAmount * _workflowFee) / BPS;
+        // Calculate protocol fee amount if applicable
+        if (_protocolFee > 0) {
+            protocolFeeAmount = _totalAmount * _protocolFee / BPS;
+            // Revert if calculated protocol fee amount rounded down to zero
+            if (protocolFeeAmount == 0) {
```
ben-kaufman, line 353: Well this just checks the fee will not be absolute 0, but the potential rounding remains, so essentially this means the issue of "discounted" minting remains. To ensure there was no rounding error you'd need to either:
1. Enforce that protocolFeeAmount * BPS == _totalAmount * _protocolFee. This is the most accurate but could lead to certain annoyance in making specific purchase amounts.
2. Add a minimum amount to buy so that the rounding error will have to be insignificant.

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -530,6 +532,13 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
                 Module__FM_BC_Bancor_Redeeming_VirtualSupply__CurveInteractionsMustBeClosed(
             );
         }
+        // Check if virtual supply is big enough to ensure compatibility with relative issuance
+        // token decimal and conversion to 18 decimals done in FM_BC_Tools._convertAmountToRequiredDecimal()
+        // so it will not result in a round down 0 value
+        if (_virtualSupply < 10 ** (issuanceTokenDecimals - 18)) {
```
ben-kaufman: There is no guarantee the issuanceTokenDecimals >= 18, the check when setting it only assures it's >= 7. So this will revert if the decimals are less than 18 and there's no guarantee they aren't

src/modules/logicModule/LM_PC_KPIRewarder_v1.sol
```diff
@@ -284,6 +284,35 @@ contract LM_PC_KPIRewarder_v1 is
...
+        uint assertionExpirationTime =
+            oo.getAssertion(assertionId).expirationTime;
+
+        if (block.timestamp <= assertionExpirationTime) {
+            revert Module__LM_PC_KPIRewarder_v1__AssertionNotStuck(assertionId);
+        }
+
+        try oo.settleAssertion(assertionId) {
+            // If the assertion can be settled, it doesn't qualify as stuck and we revert
+            revert Module__LM_PC_KPIRewarder_v1__AssertionNotStuck(assertionId);
```
ben-kaufman, line 307: I think it'd be better to just let it succeed here and settle the assertion instead of reverting

src/modules/logicModule/abstracts/oracleIntegrations/UMA_OptimisticOracleV3/OptimisticOracleIntegrator.sol
```diff
@@ -185,16 +185,40 @@ abstract contract OptimisticOracleIntegrator is
...
+            if (defaultCurrency.balanceOf(address(this)) < defaultBond) {
+                revert
+                    Module__OptimisticOracleIntegrator_InsufficientFundsToPayForBond(
+                );
+            }
+        } else {
+            (bool success, bytes memory data) = address(defaultCurrency).call(
+                abi.encodeWithSelector(
+                    defaultCurrency.transferFrom.selector,
+                    _msgSender(),
```
ben-kaufman: This still transfers from msgSender and not asserter, or you should probably require here the msgSender == asserter

src/modules/base/Module_v1.sol
Zitzak: You are right with your assessment. However as we currently do not have an open system, i.e. modules can only be added to the Module Lib with our approval, the risk managed for now. What you suggest will prob be implemented in the future

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
Zitzak: We had a discussion around this with the team and the decision is to leave it as is. Reason being that a min buy/sell amount currently would just mean moving the goal post, as we do not have a clear idea how high this should be based on what?

Similar your first point, as you state, could result in reverting even with high buy amounts which would not be great for the user experience.

If it turns out that this "discount" will be exploited then we will upgrade accordingly

src/modules/logicModule/LM_PC_KPIRewarder_v1.sol
Zitzak: This comment was included in our session and we decided to leave it as is. Reason being that the function call expects to delete assertions which are stuck. When the assumption of the user calling the function is wrong, ie. it is not stuck, then the user would be inform by this through the revert.

If we would let it success then we might do something that is unintended, and not part of the users aim in calling the function to delete stuck assertion

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
Zitzak: When addressing the issue it became clear that changing this, i.e. making sure the check would be `totalFee < BPS` instead of `<=`, would only effect the read functions as we check that the output can't be 0 elsewhere already. 
1. We first check that `minAmountOut != 0`. [Link](https://github.com/InverterNetwork/contracts/blob/dev/src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol#L230)
2. We check that the actual amount out isn't lower than the `minAmountOut`. [Link](https://github.com/InverterNetwork/contracts/blob/dev/src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol#L282)
Same goes for the sell function

So in light of this we decided not to address this issue

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
Zitzak: We will address this by updating/rephrasing the disclaimer in the Natspec at the associated function.

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
marvinkruse: We fixed it in [this commit](https://github.com/InverterNetwork/contracts/pull/619/commits/74708bf97a7b575cd07eb9369dafd61ef68b3db7).

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
marvinkruse: We fixed it in [this commit](https://github.com/InverterNetwork/contracts/pull/619/commits/74708bf97a7b575cd07eb9369dafd61ef68b3db7).

src/modules/logicModule/abstracts/oracleIntegrations/UMA_OptimisticOracleV3/OptimisticOracleIntegrator.sol
marvinkruse: We fixes this in [this commit](https://github.com/InverterNetwork/contracts/pull/619/commits/c4acd519579caf8b742b423b587333c66491b103).

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
ben-kaufman: Ok, I don't see any issue to leave that, but then I think it'll jjust be more accurate to make it < instead of <= since it is never meant to be equal and will be better to fail early.

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
ben-kaufman: Ok, so just to clarify the issue as submitted on Hats currently was not resolved. I personally think adding a minimum amount set on initialize would make sense. In any case the fix that was added does not really solve the issue, so you could also revert that change.

src/modules/logicModule/LM_PC_KPIRewarder_v1.sol
ben-kaufman: Ok. Just to mention in general a malicious token could still make the proposal stuck, but I think the oracle has a whitelist for currencies so it's ok.

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -541,13 +541,26 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
         internal
         override(VirtualIssuanceSupplyBase_v1)
     {
+        uint decimalDifference;
+        uint decimals = issuanceTokenDecimals;
+
+        if (decimals <= 18) {
+            // If issuance token has 18 or fewer decimals. Ensure that decimal difference is at least 1
+            // to end up with minRequiredSupply of 10
+            decimalDifference = 18 - decimals == 0 ? 1 : 18 - decimals;
```
ben-kaufman, line 550: Why 18 - decimals? If it is less than 18 then it can always be just 0 no (so the condition will be _virtualSupply < 1)

marvinkruse: Not sure whether you are asking why we go with `18` specifically (which is what the BancorFormula calculates with) or why we calculate `18 - decimals`, which is just to get the difference between the issuance tokens decimals and 18 decimals.

Also not sure what you mean exactly with "_it can always be just 0_". To arrive at the condition of `_virtualSupply < 1`, the `decimalDifference` would need to be `0`. I can not see any way for the `decimalDifference` to be `0` though, as we covered exactly that. The minimum would be `1`, even if the `decimals` are `18`, which leads to a condition of `_virtualSupply < 10`, which is exactly what we want.

Could elaborate or clarify this @ben-kaufman? 🙏

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
marvinkruse: Makes sense, @Zitzak addresses this [here](https://github.com/InverterNetwork/contracts/pull/619/commits/d4c37d1932543f6e36fc2c8609ace6a240355a03).

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
Zitzak: We discussed it and we are moving forward with what we have, while monitoring if abuse is being made of the "discount".

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
ben-kaufman: Sorry, I think my last comment was incorrect. Looks fine to me.

Zitzak: We did some back and forth around this, after which we conclude that you are right. Code has been adapted accordingly and the commit can be found here
https://github.com/InverterNetwork/contracts/pull/644/commits/d7849b3bcd7d9b90253ddbce701e57f96a42020a

@ben-kaufman

ben-kaufman: Looks good. I think it makes more sense this way.

PR #621, opened 2024-07-22
refactor: reposition files for clarity

Zitzak:
## Reasoning
1. Common folder only had one Lib contract in it. However the lib was only used within a module so it makes sense to move the lib to the dedicated lib folder in the module folder
2. The EXT Role based authorizer interface was already in the `role` folder so moving the implementation to the same folder makes sense. Also it is a role based authorizer so no need to have it's own folder


src/modules/authorizer/role/AUT_EXT_VotingRoles_v1.sol
FHieser, line 1: The reason we had the VotingRoles contract in a different folder was to differentiate it from the IAuthorizer based contracts that can be used as real Authorizers in a workflow.
You are bascially reverting that conscious decision here.
I think the differentiation is still relevant and correct.
Should we maybe work something out that satisfy your feeling that there is something amiss here?

FHieser (changes requested):
I have maybe something for this PR that you wont like:
Move the Mocks folder somewhere more prominent location.
I hate that it is located in the utils folder. xD

Zitzak:
This PR is replaced by:
https://github.com/InverterNetwork/contracts/pull/651

PR #624, opened 2024-07-24
Sc 294 chore complete natspec

FHieser:
Dont know what happended here on the merge.
Redid it here #625


FHieser:
See starting comments

PR #625, opened 2024-07-24
[SC-294] Chore: Complete NatSpec

FHieser:
Added natspec to different functions and cleaned up a bit


src/external/forwarder/interfaces/ITransactionForwarder_v1.sol
```diff
@@ -1,27 +1,48 @@
...
+
 interface ITransactionForwarder_v1 {
     //--------------------------------------------------------------------------
     // Structs
 
     struct SingleCall {
-        address target; // Target contract that will receive the call
-        bool allowFailure; // Is the call allowed to fail in the multicall execution
-        bytes callData; // Data of the call
+        /// @notice Target contract that will receive the call
```
fabianschu, line 12: I'm not sure if this conforms to the NatSpec standard and is being understood by the compiler, jfyi. The docs are a bit poor on this:
https://docs.soliditylang.org/en/latest/natspec-format.html

Additional discussion on this here: 
https://github.com/ethereum/solidity/issues/14217#issuecomment-1742205238

fabianschu (approved):

src/external/forwarder/interfaces/ITransactionForwarder_v1.sol
FHieser: Uff that feels bad that this isnt supported in this way.
Is this something I should change or is it okay to keep it for now?

Zitzak: Asked @fabianschu in slack. This comment can be considered closed. I will go ahead and merge the PR, as I need the changes for the docs

PR #626, opened 2024-07-25
[Audit] [SC-713] Feat: Add DepositVault

FHieser:


src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
```diff
@@ -0,0 +1,130 @@
...
+    IERC20Metadata
+} from "@oz/token/ERC20/extensions/IERC20Metadata.sol";
+
+// External Dependencies
+import {ERC165Upgradeable} from
+    "@oz-up/utils/introspection/ERC165Upgradeable.sol";
+
+// External Libraries
+import {SafeERC20} from "@oz/token/ERC20/utils/SafeERC20.sol";
+import {Strings} from "@oz/utils/Strings.sol";
```
Zitzak: Not being used

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * @title   Deposit Vault Funding Manager
+ *
+ * @notice  This contract allows users to deposit tokens that fund the workflow.
```
Zitzak:
```suggestion
 * @notice  This contract allows users to deposit tokens to fund the workflow.
```

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
```diff
@@ -0,0 +1,130 @@
...
+        _;
+    }
+
+    //--------------------------------------------------------------------------
+    // Storage
+
+    /// @dev The token that is deposited.
+    IERC20 private _token;
+
+    // Storage gap for future upgrades
```
Zitzak:
```suggestion
    /// @dev Storage gap for future upgrades
```

Otherwise this won't be picked up by the automatic doc generator from Forge. This hasn't been done in the other contracts yet.

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
```diff
@@ -0,0 +1,130 @@
...
+
+// Internal Dependencies
+import {Module_v1} from "src/modules/base/Module_v1.sol";
+
+import {Initializable} from "@oz-up/proxy/utils/Initializable.sol";
+
+// External Interfaces
+import {
+    IERC20,
+    IERC20Metadata
```
Zitzak: Not being used

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
```diff
@@ -0,0 +1,130 @@
...
+            || super.supportsInterface(interfaceId);
+    }
+
+    using SafeERC20 for IERC20;
+
+    //--------------------------------------------------------------------------
+    // Modifier
+
+    /// @dev Checks if the given Address is valid.
+    modifier validAddress(address to) {
```
Zitzak: Not that important, but we could maybe move this modifier to the Module_v1, as it is being used in both the BondingCurveBase, and prob other modules as well

Zitzak (changes requested):
Few small thinks, nothing major.

We are of course bit short on time so not made a comment of it, but would be great if the test functions can be written as defined by the testing handbook :) So not a requirement as of now, but we should start keeping ourselves to it

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
FHieser: Should I do that here?

Zitzak: Not sure. Just noticed that the same functionality is needed across multiple modules. I'll leave it up to you :)

Zitzak (approved):
LGTM

Zitzak:
@FHieser @0xNuggan I decided to merge the FM already so we can sent a link to the auditors today. Jfyi

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
```diff
@@ -0,0 +1,115 @@
...
+    /// @inheritdoc IFundingManager_v1
+    function token() public view returns (IERC20) {
+        return _token;
+    }
+
+    //--------------------------------------------------------------------------
+    // Public Mutating Functions
+
+    /// @inheritdoc IFM_DepositVault_v1
+    function deposit(uint amount) external {
```
marvinkruse, line 95: To be consistent with our other FundingManagers and what's common in the space, we should have a `depositFor(address, amount)` function here imho.

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
```diff
@@ -0,0 +1,21 @@
...
+    /// @notice Event emitted when a deposit takes place.
+    /// @param _from The address depositing tokens.
+    /// @param _amount The amount of tokens deposited.
+    event Deposit(address indexed _from, uint _amount);
+
+    //--------------------------------------------------------------------------
+    // Functions
+
+    /// @notice Deposits a specified amount of tokens into the contract from the sender's account.
+    /// @dev    Reverts if attempting self-deposits Note: when using the transactionForwarder,
```
marvinkruse: The `Note` should have it's only `@dev` tag according to the best practices of NatSpec, as it is quite important.

src/modules/fundingManager/rebasing/interfaces/IRebasingERC20.sol
```diff
@@ -19,6 +19,33 @@ import "@fm/rebasing/interfaces/IERC20Metadata.sol";
  * @author Buttonwood Foundation
  */
 interface IRebasingERC20 is IERC20Metadata {
+    //--------------------------------------------------------------------------
+    // Errors
+
+    /// @notice Function is only callable by authorized address.
```
marvinkruse: Description of the `error` is wrong.

src/modules/fundingManager/rebasing/interfaces/IRebasingERC20.sol
```diff
@@ -19,6 +19,33 @@ import "@fm/rebasing/interfaces/IERC20Metadata.sol";
...
+    error Module__RebasingERC20__CannotSelfDeposit();
+
+    /// @notice There is a cap on deposits.
```
marvinkruse: Description of the `error` is slightly wrong (_the problem is not that there is a cap, the problem is that it was reached_).

marvinkruse (changes requested):

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
Zitzak: btw, just noticing this function in the FM_Rebasing_v1
https://github.com/InverterNetwork/contracts/blob/dev/src/modules/fundingManager/rebasing/FM_Rebasing_v1.sol#L150

0xNuggan: Do we really need it in this case? The FundingManager does not track balances, so you cannot deposit "for" somebody. In the other cases the depositFor/buyFor takes funds from the sender and increases the balance of somebody else, but here the implementation of a buyFor function would look exactly the same...

0xNuggan:
took the liberty to update the comments following @marvinkruse  review.

0xNuggan (approved):

Zitzak:
@0xNuggan Please don't do more :D was just about to push everything

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
marvinkruse: > The FundingManager does not track balances, so you cannot deposit "for" somebody.

I see your point, but to me this doesn't make a big difference. A `depositFor` should still exist, as we do not know whether someone might need it (there are instances where multisigs just give an approval and a specific member can handle the deposit up to the allowance limit whenever they see fit from their own wallet), so I'd still add it just in case, personally.

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
```diff
@@ -0,0 +1,21 @@
...
+    /// @param _from The address depositing tokens.
+    /// @param _amount The amount of tokens deposited.
+    event Deposit(address indexed _from, uint _amount);
+
+    //--------------------------------------------------------------------------
+    // Functions
+
+    /// @notice Deposits a specified amount of tokens into the contract from the sender's account.
+    /// @dev    Reverts if attempting self-deposits
+    /// @dev When using the transactionForwarder, validate transaction success to prevent nonce exploitation and ensure transaction integrity.
```
marvinkruse: Formatting is off now, after the change.

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
marvinkruse: Closing this, as it's fixed overall, just the formatting is now off, but I added a new comment for that.

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
Zitzak: I'll fix it. Pushing in a min, after removing my comment edits

marvinkruse (approved):

Zitzak:
This PR has been send to Omega. Please do not merge until after their review jfyi.
@FHieser @marvinkruse @0xNuggan

marvinkruse:
(I will rebase this one)

marvinkruse:
@Zitzak @0xNuggan You reviewed it already - but I now rebased this PR after the extensive `dev` changes, only real difference here is now that I activated the prepared lines (that Felix already added) in the new deployment scripts and made sure it all runs fine. If you could check that out briefly it would be great 🙏

marvinkruse (approved):

Zitzak (approved):
LGTM

0xNuggan (approved):
LGTM

marvinkruse:
Note: This PR will now be merged. We will still monitor this PR as it will be used for a review by Omega. If there should be required fixes, a new PR will be opened up towards `dev`.

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    //--------------------------------------------------------------------------
+    // Internal Functions
+
+    /// @dev Internal function to deposit tokens.
+    /// @param from The address to deposit tokens for.
+    /// @param amount The number of tokens to deposit.
+    function _deposit(address from, uint amount) internal {
+        token().safeTransferFrom(from, address(this), amount);
```
ben-kaufman, line 112: So if I approved 20 tokens to this contract, but only wanted to deposit 10, anyone can call depositFor and make it deposit the other 10 tokens as well. This seems quite dangerous. A normal depositFor behaviour would be that the tokens are taken from the msgSender, but are credited to the from address.

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
```diff
@@ -0,0 +1,31 @@
...
+    /// @notice Deposits a specified amount of tokens into the contract from the sender's account.
+    /// @dev    Reverts if attempting self-deposits
+    /// @dev    When using the transactionForwarder, validate transaction success to prevent nonce
+    ///         exploitation and ensure transaction integrity.
+    /// @param amount The number of tokens to deposit.
+    function deposit(uint amount) external;
+
+    /// @notice Deposits a specified amount of tokens into the contract for the address defined in
+    ///         the `from` parameter.
+    /// @dev    Reverts if attempting self-deposits
```
ben-kaufman, line 25: The implementation does not revert on self deposit

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
```diff
@@ -0,0 +1,31 @@
...
+    /// @notice Event emitted when a deposit takes place.
+    /// @param _from The address depositing tokens.
+    /// @param _amount The amount of tokens deposited.
+    event Deposit(address indexed _from, uint _amount);
+
+    //--------------------------------------------------------------------------
+    // Functions
+
+    /// @notice Deposits a specified amount of tokens into the contract from the sender's account.
+    /// @dev    Reverts if attempting self-deposits
```
ben-kaufman, line 17: The implementation does not revert on self deposit

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
```diff
@@ -0,0 +1,130 @@
...
+
+    /// @inheritdoc IFM_DepositVault_v1
+    function deposit(uint amount) external {
+        address from = _msgSender();
+        _deposit(from, amount);
+    }
+
+    /// @inheritdoc IFM_DepositVault_v1
+    function depositFor(address from, uint amount) external {
+        _deposit(from, amount);
```
ben-kaufman, line 102: Missing a check that the from address is not address(this)

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    //--------------------------------------------------------------------------
+    // Internal Functions
+
+    /// @dev Internal function to deposit tokens.
+    /// @param from The address to deposit tokens for.
+    /// @param amount The number of tokens to deposit.
+    function _deposit(address from, uint amount) internal {
```
ben-kaufman, line 111: Should probably disallow also 0 amount here

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
ben-kaufman: I'd say what is implemented here is more "depositFrom" than the usual "depositFor"

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
```diff
@@ -0,0 +1,130 @@
...
+import {IOrchestrator_v1} from
+    "src/orchestrator/interfaces/IOrchestrator_v1.sol";
+import {IFundingManager_v1} from "@fm/IFundingManager_v1.sol";
+import {IFM_DepositVault_v1} from
+    "@fm/depositVault/interfaces/IFM_DepositVault_v1.sol";
+
+// Internal Dependencies
+import {Module_v1} from "src/modules/base/Module_v1.sol";
+
+import {Initializable} from "@oz-up/proxy/utils/Initializable.sol";
```
ben-kaufman, line 14: Unused import

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.0;
```
ben-kaufman, line 2: Why is the pragma in some contracts ^0.8.0 but in others 0.8.23

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
```diff
@@ -0,0 +1,31 @@
...
+
+interface IFM_DepositVault_v1 {
+    //--------------------------------------------------------------------------
+    // Events
+
+    /// @notice Event emitted when a deposit takes place.
+    /// @param _from The address depositing tokens.
+    /// @param _amount The amount of tokens deposited.
+    event Deposit(address indexed _from, uint _amount);
```
ben-kaufman, line 11: You might want to add here address indexed _to, to allow an account to deposit into another's "credit"

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
marvinkruse: Valid, the `@dev`-comment is a copy-paste-error from `FM_Rebasing_v1`, as self-deposits would be bad in that contract. 
Here, I'd argue that a `depositFor` with `from` being set to the address of the contract itself (`address(this)`) would fail anyway via the approval check (throws `ERC20InsufficientAllowance`), so there is no way for self-deposits to happen. So I'd just remove this comment for the `deposit` and `depositFor` functions, especially that - even if a self-deposit would happen (i.e. if for some reason the vault has set an approval for _itself_ - it wouldn't have any effect besides a confusing event emission.

Would you agree with logic @ben-kaufman?

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
marvinkruse: Discussed above, see [here](https://github.com/InverterNetwork/contracts/pull/626#discussion_r1707911078).

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
ben-kaufman: Yes it should fail on approve check (unless a "special" token implementations don't check approve if from is msg.sender). I think you can either just remove the comment, or require the from address is not address(this).

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
marvinkruse: Also discussed above ([here](https://github.com/InverterNetwork/contracts/pull/626#discussion_r1707911078)). I'd argue that this would just be a waste of gas to check, as a self-deposit can only happen if the contract has for some reason self-approved before, and even then it's just a weird `Deposit` being emitted with a false value, no funds are moved. 

Do you think the check should really be added @ben-kaufman?

ben-kaufman: I mostly commented this because it appeared as a requirement in the documentation, but I don't think it's so necessary

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
marvinkruse: Yeah, fair. Not a big fan of wasting gas for every user just to ensure that a potential 0 depositing user doesn't pay gas/send their tx if they forgot to put something in the `amount` field, but I believe we do it in other contracts as well, so I guess that's a fair point.

As a 0-deposit has literally zero impact here though and should just be filtered on the UI/SDK level, I'd personally lean towards not changing anything, unless @Zitzak or @0xNuggan think it makes sense.

Or did I miss any security-related implication you had in mind @ben-kaufman?

ben-kaufman: No that's mostly a matter of prefence, shouldn't matter much

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
marvinkruse: Actual contracts are all `0.8.23` (or should be), while interfaces are all `^0.8.0`, so other projects can import our interfaces easier without needing multiple compiler versions potentially. We just following the example of other projects here.

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
marvinkruse: Valid, fixed in [this commit](https://github.com/InverterNetwork/contracts/pull/646/commits/8eb10a27e4b749a98bd0f2dde49414cd32244193). PR that addresses all of the fixes is #646. 🙏

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
marvinkruse: Very valid point, this is not the traditional `depositFor` where a user would get receipt tokens, withdraw rights, or anything else in return. It's a bit confusing. I'd probably just rename it then to `depositFrom`, but...

I'd personally just remove this function. Looking at it now, I don't see any reason to have it. If someone can't (for whatever reason) call the funding manager directly to deposit (weird wallet that can't call arbitrary contracts, or whatever), they can always just transfer the tokens directly into the DepositVault anyway via a regular token transfer, as it doesn't matter who sent the tokens and no accounting happening.

I've removed it in [this commit](https://github.com/InverterNetwork/contracts/pull/646/commits/ef74cf6e825dc685712d95b1f9e3614da9c0eb20). The PR handling all fixes is #646. Let me know what you think @ben-kaufman - I'll also give @Zitzak and @0xNuggan the chance to comment here as well, maybe I missed something, in which case we can still revert my removal of course.

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
marvinkruse: Now that I removed `depositFor`, this is not applicable anymore, right?
I've removed it in [this commit](https://github.com/InverterNetwork/contracts/pull/646/commits/ef74cf6e825dc685712d95b1f9e3614da9c0eb20). The PR handling all fixes is #646.

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
marvinkruse: Okay, makes sense, thank you 🙏 I removed the comment itself in [this commit](https://github.com/InverterNetwork/contracts/pull/646/commits/a31b5938349b9456298d07d30f7f8b900179ea23), so we won't need/add a check itself here. The PR handling all fixes is #646.

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
marvinkruse: Verified that it fails on the approve check, and that there is no impact besides the event even if that check wouldn't make it fail. I removed the comment itself in [this commit](https://github.com/InverterNetwork/contracts/pull/646/commits/a31b5938349b9456298d07d30f7f8b900179ea23), so we won't need/add a check itself here. The PR handling all fixes is #646.

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
marvinkruse: Comment itself removed from the `deposit` function, `depositFor` is removed altogether, so this one should be done as well.

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
marvinkruse: Okay, thank you 🙏 I personally wouldn't change anything then, but I'll wait for @Zitzak or @0xNuggan to chime in briefly. 👍

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
ben-kaufman: Yes I think it makes sense to remove it

PR #620, opened 2024-07-22
[SC-710] Implement audit suggestions for ERC20issuance

0xNuggan:
**Draft for now until their review is final**

## What has been done
- Updated contract description text
- Made decimals variable `immutable`
- Refactored imports and removed unused SafeERC20 import

PR #628, opened 2024-07-28
Fix: Remove unused import in Issuance Token

marvinkruse:
Within the Omega audit of the Issuance Token, they raised that `SafeERC20` and `Context` were unnecessary imports as they are not used.  The original PR #620 did not address those fully, so I just remove the last one here.


Zitzak (approved):
LGTM

PR #629, opened 2024-07-29
[SC-705] Chore: correct spelling

Zitzak:


src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -131,13 +131,10 @@ contract AUT_Roles_v1 is
         _setRoleAdmin(DEFAULT_ADMIN_ROLE, ORCHESTRATOR_OWNER_ROLE);
 
         // Set up MANAGER role structure:
```
marvinkruse: `_setRoleAdmin` doesn't have a specific line documenting its behavior anymore, as the like `Set up MANAGER... ` is the overall heading of what's happening in all of these lines. Was this intentional? I'd have left that line in.

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -131,13 +131,10 @@ contract AUT_Roles_v1 is
...
-        // -> set OWNER as admin of DEFAULT_ADMIN_ROLE
         _setRoleAdmin(ORCHESTRATOR_MANAGER_ROLE, ORCHESTRATOR_OWNER_ROLE);
         // grant MANAGER Role to specified address
         _grantRole(ORCHESTRATOR_MANAGER_ROLE, initialManager);
-
-        // If there is no initial owner specfied or the initial owner is the same as the deployer
-
+        // grant Owner Role to initial owner address
```
marvinkruse: Above, the role names are capitalized (`MANAGER`, `OWNER`), here the first `Owner` is not

marvinkruse (changes requested):
Otherwise good :)

src/modules/authorizer/role/AUT_Roles_v1.sol
Zitzak: Followed Omega's advice. However see your point in this

marvinkruse (approved):
thank you, lgtm!

PR #630, opened 2024-07-30
Feat: PIM workflow factory [SC-718]

fabianschu:


src/external/interfaces/IOwnable.sol
```diff
@@ -0,0 +1,22 @@
+/**
```
marvinkruse: Without having looking at this too extensively, is this because OZ doesn't offer the `IOwnable`?

src/external/token/IERC20Issuance_v1.sol
```diff
@@ -31,4 +33,13 @@ interface IERC20Issuance_v1 is IERC20 {
     /// @param _from The address of the owner.
     /// @param _amount The amount of tokens to burn.
     function burn(address _from, uint _amount) external;
+
+    // Read
+
+    /// @notice Mapping of allowed minters
```
marvinkruse, line 39: Is missing the `@return` tag here I believe.

src/factories/workflow-specific/PIM_WorkflowFactory.sol
```diff
@@ -0,0 +1,187 @@
...
+// External Interfaces
+import {IERC20} from "@oz/token/ERC20/IERC20.sol";
+
+// External Implementations
+import {ERC20Issuance_v1} from "src/external/token/ERC20Issuance_v1.sol";
+
+// External Dependencies
+import {Ownable} from "@oz/access/Ownable.sol";
+
+contract PIM_WorkflowFactory is Ownable, IPIM_WorkflowFactory {
```
marvinkruse: More of a general question, also to @0xNuggan @FHieser and @0xNuggan - do we want this one to be called like it is or append a `v1` here as well?

src/factories/workflow-specific/PIM_WorkflowFactory.sol
```diff
@@ -0,0 +1,187 @@
...
+    "src/factories/interfaces/IPIM_WorkflowFactory.sol";
+
+// External Interfaces
+import {IERC20} from "@oz/token/ERC20/IERC20.sol";
+
+// External Implementations
+import {ERC20Issuance_v1} from "src/external/token/ERC20Issuance_v1.sol";
+
+// External Dependencies
+import {Ownable} from "@oz/access/Ownable.sol";
```
marvinkruse: Is there a specific reason why you opted to go with `Ownable` instead of the `Ownable2Step` variant (which we used usually)?

src/factories/workflow-specific/PIM_WorkflowFactory.sol
```diff
@@ -0,0 +1,187 @@
...
+
+    /// @inheritdoc IPIM_WorkflowFactory
+    function withdrawFee(IERC20 token, address to) external onlyOwner {
+        token.transfer(to, token.balanceOf(address(this)));
+    }
+
+    //--------------------------------------------------------------------------
+    // Internal Functions
+
+    function _manageInitialCollateral(
```
marvinkruse: Do we want or need to ensure that `initialCollateralSupply` is `>0` here? It looks to me like this could be super low or zero otherwise, which I'm not sure we want, but might be a non-technical decision as well.

src/factories/workflow-specific/PIM_WorkflowFactory.sol
marvinkruse: Second general question: Do we want or need MetaTx support here? Not sure how beneficial it is, as the user needs to approve the collateral token anyways I'd assume, so it's not a fully gas-less one, unless the collateral token has a `permit` function.

src/factories/workflow-specific/PIM_WorkflowFactory.sol
```diff
@@ -0,0 +1,187 @@
...
+            IERC20(collateralToken).transferFrom(
+                msg.sender, address(this), feeAmount
+            );
+        }
+    }
+
+    function _transferTokenOwnership(
+        ERC20Issuance_v1 issuanceToken,
+        address newAdmin
+    ) private {
```
marvinkruse: Did you intentionally make this one `private` while `_manageInitialCollateral` is `internal`?

src/factories/workflow-specific/PIM_WorkflowFactory.sol
```diff
@@ -0,0 +1,187 @@
...
+            issuanceToken.renounceOwnership();
+        } else {
+            issuanceToken.transferOwnership(newAdmin);
+        }
+    }
+
+    function _transferWorkflowAdminRights(
+        IOrchestrator_v1 orchestrator,
+        address newAdmin
+    ) private {
```
marvinkruse: Same question as for `_transferTokenOwnership` with `private`.

marvinkruse (dismissed):
Good work on this, looks great! Just had some questions :)

src/factories/workflow-specific/PIM_WorkflowFactory.sol
fabianschu: Should have been private, will change. Ty

src/factories/workflow-specific/PIM_WorkflowFactory.sol
fabianschu: I took the issuance token as inspo. I'll change it to 2step

src/factories/workflow-specific/PIM_WorkflowFactory.sol
fabianschu: Good question, I just added support so that we have it. If anyone thinks this is a bad idea, lmk

src/factories/workflow-specific/PIM_WorkflowFactory.sol
fabianschu: Good question, @0xNuggan could you help out here?

src/factories/workflow-specific/PIM_WorkflowFactory.sol
fabianschu: I added a v1, lmk if this is not desired.

src/external/interfaces/IOwnable.sol
0xNuggan: I'm sorry, I don't understand x) What is the purpose of this file? It doesn't seem to be used anywhere.

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
```diff
@@ -0,0 +1,228 @@
...
+        if (PIMConfig.isRenouncedIssuanceToken) {
+            _transferTokenOwnership(issuanceToken, address(0));
+        } else {
+            _transferTokenOwnership(
+                issuanceToken, PIMConfig.issuanceTokenParams.initialAdmin
+            );
+        }
+
+        // if renounced workflow flag is set, renounce admin rights over workflow, else transfer admin rights to initial admin
+        if (PIMConfig.isRenouncedWorkflow) {
```
0xNuggan, line 144: I think we'de need some kind of check here that if the deployer renounces workflow ownership, the project fees have to be set to 0. If not, we'd end up with stuck funds.  Alternatively we could do the thing we mentioned of giving admin role to the factory, and storing a mapping(deployer => worklow) that routes the withdrawProjectCollateralFee() function. (Although what happens with lost private keys in that case? maybe easier to just set fees to 0)

src/factories/workflow-specific/PIM_WorkflowFactory.sol
fabianschu: @0xNuggan ping

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
fabianschu: We have decided in the last tech sync that we go with the approach where fees can be claimed via the factory. This is currently being implemented in a follow-up PR (draft: https://github.com/InverterNetwork/contracts/pull/631).

Lost keys is a "pech gehabt" scenario imo. This is the same for lost admin keys in a regular workflow, no?

src/factories/workflow-specific/PIM_WorkflowFactory.sol
fabianschu: In my personal opinion I wouldn't impose restrictions through input validations here, because it seems to me that we are still exploring the behaviour of the curve and therefore can use the flexibility.

EDIT: there is already a "non-zero" validation in the bonding curve itself

0xNuggan: supply cannot be 0, but it can be very small.

0xNuggan: **IF** we were to add safety configuration validation at this level, it shuld be that the initialStatic price if the chosen colateral/supply is bigger than 0

0xNuggan: But it's up to debate if we want to do it or enforce it through the frontend/simulation

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
0xNuggan: Alright, I see! Then all good

0xNuggan (approved):
My only open question is about the need of the IOwnable interface. Apart form that, everything looks good to me!

src/factories/workflow-specific/PIM_WorkflowFactory.sol
fabianschu: If initialStaticPrice < 0 breaks the underlying math, I think we should add validation. I won't reimplement/generalize some `pure` version of `getStaticPriceForBuying`, but just add an if/revert after the curve has been deployed.

Edit: since you pointed me at the fact that getStaticPrice just underflows I'll have to reimplement the calculations for a validation.

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
```diff
@@ -0,0 +1,291 @@
...
+        // get bonding curve / funding manager
+        address fundingManager = address(orchestrator.fundingManager());
+
+        // make sure that curve configuration doesn't allow a negative static price since would break the math
+        if (IBondingCurveBase_v1(fundingManager).getStaticPriceForBuying() < 0)
+        {
+            revert
+                IPIM_WorkflowFactory_v1
+                .PIM_WorkflowFactory__InvalidConfiguration();
+        }
```
fabianschu: Here is the "input validation"

src/factories/workflow-specific/PIM_WorkflowFactory.sol
fabianschu: @0xNuggan Any chance that you have a config at hand that would break the invariant to write a test?

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
0xNuggan: we need to enforce (staticPrice > 0). The function returns a uint, so the value wouldn't ever be negative anyway

0xNuggan (changes requested):

test/factories/workflow-specific/PIM_WorkflowFactory_v1.t.sol
```diff
@@ -0,0 +1,516 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
0xNuggan, line 1: Also, without wanting to be pedantic. it would be great if you could add test trees to this file for clearer structuring. There's a great VSCode extension for this: https://marketplace.visualstudio.com/items?itemName=aprilandjan.ascii-tree-generator

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
fabianschu: Dang, I was assuming that it returns `int` but didn't check. In this case I will have to reimplement/generalize a `pure` version of getStaticPrice after all..

0xNuggan:
Copying another coment over sine it has been stuck in "pending hell":
Without wanting to be pedantic. it would be great if you could add test trees to this file for clearer structuring. There's a great VSCode extension for this: https://marketplace.visualstudio.com/items?itemName=aprilandjan.ascii-tree-generator

src/factories/workflow-specific/PIM_WorkflowFactory.sol
0xNuggan: INITIAL_SUPPLY = 62131676819845007200000;
INITIAL_RESERVE = 12413909028605033;
RESERVE_RATIO = 199_800;

-> This would give a spot price of 1. Big sells would push the spot price to 0, breaking the invariant (sell(x) => y && buy(y) => x)

fabianschu:
I am not sure how exactly I can implement a robust validation. I checked out the config that you gave me 
```
INITIAL_SUPPLY = 62131676819845007200000;
INITIAL_RESERVE = 12413909028605033;
RESERVE_RATIO = 199_800;
```
and as you said for this config `getStaticPrice` returns `1`. This is because within `getStaticPrice` the nominator is only slightly bigger than the denominator and therefore it rounds to one. Ofc I could add a validation to make sure that `getStaticPrice() > 1`, but I don't properly understand what I'm doing here. 

You're saying it is not the initial config that leads to `getStaticPrice` becoming negative, but the initial config _in combination with_ a "big sale". 

Purely based on intuition (and not based on an understanding of the math) I assume that the closer `getStaticPrice` is to zero, the higher the chance that the assumptions that the math implementation requires can be broken (= invariant doesn't hold). If that is correct my question is how close to zero do we want the user to be able to set the initial spot price with their initial config? Is `spotPrice > 1` sufficient, or `spotPrice > 2`, `spotPrice > 10`...?

fabianschu:
Based on my discussions w/ @0xNuggan this PR contains some new changes:
1. new flag to indicate whether the curve should be initialized with real liquidity (even if miniscule amount). If `false`, initialIssuanceSupply amount is burnt and deployer doesn't send initialCollateralSupply to curve. This way the deplyoer cant get obscene amounts of issuance tokens for very cheap.
2. creation fee mechanism has been removed since it doesn't make sense together with 1.; this also gives more flexibility for the use cases that the factory can be used for (client work)
3. new feature: deployer can specify an initial purchase from the curve in terms of collateral token spent, so that they have preferential access to the first purchase from the curve (figuratively speaking this purchase "starts" at the point on the curve given by the initial virtual supplies)

Still todo:
- [ ] write fuzz test that checks that if user immediately sells and buys back that they receive the end up with the same balance (to minimize chance of imprecision errors)

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
```diff
@@ -0,0 +1,289 @@
...
+    ) private {
+        if (withInitialLiquidity) {
+            // collateral token is paid for by the msg.sender
+            collateralToken.transferFrom(
+                _msgSender(), address(fundingManager), initialCollateralSupply
+            );
+            // issuance token is minted to the the specified recipient
+            issuanceToken.mint(recipient, initialIssuanceSupply);
+        } else {
+            // issuance token is minted to the the specified recipient
```
0xNuggan: comment copied over

test/utils/helpers/EventHelpers.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
0xNuggan: It looks like we aren't using this anywhere?

test/factories/workflow-specific/PIM_WorkflowFactory_v1.t.sol
```diff
@@ -0,0 +1,565 @@
...
+            .calculatePurchaseReturn(collateralAmountOut);
+        // now use the collateral from the sale to buy back
+        IBondingCurveBase_v1(fundingManager).buy(
+            collateralAmountOut, issuanceAmountOut
+        );
+        uint staticPriceAfter = IBondingCurveBase_v1(fundingManager).getStaticPriceForBuying();
+        // CHECK: if recipient SELLS complete stack and BUYS BACK they end up with same balance of issuanceToken
+        assertEq(
+            balanceBefore, issuanceToken.balanceOf(pimConfig.recipient)
+        );
```
fabianschu: This assertion fails because there is a slight imprecision:
image

test/factories/workflow-specific/PIM_WorkflowFactory_v1.t.sol
```diff
@@ -0,0 +1,565 @@
...
```
fabianschu: Same here: 
image

fabianschu:
I found a slight imprecision that occurs if when the recipient of the initial purchase immediately sells and buys back the initial purchase. It is so small that the static price doesn't seem to be affected. I'm not sure what to do about it. Is that a surprise? Is that expected? If it's a surprise, is it problematic? In the given test cases the imprecision seems to be really small (less than a percent per mille), but I'm not quite sure if there are reasonable configurations where the imprecision grows to become problematic. Ideally I would write a fuzz test to check this, but it is hard because I'm having troubles defining the space of "reasonable configurations". Wdyt?




NEW

[10:34](https://inverternetwork.slack.com/archives/D06UAE7BLMS/p1722976469339829)
It is so small that the static price doesn't seem to be affected
10:35
I'm not sure what to do about it. Is that a surprise? Is that expected? If it's a surprise, is it problematic?
10:36
I marked the respective sections in the pr: https://github.com/InverterNetwork/contracts/pull/630
[10:39](https://inverternetwork.slack.com/archives/D06UAE7BLMS/p1722976784961489)
In the given test cases the imprecision seems to be really small (less than a "promille"), but I'm not quite sure if there are reasonable configurations where the imprecision grows to become problematic. Ideally I would write a fuzz test to check this, but it is hard because I'm having troubles defining the space of "reasonable configurations". (edited) 
10:39
What are your thoughts on this?

test/factories/workflow-specific/PIM_WorkflowFactory_v1.t.sol
```diff
@@ -0,0 +1,574 @@
...
+        // get the static price after the re-buy
+        uint staticPriceAfter =
+            IBondingCurveBase_v1(fundingManager).getStaticPriceForBuying();
+        // CHECK: if recipient SELLS complete stack and BUYS BACK they end up with same balance of issuanceToken
+        assertApproxEqRel(
+            balanceBefore,
+            issuanceToken.balanceOf(pimConfig.recipient),
+            0.00001e18
+        );
```
fabianschu, line 234: This fails, with fuzzing we get to deviations of up to 24%. In this case also the static price comparison in the next line starts to fail with significant deviations:

image

The `firstCollateralIn` value (= the collateral token amount used for the first purchase) that leads to this scenario is quite high: `79466106564583438811825437264293701`

0xNuggan (approved):

src/external/token/IERC20Issuance_v1.sol
```diff
@@ -31,4 +33,14 @@ interface IERC20Issuance_v1 is IERC20 {
...
+    /// @param _minter The address of the minter.
+    /// @return If the address is allowed to mint or not.
+    function allowedMinters(address _minter) external view returns (bool);
+
+    // /// @notice Returns contract owner
+    // function owner() external view returns (address);
```
marvinkruse, line 45: Why is there an out-commented function here?

src/factories/interfaces/IPIM_WorkflowFactory_v1.sol
```diff
@@ -0,0 +1,138 @@
...
+
+// Internal Dependencies
+import {ERC20Issuance_v1} from "src/external/token/ERC20Issuance_v1.sol";
+
+// External Interfaces
+import {IERC20} from "@oz/token/ERC20/IERC20.sol";
+
+interface IPIM_WorkflowFactory_v1 {
+    //--------------------------------------------------------------------------
+    // Errors
```
marvinkruse, line 25: Missing a newline after the `// Errors` part.

src/factories/interfaces/IPIM_WorkflowFactory_v1.sol
```diff
@@ -0,0 +1,138 @@
...
+    event PIMWorkflowCreated(
+        address indexed bondingCurve,
+        address indexed issuanceToken,
+        address indexed deployer,
+        address recipient,
+        bool isRenouncedIssuanceToken,
+        bool isRenouncedWorkflow
+    );
+
+    /// @notice Event emitted when factory owner sets new fee.
```
marvinkruse, line 50: Incomplete `notice`, it's not the `fee` that is set but the `fee receipient` (I assume).

src/factories/interfaces/IPIM_WorkflowFactory_v1.sol
```diff
@@ -0,0 +1,138 @@
...
+    /// @param oldRecipient The previous pim fee recipient.
+    /// @param newRecipient The new pim fee recipient.
+    event PimFeeRecipientUpdated(
+        address indexed oldRecipient, address indexed newRecipient
+    );
+
+    /// @notice Event emitted when PIM fee (buy/sell fees) is claimed.
+    /// @param claimer The address of the claimer.
+    /// @param amount The amount claimed.
+    event PimFeeClaimed(address indexed claimer, uint amount);
```
marvinkruse, line 60: General: In our `LM_PC_Bounties_v1` we use "claimant", which is a synonym of "claimer", but I'd prefer to be consistent, so let's use "claimant" here as well imho.

src/factories/interfaces/IPIM_WorkflowFactory_v1.sol
```diff
@@ -0,0 +1,138 @@
...
+        uint firstCollateralIn;
+        bool isRenouncedIssuanceToken;
+        bool isRenouncedWorkflow;
+        bool withInitialLiquidity;
+    }
+
+    //--------------------------------------------------------------------------
+    // Functions
+
+    /// @notice Deploys a workflow with a bonding curve and an issuance token
```
marvinkruse, line 109: Missing a `.` at the end.

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
```diff
@@ -0,0 +1,289 @@
...
+    function withdrawPimFee(address fundingManager, address to)
+        external
+        onlyPimFeeRecipient(fundingManager)
+    {
+        uint amount =
+            IBondingCurveBase_v1(fundingManager).projectCollateralFeeCollected();
+        IBondingCurveBase_v1(fundingManager).withdrawProjectCollateralFee(
+            to, amount
+        );
+        emit IPIM_WorkflowFactory_v1.PimFeeClaimed(_msgSender(), amount);
```
marvinkruse, line 176: More of a question than a comment: should this also contain the `to` address or do you think whoever triggered this is enough to have in the event?

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    /// @inheritdoc IPIM_WorkflowFactory_v1
+    function transferPimFeeEligibility(address fundingManager, address to)
+        external
+        onlyPimFeeRecipient(fundingManager)
+    {
+        _pimFeeRecipients[fundingManager] = to;
+        emit IPIM_WorkflowFactory_v1.PimFeeRecipientUpdated(_msgSender(), to);
```
marvinkruse, line 185: Is it not important to know, for which `fundingManager` the `pimFeeRecipient` was updated?

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
marvinkruse: Also, is it not important to know for much `fundingManager` this was withdrawn?

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
```diff
@@ -0,0 +1,289 @@
...
+        if (withInitialLiquidity) {
+            // collateral token is paid for by the msg.sender
+            collateralToken.transferFrom(
+                _msgSender(), address(fundingManager), initialCollateralSupply
+            );
+            // issuance token is minted to the the specified recipient
+            issuanceToken.mint(recipient, initialIssuanceSupply);
+        } else {
+            // issuance token is minted to the burn address
+            issuanceToken.mint(address(0xDEAD), initialIssuanceSupply);
```
marvinkruse, line 209: More of a question as I just don't immediately understand this: Why is this minted to a non-accessible address? Is this just to get the curve to the right area of pricing, but the tokens shouldn't be used outside of that?

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
```diff
@@ -0,0 +1,289 @@
...
+            issuanceToken.transferOwnership(newAdmin);
+        }
+    }
+
+    function _transferWorkflowAdminRights(
+        IOrchestrator_v1 orchestrator,
+        address newAdmin
+    ) private {
+        bytes32 adminRole = orchestrator.authorizer().getAdminRole();
+        // if renounced flag is set, add zero address as admin (because workflow must have at least one admin set)
```
marvinkruse, line 249: I understand the comment, but I don't see where the zero address is being set here, as the input `newAdmin` can be non zero. Am I missing something? Comment confuses me :D

marvinkruse (commented):
@fabianschu Saw that this was already merged before I could get my second pass in (my bad), but still have some comments after checking now.

fabianschu:
I will create a new PR to address your valid points. Sorry for the quick merge, the rationale here was to unblock SDK development 🙏🏼

PR #631, opened 2024-07-30
Feat: Unruggable fee claim [SC-726]

fabianschu:


test/factories/workflow-specific/PIM_WorkflowFactory_v1.t.sol
```diff
@@ -360,45 +337,106 @@ contract PIM_WorkflowFactory_v1Test is E2ETest {
...
+        // TODO: why does this fail if the fourth boolean is set to true?
+        vm.expectEmit(true, true, true, false);
+        emit IBondingCurveBase_v1.ProjectCollateralFeeWithdrawn(address(this), 0);
+        factory.withdrawPimFee(address(orchestrator.fundingManager()), alice);
+
+    }
+
+    // TODO: I dont understand why this test is failing. If I comment out the `vm.expectRevert` I can see that the `withdrawPimFee` function reverts
+
+    // function testWithdrawPimFee__FailsIfCallerIsNotPimFeeRecipient() public {
```
fabianschu: This test confuses me. I want to check that the fee withdrawal fails, if the caller is not stored as eligible fee recipient.

test/factories/workflow-specific/PIM_WorkflowFactory_v1.t.sol
```diff
@@ -360,45 +337,106 @@ contract PIM_WorkflowFactory_v1Test is E2ETest {
...
+    //         })
+    //     );
+    //     // TODO: I also dont understand why using `vm.startPrank` behaves differently than using `vm.prank`
+
+    //     // 1. VERSION USING STARTPRANK
+    //     //
+    //     vm.startPrank(alice);
+    //     // vm.expectRevert();
+    //     factory.withdrawPimFee(address(orchestrator.fundingManager()), alice);
+    //     vm.stopPrank();
```
fabianschu: If I uncomment this then l.433 reverts with the expected error. If I then uncomment the `expectRevert` check, the test fails because "call did not revert as expected". 

I've played with the order of startPrank and expectRevert, didn't help.

test/factories/workflow-specific/PIM_WorkflowFactory_v1.t.sol
```diff
@@ -360,45 +337,106 @@ contract PIM_WorkflowFactory_v1Test is E2ETest {
...
+
+    //     // 2. VERSION USING PRANK
+    //     // 
+    //     // // vm.expectRevert();
+    //     // vm.prank(alice);
+    //     // factory.withdrawPimFee(address(orchestrator.fundingManager()), alice);
```
fabianschu: If I uncomment this instead, so that I have just the following lines, the call doesn't even revert anymore. It seems as if `prank` and `startPrank` are behaving differently in both cases.

```
        vm.prank(alice);
        factory.withdrawPimFee(address(orchestrator.fundingManager()), alice);
```

test/factories/workflow-specific/PIM_WorkflowFactory_v1.t.sol
```diff
@@ -360,45 +337,106 @@ contract PIM_WorkflowFactory_v1Test is E2ETest {
...
+    //     vm.startPrank(alice);
+    //     // vm.expectRevert();
+    //     factory.withdrawPimFee(address(orchestrator.fundingManager()), alice);
+    //     vm.stopPrank();
+
+    //     // 2. VERSION USING PRANK
+    //     // 
+    //     // vm.expectRevert();
+    //     // vm.prank(alice);
+    //     // factory.withdrawPimFee(address(orchestrator.fundingManager()), alice);
```
fabianschu: If I uncomment these two lines, then it seems as if the `withdrawPimFee` isn't even reverting anymore. It behaves different from what `startPrank` does above.

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
```diff
@@ -27,10 +32,33 @@ contract PIM_WorkflowFactory_v1 is
...
+
+    modifier onlyPimFeeRecipient(address fundingManager) {
+        console.log("MODIFIER");
+        console.log("FUNDING MANAGER: ", fundingManager);
+        console.log("RECIPIENT: ", _pimFeeRecipients[fundingManager]);
+        if (_msgSender() != _pimFeeRecipients[fundingManager]) {
+            revert PIM_WorkflowFactory__OnlyPimFeeRecipient();
+        }
+        _;
+    }
```
fabianschu: This modifier makes the eligibility check and should revert.

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
```diff
@@ -132,19 +167,32 @@ contract PIM_WorkflowFactory_v1 is
...
+    // onlyPimFeeRecipient
+
+    function withdrawPimFee(address fundingManager, address to)
+        external
+        onlyPimFeeRecipient(fundingManager)
+    {
+        IBondingCurveBase_v1(fundingManager).withdrawProjectCollateralFee(
+            to, IBondingCurveBase_v1(fundingManager).projectCollateralFeeCollected()
+        );
+    }
```
fabianschu: Here is the implementation for the withdrawal function.

src/factories/interfaces/IPIM_WorkflowFactory_v1.sol
```diff
@@ -21,6 +21,10 @@ import {ERC20Issuance_v1} from "src/external/token/ERC20Issuance_v1.sol";
 import {IERC20} from "@oz/token/ERC20/IERC20.sol";
 
 interface IPIM_WorkflowFactory_v1 {
+    //--------------------------------------------------------------------------
+    // Errors
+    error PIM_WorkflowFactory__OnlyPimFeeRecipient();
```
marvinkruse, line 27: I believe we should add a `@notice` tag to errors as well, at least that is what we did so far (I believe, just checked a few contracts)?

src/factories/interfaces/IPIM_WorkflowFactory_v1.sol
```diff
@@ -40,9 +44,24 @@ interface IPIM_WorkflowFactory_v1 {
...
+
+    /// @notice Event emitted when factory owner withdraws accumulated creation fees.
+    /// @param fundingManager The address of the funding manager from which to withdraw fees.
+    /// @param to The address to which the fees are sent.
+    /// @param amount The amount of fees that were withdrawn.
+    event CreationFeeWithdrawn(
+        address indexed fundingManager, address indexed to, uint amount
+    );
+
+    /// @notice Event emitted factory owner sets new fee..
```
marvinkruse: Missing a `when` between `emitted` and `factory`.

marvinkruse: Also ends with `..` instead of `.`.

src/factories/interfaces/IPIM_WorkflowFactory_v1.sol
```diff
@@ -85,33 +106,46 @@ interface IPIM_WorkflowFactory_v1 {
...
+
+    /// @notice Updates who can claim the buy/sell fees of a given bonding curve.
+    /// @dev Only callable by the currently eligiblefee recipient.
+    /// @param fundingManager The address of the bonding curve from which to withdraw fees.
+    /// @param to The address that should be eligible to claim fees in the future.
+    function transferPimFeeEligibility(address fundingManager, address to)
+        external;
+
+    /// @notice Withdraws the buy/sell fees of a given bonding curve.
+    /// @dev Only callable by the currently eligiblefee recipient.
```
marvinkruse: Missing space in `eligiblefee` I believe

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
```diff
@@ -104,10 +129,12 @@ contract PIM_WorkflowFactory_v1 is
...
         }
 
-        // if renounced workflow flag is set, renounce admin rights over workflow, else transfer admin rights to initial admin
+        // if renounced workflow flag is set factory keeps admin rights over workflow, else transfer admin rights to initial admin
         if (PIMConfig.isRenouncedWorkflow) {
-            _transferWorkflowAdminRights(orchestrator, address(0));
+            // record the deployer as fee recipient eligible to claim buy/sell fees
+            _pimFeeRecipients[fundingManager] = _msgSender();
         } else {
+            // TODO: check if initial admin is the right target
```
marvinkruse: `TODO`, just FYI.

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
```diff
@@ -27,10 +29,30 @@ contract PIM_WorkflowFactory_v1 is
...
+    //--------------------------------------------------------------------------
+    // State Variables
+
     // store address of orchestratorfactory
     address public orchestratorFactory;
     // relative fees on collateral token in basis points
-    uint public fee;
+    uint public creationFee;
+
+    // mapping of orchestrator address to fee recipient address
```
marvinkruse: Mentions `orchestrator address` but mapping is using `fundingManager`

marvinkruse (changes requested):
Good stuff, left some comments!

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
fabianschu: Thanks! I added this because I found it a bit weird to assign the workflow admin role to the issuance token admin. What I did now is to pass a distinct `admin` param to the factory (within the `PIMConfig` struct) that I use for both: token owner and workflow admin rights. I know this doesn't really belong into this PR, but I hope it's fine anyways 👼🏼

src/factories/interfaces/IPIM_WorkflowFactory_v1.sol
```diff
@@ -23,6 +23,7 @@ import {IERC20} from "@oz/token/ERC20/IERC20.sol";
 interface IPIM_WorkflowFactory_v1 {
     //--------------------------------------------------------------------------
     // Errors
+    /// @notice Error thrown when an unpermissioned address tries to claim fees to to transfer role.
```
marvinkruse: `tries to claim fees to to transfer role` seems like it contains a `to` too much? :D

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
```diff
@@ -125,20 +125,15 @@ contract PIM_WorkflowFactory_v1 is
...
             _transferTokenOwnership(issuanceToken, address(0));
         } else {
-            _transferTokenOwnership(
-                issuanceToken, PIMConfig.issuanceTokenParams.initialAdmin
-            );
+            _transferTokenOwnership(issuanceToken, PIMConfig.admin);
         }
 
-        // if renounced workflow flag is set factory keeps admin rights over workflow, else transfer admin rights to initial admin
+        // if renounced workflow flag is set factory keeps admin rights over workflow, else transfer admin rights to deployer
```
marvinkruse: is it the `deployer` technically? Isn't it a value that the deployer may set, but doesn't need to be the deployer? Or is this just because the UI will set this to the same address as the deployer normally?

marvinkruse (changes requested):
Thank you for the quick fixes, 2 more comments :)

src/factories/workflow-specific/PIM_WorkflowFactory_v1.sol
fabianschu: outdated docs, it should be the "admin specified in the params", will change.

marvinkruse (approved):
thank you for the lightning-fast fixes 🙏 
lgtm! probably good if @0xNuggan looks over this as well if he can fine time.

0xNuggan (approved):
Looks good! Good to rename the creationFee

PR #627, opened 2024-07-26
[SC-446] Task: Update Deployment Scripts

0xNuggan:

PR #632, opened 2024-07-31
[SC-446] Task: Update Deployment Scripts v2

FHieser:
Redid the PR #627 because of the merge issues.


script/SingletonDeployer_v1.s.sol
```diff
@@ -0,0 +1,275 @@
...
+
+    //Authorizer
+    address public mod_Aut_Roles_v1;
+    address public mod_Aut_TokenGated_Roles_v1;
+    address public mod_Aut_EXT_VotingRoles_v1;
+
+    //Funding Managers
+    address public mod_FM_BC_Bancor_Redeeming_VirtualSupply_v1;
+    address public mod_FM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1;
+    address public mod_FM_Rebasing_v1; //@note Do we add this?
```
0xNuggan: We can leave it out since we are removing it.

script/SingletonDeployer_v1.s.sol
```diff
@@ -0,0 +1,275 @@
...
+
+    function createExternalSingletons() public {
+        console2.log(
+            "--------------------------------------------------------------------------------"
+        );
+        console2.log("Create External Singletons");
+        console2.log("-External Contracts");
+
+        //External contracts
+        ext_InverterReverter_v1 = factory.deployWithCreate2( //@note no Beacon structure here right?
```
0xNuggan: I think that's correct, no beacon

script/SingletonDeployer_v1.s.sol
```diff
@@ -0,0 +1,275 @@
...
+        factorySalt, vm.getCode("InverterReverte_v1.sol:InverterReverter_v1")); //@note can we add a string parameter for the factory? So we can log which contract is called here?
```
0xNuggan: If not maybe add a log before?  "---------\n Deploying Governor implementation:" and then the output of the factory.

script/SingletonDeployer_v1.s.sol
```diff
@@ -0,0 +1,275 @@
...
+            )
+        );
+
+        //Modules
+        console2.log("-Modules");
+
+        //Authorizer
+        console2.log("--Authorizer");
+
+        mod_Aut_Roles_v1 = factory.deployWithCreate2(
```
0xNuggan: for readability, I'd mention in the variable names that these are the implementations. What do you think?

script/SingletonDeployer_v1.s.sol
```diff
@@ -0,0 +1,275 @@
...
+    bytes32 factorySalt = keccak256(abi.encodePacked("inverter-deployment"));
+
+    //External contracts
+
+    address public ext_InverterReverter_v1;
+    address public ext_Governor_v1;
+    address public ext_TransactionForwarder_v1;
+    address public ext_FeeManager_v1;
+
+    //@note Add IssuanceToken here?
```
0xNuggan: I don't think so. Each project should create its own IssuanceToken. We could create a factory contract for it, but in my opinion this is overkill for now. Fabi's PIM factory should be enough for now.

0xNuggan:
Thanks, this looks good! Left some comments. So missing here are:
- Testing that everything works (pending on merge with the create2 PR
- A "Testnet deployment script" that does the same as the deployment script, but also adds some periphery contracts like the BancorFormula, ERC20Mock and UMAoracleMock
- A script to deploy implementation and beacon of a new standalone module, and register it in an existing factory
- Some workflow deployment scripts based on our usecases.

Also as a question: When we develop a new module and add it here, which are the files we would need to touch and where? I think it would be valuable to document that


Great work! 🔥

dev.env
```diff
@@ -13,42 +13,52 @@
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 # ------------------------------------------------------------------------------
-# Common
+
+# RPC endpoints (these are public links, subsitute with alchemy ir similar for less rate limiting)
```
marvinkruse:
```suggestion
# RPC endpoints (these are public links, substitute with alchemy or similar for less rate limiting)
```

script/DeploymentScript.s.sol
```diff
@@ -0,0 +1,128 @@
...
+
+import {ModuleBeaconDeployer_v1} from "script/ModuleBeaconDeployer_v1.s.sol";
+
+import {Governor_v1} from "@ex/governance/Governor_v1.sol";
+import {FeeManager_v1} from "@ex/fees/FeeManager_v1.sol";
+import {ModuleFactory_v1} from "src/factories/ModuleFactory_v1.sol";
+import {OrchestratorFactory_v1} from "src/factories/OrchestratorFactory_v1.sol";
+
+contract DeploymentScript is ModuleBeaconDeployer_v1 {
+    //@note not upgradable right?
```
marvinkruse: You mean the `Reverter`? If so, yes, non-upgradeable.

script/DeploymentScript.s.sol
```diff
@@ -0,0 +1,128 @@
...
+    address public moduleFactory;
+    address public orchestratorFactory;
+
+    function run() public {
+        console2.log(
+            "--------------------------------------------------------------------------------"
+        );
+        console2.log("Start Deployment");
+
+        //Create External Singletons
```
marvinkruse: General thing: Please leave a space between `//` and the comment :D If you want to have an easy way to not change your habits and still provide code that doesn't trigger me (😂) here is a regex command that can help you detect any of these without false positives.

```regex
(?<!:)\/\/(?!\s|\/|-{2,})
```

Just put it into the search on VSCode, click the "regex" button that looks like `.*` and exclude `lib`. Then just put `// ` (with a whitespace after the //) into the replace line and go for it. 

Screenshot 2024-08-01 at 13 12 07

marvinkruse: And sorry in advance for this 😂

marvinkruse (changes requested):
General things have been mentioned by @0xNuggan already I believe, will add more comments later.

Really good work 🙏

script/DeploymentScript.s.sol
FHieser: lol

FHieser: dog-trigger

marvinkruse (approved):

PR #634, opened 2024-08-02
[SC-760] Create Workflow Deployment Scripts

marvinkruse:
Ready to be worked on now.


marvinkruse:
@0xNuggan This is the last commit that Felix added to his Scripts PR, now based on `dev` with his script changes. You can either build upon it or scrap it and do it however you want. Just FYI.

script/utils/CreateAndDeployModuleBeacon.s.sol
FHieser, line 1: In regards of this todo -> // @TODO Above explanation is lacking the bytes-encoded optionalParams
The ??? should be replaced by the following:
<use the result of abi.encode(<paramter1>,<parameter2>, ....) in a solidity shell>

0xNuggan:
**Note before launch:** The main components of this PR have been merged into dev already. This PR will stay open after mainnet launch to be updated with the last batch of changes and filled out with the missing workflow scripts.

fabianschu:
I did two changes:
1. there was a bug in how the curve interaction role was assigned to the workflow admin after deployment which is fixed
2. I removed the issuance token as return value from the `createPIMWorkflows` function to have consistency between the regular factory and this one for the sdk

fabianschu:
from my pov this can be closed

Zitzak:
Closing the PR as we internally decided to use the control room for deployments.

PR #635, opened 2024-08-05
[64] - Review Fix : Make asserter pay for the bond

0xNuggan:
Fixed as per comment https://github.com/InverterNetwork/contracts/pull/619#discussion_r1700629388

### DIFF
--- a/src/proxies/InverterBeacon_v1.sol
+++ b/src/proxies/InverterBeacon_v1.sol
@@ -63,10 +63,10 @@ contract InverterBeacon_v1 is IInverterBeacon_v1, ERC165, Ownable2Step {
     ) {
         if (
             // Minor Version cant go down
-            newMinorVersion < minorVersion
             // Patch Version cant go down or stay the same if minorVersion stays the same
-            || newPatchVersion <= patchVersion
-                && newMinorVersion == minorVersion
+            newMinorVersion < minorVersion
+                || newPatchVersion <= patchVersion
+                    && newMinorVersion == minorVersion
         ) {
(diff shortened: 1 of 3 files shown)


src/proxies/InverterBeacon_v1.sol
```diff
@@ -63,10 +63,10 @@ contract InverterBeacon_v1 is IInverterBeacon_v1, ERC165, Ownable2Step {
     ) {
         if (
             // Minor Version cant go down
-            newMinorVersion < minorVersion
             // Patch Version cant go down or stay the same if minorVersion stays the same
-            || newPatchVersion <= patchVersion
-                && newMinorVersion == minorVersion
+            newMinorVersion < minorVersion
+                || newPatchVersion <= patchVersion
+                    && newMinorVersion == minorVersion
```
marvinkruse, line 69: What happened here? Your version looks fine, but interesting that the "before" part looks so weird...

marvinkruse (approved):

src/proxies/InverterBeacon_v1.sol
marvinkruse: Oh I see, you just didn't like the formatting, the version before was fine, just had comments within the expression :D Or did you not change this? @0xNuggan

marvinkruse: Oh it formats it the same for me locally after updating my Foundry, ignore this 😂

0xNuggan: That was the formatter. The comments were originally inlined directly inside the if() statement, which probably goes against some rule.

PR #637, opened 2024-08-05
Fix: Replace Module Metadata Link with correct ones

marvinkruse:
The last commit with the formatting will be rebased away once #636 is merged.


marvinkruse:
Might fail until #636 is merged, as the CI is failing quite often because of that test. So if this fails just ignore for now, what is changed here is trivial.

0xNuggan (approved):

PR #640, opened 2024-08-06
[Audit] Apply Hats Fixes from Macro Audit

marvinkruse:
Applying the fixes after the Hats Review of the Macro Audit PR #540 (slightly adapted the one from @Zitzak as the `init` function of the `AUT_Roles_v1` contract is very different.

(Did a modification that was probably required, the discussion can be found below in my comment.)

Branch name was a mistake, should have been `apply-macro-audit-omega-fixes`

### DIFF
--- a/src/modules/authorizer/IAuthorizer_v1.sol
+++ b/src/modules/authorizer/IAuthorizer_v1.sol
@@ -27,13 +27,16 @@ interface IAuthorizer_v1 is IAccessControlEnumerable {
     //--------------------------------------------------------------------------
     // Functions
 
-    /// @notice Asks whether an address holds the required module role to execute
+    /// @notice Asks whether an address holds the required role to execute
     ///         the current transaction.
+    /// @dev The calling contract needs to generate the right role ID using its
+    ///      own address and the role identifier.
+    ///      In modules, this function should be used instead of hasRole, as
+    ///      there are Authorizer-specific checks that need to be performed.
     /// @param role The identifier of the role we want to check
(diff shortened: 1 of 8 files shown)


marvinkruse:
@0xNuggan I did another commit here, as the whole role check wasn't working and a bit confusing after the merge with `dev`. The problem that with the new `hasModuleRole` it now generated the role id twice (once in the base `Module_v1` and a second time in `hasModuleRole`).

While looking at it, I came to the conclusion, that we do need something else than `hasRole` (so we can enforce authorizer-specific logik like token-gated checks), but it should **not** generate the role id so it's more versatile. My [last commit](https://github.com/InverterNetwork/contracts/pull/640/commits/bab4637fef84c1a4ba25c4de2c7d42a17e2005ed) (_please check it out specifically if possible_) addresses this and replaces `hasModuleRole` with a generalized `checkForRole` function that should be used (and just calls the underlying `hasRole` in case there is no token gating or so.

We can maybe discuss this briefly after the Tech Sync, if you don't like this or have a better idea feel free to just remove my commit again, all good :)

0xNuggan:
Thanks for this! As long as when granting a role the authorizer encodes the name with the module address (which still happens), the changes should be fine. Also, good catch on the base module_v1 modifier

0xNuggan:
Do you think this is something that needs to be notified explicitly to the SDK team? in case they are sending encoded/nonencoded `roleId`s around

0xNuggan (approved):

PR #641, opened 2024-08-06
Bug(SC-755): Rephrase disclaimer in Orchestrator for clarity

Zitzak:


src/orchestrator/interfaces/IOrchestrator_v1.sol
```diff
@@ -137,7 +137,7 @@ interface IOrchestrator_v1 is IModuleManagerBase_v1 {
         external;
 
     /// @notice Executes replaces the current payment processor with `paymentProcessor_`
-    /// @notice !!! IMPORTANT !!! When changing the PaymentProcessor the current ongoing payment orders are lost. Make sure to resolve those payments properly beforehand.
+    /// @notice !!! IMPORTANT !!! When changing the PaymentProcessor the current ongoing payments are lost. Make sure all payment orders in the Logic Modules are resolved properly beforehand.
```
marvinkruse: Looks good to me. Only issue I have is that the original version already didn't follow our "roughly 80 character" linebreak rule, so could you maybe fix that and add a linebreak there?

marvinkruse (approved):

PR #642, opened 2024-08-06
Chore: Rename Dependency

marvinkruse:
**Status**
⌛ Waiting for review...

--- 
**Description**
Found out that there is a way to define the name of `forge install` dependencies, so I've renamed the UMA protocol dependency from `protocol` to `uma-protocol`, to make it a bit clearer.

Made sure to use the same commit hash for the dependency itself, so nothing changed there.

Lmk if this is okay for you in general @0xNuggan 🙏


0xNuggan (approved):
nice, didn't know you could do that.

PR #648, opened 2024-08-09
[SC-754] Bug: fix minimum decimal amount for issuance token

Zitzak:


src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -541,10 +541,23 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
...
+        } else {
+            // If issuance token has more than 18 decimals
+            decimalDifference = issuanceTokenDecimals - 18;
+        }
+        minRequiredSupply = 10 ** decimalDifference;
         // Check if virtual supply is big enough to ensure compatibility with relative issuance
         // token decimal and conversion to 18 decimals done in FM_BC_Tools._convertAmountToRequiredDecimal()
         // so it will not result in a round down 0 value
-        if (_virtualSupply < 10 ** (issuanceTokenDecimals - 18)) {
+        if (_virtualSupply < minRequiredSupply) {
```
marvinkruse: Logic is correct from what I checked. Just some gas stuff.

Caching the `issuanceTokenDecimals` saves gas, as we otherwise access contract state 4 times. We also declare `minRequiredSupply` above, set it later and then just use it once, so it's cheaper to just use that directly, saves an `MSTORE` and an `MLOAD`.

```suggestion
        uint decimalDifference;
        uint decimals = issuanceTokenDecimals;

        if (decimals <= 18) {
            // If issuance token has 18 or fewer decimals. Ensure that decimal difference is at least 1
            // to end up with minRequiredSupply of 10
            decimalDifference =
                18 - decimals == 0 ? 1 : 18 - decimals;
        } else {
            // If issuance token has more than 18 decimals
            decimalDifference = decimals - 18;
        }
        
        // Check if virtual supply is big enough to ensure compatibility with relative issuance
        // token decimal and conversion to 18 decimals done in FM_BC_Tools._convertAmountToRequiredDecimal()
        // so it will not result in a round down 0 value
        if (_virtualSupply < 10 ** decimalDifference) {
```

test/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.t.sol
```diff
@@ -1160,7 +1160,7 @@ contract FM_BC_Bancor_Redeeming_VirtualSupplyV1Test is ModuleTest {
     /* Test getVirtualIssuanceSupply() */
 
     function testGetVirtualIssuanceSupply(uint _supply) public {
-        vm.assume(_supply > 0);
+        vm.assume(_supply > 10);
```
marvinkruse, line 1163: Should we use `INITIAL_ISSUANCE_SUPPLY` here just to be clear of where this number comes from?

test/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.t.sol
```diff
@@ -1268,7 +1268,7 @@ contract FM_BC_Bancor_Redeeming_VirtualSupplyV1Test is ModuleTest {
     function testInternalSetVirtualIssuanceSupply_WorksGivenIssuanceSupplyBiggerThanZero(
         uint _newSupply
     ) public {
-        vm.assume(_newSupply != 0);
+        vm.assume(_newSupply > 10);
```
marvinkruse, line 1271: Same as above, should we use `INITIAL_ISSUANCE_SUPPLY` here just to be clear of where this number comes from?

marvinkruse (changes requested):

test/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.t.sol
marvinkruse: Not doing this, not needed.

test/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.t.sol
marvinkruse: Not doing this, not needed.

marvinkruse (approved):

PR #649, opened 2024-08-09
Scripts: Improve Scripts Before Deployment

marvinkruse:
I took the changes that @0xNuggan did in #634 on the general scripts and improved them from there. What changed?
- Harmonized the logged output so it looks nice ([see screenshot](https://github.com/user-attachments/assets/1bc08857-7647-4bdd-81ed-4476cbe0b551))
- Add checks in scripts to ensure that deployment was successful
  - Basically checking if the important contracts are initialized correctly and ensure that no-one just initialized them before we could
- Ensure correctly set addresses, otherwise revert via check modifier
  - If multisigs, treasury or deterministic factory is not set -> revert
  - In the Testnet script, I just set these to the deployer if they wouldn't be set otherwise
  - For this to work, I no have these addresses as the zero-address in the env file. There was no reason to have this as 0x0...01, etc.
- Simplified env file further
- Cleaned the imports up (i.e. `SingletonDeployer` doesn't need everything imported)
- Set `deterministic-factory` dependency to proper tag (v1.0)

### DIFF
--- a/dev.env
+++ b/dev.env
@@ -26,32 +26,27 @@ export POLYGON_CARDONA_RPC_URL=https://rpc.cardona.zkevm-rpc.com
 # ------------------------------------------------------------------------------
 # Wallets
 
-# Default Anvil Wallets
+# Default Anvil Wallet
 export ANVIL_WALLET_DEPLOYER=0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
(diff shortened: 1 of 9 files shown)


0xNuggan (changes requested):
Thanks fot this!! Looks great. As mentioned in the DMs, we should move the BancorFormula to the general deployment script, but after that that I think we can merge it

0xNuggan (approved):

PR #655, opened 2024-08-16
Fix: Resolve Deployment Issues

marvinkruse:
This PR resolves...
* the TransactionForwarder not being upgradeable (was deployed as upgradeable though, hence the change)
* the Scripts not catching that, so I added an extensive list of checks that are performed at the end of the deployment
  * Check is Governor owns everything + the right multisigs own the Governor
  * Factories, Governor + Fee Manager are initialized correctly
  * Modules are all deployed with matching metadata
  * Beacons are all pointing to the right reverted and reference the right version
  * TransactionForwarder is initialized correctly (EIP712 data)
* Typo in the title of a few contracts

### DIFF
--- a/script/deploymentScript/DeploymentScript.s.sol
+++ b/script/deploymentScript/DeploymentScript.s.sol
@@ -7,13 +7,25 @@ import "forge-std/Script.sol";
 import {ModuleBeaconDeployer_v1} from
     "script/deploymentSuite/ModuleBeaconDeployer_v1.s.sol";
 
-// Constants
-import {Governor_v1} from "@ex/governance/Governor_v1.sol";
-import {FeeManager_v1} from "@ex/fees/FeeManager_v1.sol";
-import {ModuleFactory_v1} from "src/factories/ModuleFactory_v1.sol";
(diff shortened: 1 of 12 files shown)


Zitzak:
Question around the TxForwarder, I thought it wasn't supposed to be upgradeable, but was deploy as if. But looking at the changes we are going with it being upgradeable? @marvinkruse 

Or did I have the wrong?

Zitzak (approved):
LGTM

marvinkruse:
> Question around the TxForwarder, I thought it wasn't supposed to be upgradeable, but was deploy as if. But looking at the changes we are going with it being upgradeable? @marvinkruse
> 
> Or did I have the wrong?

No, the discussion was a bit confusing, you don't have anything wrong. Original idea was to have it non-upgradeable, as that is just a bit safer for the user. But that has the big downside that if there is a bug in there, we can't ever change it, as the tx forwarder is hardcoded in each contract. so we'll have it also upgradeable, and with that being behind our timelock etc. it's fine.

That was also the only option where we basically don't change the code (besides the imports), the other option would have been to change all other contracts and have the tx forwarder address be changeable.

Does that make sense?

0xNuggan (approved):

PR #658, opened 2024-08-27
Feat: PIM factories (restricted and immutable)

fabianschu:
This PR consolidates the efforts on the two PIM factories:
* it removes the outdated generalized factory
* it adds the restricted factory (Qacc use case)
* it adds the immutable factory (coin launcher use case)


fabianschu:
Open questions for me at this point:

Restricted PIM Factory:
* right now all privileges that exist go to the address that is set as `initialAdmin`
* this comprises:
  * being workflow admin (1)
  * receiving initialIssuanceSupply (2)
  * being token owner for issuance token (3)
  * receiving CURVE_INTERACTION_ROLE (4) 

However, I learned this might not be ideal for qacc, since the participating projects won't be orchestrator admins. Would it make sense to make a "split" between (1) and (2,3,4)? So the user passes another address to the factory that receives privileges (2,3,4) while the address passed as `initialAdmin` "only" receives privilege (1)? @0xNuggan

script/deploymentScript/DeployWorkflowSpecificFactory.s.sol
```diff
@@ -0,0 +1,55 @@
...
+
+        vm.startBroadcast(vm.envUint("DEPLOYER_PRIVATE_KEY"));
+
+        if (isEqual(factoryType, "RESTRICTED")) {
+            {
+                console2.log(
+                    "Deploying Restricted_PIM_Factory_v1 at address: ",
+                    address(
+                        new Restricted_PIM_Factory_v1(
+                            orchestratorFactory, trustedForwarder
```
marvinkruse: I could be wrong, but this would also execute if both addresses (or at least one) is `address(0)`, right? Maybe we should add something like we have in the `DeploymentScript.s.sol` to ensure that the deployment is gonna be okay? (see [here](https://github.com/InverterNetwork/contracts/blob/2a8a4c80ff4f24a59546d4e6126b81bc51228c94/script/deploymentScript/DeploymentScript.s.sol#L458))

src/factories/interfaces/IImmutable_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,91 @@
...
+        address indexed fundingManager,
+        address indexed claimer,
+        address indexed to,
+        uint amount
+    );
+
+    //--------------------------------------------------------------------------
+    // Functions
+
+    /// @notice Deploys a new issuance token and uses that to deploy a workflow with an open yet immutable bonding curve.
```
marvinkruse: I believe @Zitzak reformatted our natspec to also adhere to a specific max length, did you double check this here already? otherwise we might wanna do that

src/factories/interfaces/IImmutable_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,91 @@
...
+    /// @param workflowConfig The workflow's config data.
+    /// @param fundingManagerConfig The config data for the orchestrator's {IFundingManager_v1} instance.
+    /// @param authorizerConfig The config data for the orchestrator's {IAuthorizer_v1} instance.
+    /// @param paymentProcessorConfig The config data for the orchestrator's {IPaymentProcessor_v1} instance.
+    /// @param moduleConfigs Variable length set of optional module's config data.
+    /// @param issuanceTokenParams The issuance token's parameters (name, symbol, decimals, maxSupply).
+    /// @param initialPurchaseAmount The volume of the first purchase in terms of collateral token.
+    /// @return CreatedOrchestrator Returns the created orchestrator instance.
```
marvinkruse: I believe the first letter is usually lower-case? `createdOrchestrator`

src/factories/workflow-specific/Immutable_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,184 @@
...
+// External Interfaces
+import {IERC20} from "@oz/token/ERC20/IERC20.sol";
+
+// External Implementations
+import {ERC20Issuance_v1} from "src/external/token/ERC20Issuance_v1.sol";
+
+// External Dependencies
+import {ERC2771Context, Context} from "@oz/metatx/ERC2771Context.sol";
+
+contract Immutable_PIM_Factory_v1 is
```
marvinkruse: Missing the header natspec (title, description, security notice)

src/factories/workflow-specific/Immutable_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,184 @@
...
+    ERC2771Context,
+    IImmutable_PIM_Factory_v1
+{
+    //--------------------------------------------------------------------------
+    // State Variables
+
+    /// @dev	store address of {Orchestratorfactory_v1}.
```
marvinkruse:
```suggestion
    /// @dev	Store address of {Orchestratorfactory_v1}.
```

src/factories/workflow-specific/Immutable_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,184 @@
...
+    address public orchestratorFactory;
+
+    /// @dev	mapping of Funding Manager address to `feeRecipient` address.
```
marvinkruse:
```suggestion
    /// @dev	Mapping of Funding Manager address to `feeRecipient` address.
```

src/factories/workflow-specific/Immutable_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,184 @@
...
+    ) external returns (IOrchestrator_v1 orchestrator) {
+        // deploy issuance token
+        ERC20Issuance_v1 issuanceToken = new ERC20Issuance_v1(
+            issuanceTokenParams.name,
+            issuanceTokenParams.symbol,
+            issuanceTokenParams.decimals,
+            issuanceTokenParams.maxSupply,
+            address(this) // assigns owner role to itself initially to manage minting rights temporarily
+        );
+        // MODIFY AUTHORIZER CONFIG
```
marvinkruse: A line break between logical sections would make sense IMHO, so here before this like (`// MODIFY AUTHORIZER CONFIG`), then again before `// MODIFY AUTHORIZER CONFIG`, etc.

src/factories/workflow-specific/Immutable_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,184 @@
...
+            (
+                address,
+                IFM_BC_Bancor_Redeeming_VirtualSupply_v1.BondingCurveProperties,
+                address
+            )
+        );
+        fundingManagerConfigData =
+            abi.encode(address(issuanceToken), bcProperties, collateralToken);
+        fundingManagerConfig.configData = fundingManagerConfigData;
+        orchestrator = IOrchestratorFactory_v1(orchestratorFactory)
```
marvinkruse: Line break here could also help visually, before the `createOrchestrator` call.

src/factories/workflow-specific/Immutable_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,184 @@
...
+            fundingManager, _msgSender(), to, amount
+        );
+    }
+
+    /// @inheritdoc IImmutable_PIM_Factory_v1
+    function transferPimFeeEligibility(address fundingManager, address to)
+        external
+        onlyPimFeeRecipient(fundingManager)
+    {
+        _pimFeeRecipients[fundingManager] = to;
```
marvinkruse: Should we maybe check for an invalid (i.e. zero) address here? The risk of having a wrong address here won't be resolved (unless we add a 2-step-mechanism) but maybe just slightly reduced? Not sure.

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,186 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity 0.8.23;
+
+import "forge-std/console.sol";
```
marvinkruse: Left the console in.
```suggestion
```

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,186 @@
...
+import {ERC2771Context, Context} from "@oz/metatx/ERC2771Context.sol";
+
+contract Restricted_PIM_Factory_v1 is
+    ERC2771Context,
+    IRestricted_PIM_Factory_v1
+{
+    //--------------------------------------------------------------------------
+    // State Variables
+
+    // store address of orchestratorfactory
```
marvinkruse:
```suggestion
    // Store address of {Orchestratorfactory_v1}.
```

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,186 @@
...
+            issuanceToken.transferOwnership(newAdmin);
+        }
+    }
+
+    function _transferWorkflowAdminRights(
+        IOrchestrator_v1 orchestrator,
+        address newAdmin
+    ) private {
+        bytes32 adminRole = orchestrator.authorizer().getAdminRole();
+        // if renounced flag is set, add zero address as admin (because workflow must have at least one admin set)
```
marvinkruse: Is this comment accurate? It's a bit confusing as I don't see this being done here unless `realAdmin` is set to zero? Maybe I'm missing something.

src/modules/fundingManager/bondingCurve/interfaces/IFM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.0;
+
+interface IFM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1 {
```
marvinkruse, line 4: Shouldn't this also contain the error `Module__FM_BC_Restricted_Bancor_Redeeming_VirtualSupply__FeatureDeactivated` as that is specific to the `FM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1`?

script/deploymentScript/DeployWorkflowSpecificFactory.s.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import "forge-std/Script.sol";
+import "forge-std/console.sol";
```
marvinkruse: `console` is already imported via `Script` (see [here](https://github.com/foundry-rs/forge-std/blob/master/src/Script.sol))

```suggestion
```

src/factories/interfaces/IImmutable_PIM_Factory_v1.sol
marvinkruse:
```suggestion
    /// @return createdOrchestrator Returns the created orchestrator instance.
```

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,186 @@
...
+// External Interfaces
+import {IERC20} from "@oz/token/ERC20/IERC20.sol";
+
+// External Implementations
+import {ERC20Issuance_v1} from "src/external/token/ERC20Issuance_v1.sol";
+
+// External Dependencies
+import {ERC2771Context, Context} from "@oz/metatx/ERC2771Context.sol";
+
+contract Restricted_PIM_Factory_v1 is
```
marvinkruse: Missing natspec header (title, description, security notice) as well

marvinkruse (changes requested):
Good stuff in general, nice! 🙌 

(I haven't fully looked through the tests yet, just glossed over them, but the rest I reviewed.)

script/deploymentScript/DeployWorkflowSpecificFactory.s.sol
fabianschu: Great idea!

src/factories/interfaces/IImmutable_PIM_Factory_v1.sol
fabianschu: Yeah, looks a bit long, I shortened that one but I'm not sure where I can find the actual max length. @Zitzak can you help out? :)

fabianschu:
I addressed your comments @marvinkruse. 
In addition to that I had to make some additional changes to cater for qacc specific needs:
* the restricted factory now contains a pre-funding mechanism where anyone can sponsor the initial collateral supply required for the deployment of a PIM workflow (sponsor can also withdraw their pre-funding again)
* to be able to use pre-funding, deployer needs to set sponsor as workflow admin
* initial issuance supply goes to deployer
* curve interaction role also goes to deployer

Note: fee claim for deployer is planned to be enabled through admin's Multisig via Zodiac Roles

src/factories/interfaces/IImmutable_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    //--------------------------------------------------------------------------
+    // Events
+
+    /// @notice Event emitted when a new PIM workflow is created.
+    /// @param fundingManager The address of the funding manager.
+    /// @param issuanceToken The address of the issuance token.
+    /// @param deployer The address of the deployer.
+    event PIMWorkflowCreated(
+        address indexed fundingManager,
```
0xNuggan: I guess this is for indexing reasons, but why return the fundingManager instead of the Orchestrator? Particularly as the other Interface returns the orchestrator

src/factories/interfaces/IRestricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,108 @@
...
+        address indexed deployer
+    );
+
+    /// @notice Event emitted when new funding is added.
+    /// @param paymaster Address that pays funding.
+    /// @param actor Address that can use new funding.
+    /// @param token Address of token used for funding.
+    /// @param amount Funding amount.
+    event FundingAdded(
+        address indexed paymaster,
```
0xNuggan: Not sure I like calling it "paymaster" since that term is used in account abstraction. Maybe "sponsor" is better?

src/factories/interfaces/IImmutable_PIM_Factory_v1.sol
0xNuggan: The implementation returns the orchestrator address, so thi is a bug :)

fabianschu: woopsies

src/factories/interfaces/IImmutable_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,109 @@
...
+import {IOrchestratorFactory_v1} from
+    "src/factories/interfaces/IOrchestratorFactory_v1.sol";
+import {IBondingCurveBase_v1} from
+    "@fm/bondingCurve/interfaces/IBondingCurveBase_v1.sol";
+
+// Internal Dependencies
+import {ERC20Issuance_v1} from "src/external/token/ERC20Issuance_v1.sol";
+
+/**
+ * @title   Immutable PIM Factory
```
marvinkruse: Our usual naming includes the word "Inverter" in the `@title` now, so "Inverter Payment Router" instead of just "Payment Router". I think we should be consistent here as well then?

"Immutable Inverter PIM Factory" or so? And for Restricted as well.

src/factories/workflow-specific/Immutable_PIM_Factory_v1.sol
marvinkruse: Saw the thumbs-up here but nothing has changed, right? Which is fine, just wanna make sure that we don't forget about it, in case you intended to add something here. If you think this should be left like it is, that's also fine.

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
marvinkruse: You added the Natspec headers to the interfaces, instead of to the contracts. We have them on the contracts, not the interfaces so far. 
Yesterday in the dev guidelines call we decided we want to have it in both actually (as I also see that having it in the interface makes a lot of sense). We will then (later) add the functionality to the CI bot to verify that both interface and contract have the same Natspec header (just so we don't forget to add it for one of them or if we change it but don't update the other one). Does that make sense?

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,238 @@
...
+            moduleConfigs
+        );
+
+        // get bonding curve / funding manager
+        address fundingManager = address(orchestrator.fundingManager());
+
+        // enable bonding curve to mint issuance token
+        issuanceToken.setMinter(fundingManager, true);
+
+        // transfer initial collateral supply from msg.sender to  bonding curve and mint issuance token to recipient
```
marvinkruse: Double space between `to` and `bonding curve`.

```suggestion
        // Transfer initial collateral supply from msg.sender to bonding curve and mint issuance token to recipient
```

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,238 @@
...
+    //--------------------------------------------------------------------------
+    // State Variables
+
+    // Stores address of orchestratorfactory.
+    address public orchestratorFactory;
+    // Stores available fundings.
+    mapping(
+        address sponsor
+            => mapping(address actor => mapping(address token => uint amount))
+    ) fundings;
```
marvinkruse: Missing the visibility keyword.

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,238 @@
...
+        if (availableFunding < initialCollateralSupply) {
+            revert IRestricted_PIM_Factory_v1.InsufficientFunding(
+                availableFunding
+            );
+        }
+
+        fundings[admin][actor][address(collateralToken)] -=
+            initialCollateralSupply;
+
+        // collateral token is paid for by the msg.sender
```
marvinkruse: This comment is confusing me slightly, as we are using the tokens that are deposited by the sponsor, not necessarily the msg.sender?

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,238 @@
...
+        IBondingCurveBase_v1 fundingManager,
+        IERC20 collateralToken,
+        ERC20Issuance_v1 issuanceToken,
+        uint initialCollateralSupply,
+        uint initialIssuanceSupply,
+        address admin,
+        address actor
+    ) private {
+        uint availableFunding = fundings[admin][actor][address(collateralToken)];
+        if (availableFunding < initialCollateralSupply) {
```
marvinkruse: As it reverts here this basically means that it's not an option to pay this as the deployer in that specific moment, there needs to be a sponsor that pre-funded, right? Yes, the actor can be the sponsor, but this entails two calls (first `addFunding` before creating the workflow. Is that intentional? In my mind a fallback mechanism would make sense where the sponsors funds are taken, and if not enough, the deployer is paying (if the approved it)? But if that's not required for the client or I am just missing something then feel free to ignore this ofc.

marvinkruse (changes requested):

script/deploymentScript/DeployWorkflowSpecificFactory.s.sol
marvinkruse: Got a 👍 but no changes, was that intentional? (just double checking)

src/factories/interfaces/IImmutable_PIM_Factory_v1.sol
marvinkruse: Also not changed yet, just fyi.

src/factories/workflow-specific/Immutable_PIM_Factory_v1.sol
marvinkruse: Hasn't changed yet, if I'm not mistaking

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
0xNuggan: Just out of curiosity, @marvinkruse : Do you have some kind of plugin that highlights these kind of style mistakes? I would never have noticed that double space :rofl:

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
0xNuggan: In this case the client will pre-seed the contract with the funds, since projects aren't supposed to pay anything. A fallback mechanism makes sense, but is somewhat out of scope for this case?

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
0xNuggan: yes, it should say either sponsor or factory

script/deploymentScript/DeployWorkflowSpecificFactory.s.sol
fabianschu: No, not intentional. Done now :)

src/factories/workflow-specific/Immutable_PIM_Factory_v1.sol
fabianschu: Sorry, I overlooked this. Added a check now.

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
fabianschu: I just removed the whitespace but didn't capitalize the first letter for consistency :P

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
fabianschu: I made this public and added the getter to the interface.

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
fabianschu: yea, outdated comment is fixed

src/factories/workflow-specific/Restricted_PIM_Factory_v1.sol
fabianschu: If you don't mind I'll keep it qacc-specific and refrain from adding the fallback.

src/factories/interfaces/IImmutable_PIM_Factory_v1.sol
Zitzak: I believe I followed this guideline
https://docs.soliditylang.org/en/latest/style-guide.html#maximum-line-length

marvinkruse:
@fabianschu Overall looks good to me, one thing I just noticed though: we want to deploy the factories any time a testnet deployment goes through, potentially even with every deployment (but I'm not sure about that last part). Two things come to mind there: (a) it should be included in the TestnetDeploymentScript to automatically just deploy both factories and (b) for that it would probably be cool to have not just a `run()` function in the `DeployWorkflowSpecificFactory.s.sol` but also a run(orchestratorFactory, trustedForwarder, factoryType)` or something like that - if that makes sense?

src/factories/custom/Immutable_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,210 @@
...
+        _pimFeeRecipients;
+
+    //--------------------------------------------------------------------------
+    // Modifiers
+
+    /// @dev	Modifier to guarantee the caller is the fee recipient for the given funding manager.
+    modifier onlyPimFeeRecipient(address fundingManager) {
+        if (_msgSender() != _pimFeeRecipients[fundingManager]) {
+            revert PIM_WorkflowFactory__OnlyPimFeeRecipient();
+        }
```
0xNuggan, line 68: Following our new repo guidelines, this should be externalized to an internal function to save gas

0xNuggan: Not gonna be an issue here, but for consistency

test/external/MintWrapper.t.sol
```diff
@@ -0,0 +1,140 @@
...
+        vm.expectEmit(true, true, true, true);
+        emit IERC20Issuance_v1.MinterSet(minter, false);
+
+        wrapper.setMinter(minter, false);
+
+        assertFalse(wrapper.allowedMinters(minter));
+    }
+
+    function test_setMinter_Idempotence(address minter, bool allowed) public {
+        // Sometimes we the initial state is that the address is allowed
```
0xNuggan, line 67: typo

0xNuggan (changes requested):
just two small things

0xNuggan (changes requested):
Just two small things

fabianschu:
Will be split into multiple PRs

PR #661, opened 2024-09-11
Sc 784 bug add event to projectcollateralfeecollected

FHieser:
Add event and rename internal variable


FHieser:
@marvinkruse should I add the version in the contract header as we talked about in the Versioning Call or are we doing this later?

fabianschu (approved):
looks good to me

0xNuggan (approved):
LGTM

PR #662, opened 2024-09-12
Sc 786 story add fee collection to depositvault contract

FHieser:
Updated the Vault to take fees on deposit

### DIFF
--- a/src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
+++ b/src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
@@ -9,7 +9,7 @@ import {IFM_DepositVault_v1} from
     "@fm/depositVault/interfaces/IFM_DepositVault_v1.sol";
 
 // Internal Dependencies
-import {Module_v1} from "src/modules/base/Module_v1.sol";
+import {Module_v1, IModule_v1} from "src/modules/base/Module_v1.sol";
 
 // External Interfaces
 import {IERC20} from "@oz/token/ERC20/extensions/IERC20Metadata.sol";
@@ -145,7 +145,9 @@ contract FM_DepositVault_v1 is
 
             // transfer fee amount
             _token.safeTransfer(_treasury, _feeAmount);
-            emit ProtocolFeeTransferred(address(_token), _treasury, _feeAmount);
+            emit IModule_v1.ProtocolFeeTransferred(
+                address(_token), _treasury, _feeAmount
+            );
         }
     }
 
--- a/src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
+++ b/src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
@@ -102,12 +102,14 @@ contract FM_DepositVault_v1 is
             bytes4(keccak256(bytes("deposit(uint)")))
         );
 
-        uint protocolFeeAmount = amount * collateralFeePercentage / BPS;
+        // skip protocol fee collection if fee is 0
+        if (collateralFeePercentage > 0) {
+            uint protocolFeeAmount = amount * collateralFeePercentage / BPS;
 
-        _processProtocolFeeViaTransfer(
-            collateralTreasury, _token, protocolFeeAmount
-        );
-
+            _processProtocolFeeViaTransfer(
+                collateralTreasury, _token, protocolFeeAmount
+            );
+        }
         emit Deposit(from, amount);
     }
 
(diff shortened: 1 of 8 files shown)


src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
```diff
@@ -2,6 +2,13 @@
 pragma solidity ^0.8.0;
 
 interface IFM_DepositVault_v1 {
+    //--------------------------------------------------------------------------
+    // Errors
+
+    /// @notice Receiver address can not be zero address or.
```
Zitzak:
```suggestion
    /// @notice Receiver address can not be zero address or
```

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
```diff
@@ -94,6 +97,17 @@ contract FM_DepositVault_v1 is
         address from = _msgSender();
         token().safeTransferFrom(from, address(this), amount);
 
+        (uint collateralFeePercentage, address collateralTreasury) =
+        _getFeeManagerCollateralFeeData(
+            bytes4(keccak256(bytes("deposit(uint)")))
+        );
+
+        uint protocolFeeAmount = amount * collateralFeePercentage / BPS;
```
Zitzak: It probably makes sense to have a similar flow here as within the [BondingCurveBase_v1](https://github.com/InverterNetwork/contracts/blob/main/src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol#L357), i.e. check if calculation is needed and check for rounding down to zero situation.

Zitzak: The _processProtocolFeeViaTransfer function could then also be moved into the if statement, removing the need for the check in the funciton.

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
```diff
@@ -10,6 +17,14 @@ interface IFM_DepositVault_v1 {
     /// @param  _amount The amount of tokens deposited.
     event Deposit(address indexed _from, uint _amount);
 
+    /// @notice Event emitted when protocol fee has been transferred to the treasury.
+    /// @param  token The token received as protocol fee.
+    /// @param  treasury The protocol treasury address receiving the token fee amount.
+    /// @param  feeAmount The fee amount transferred to the treasury.
+    event ProtocolFeeTransferred(
```
Zitzak: We are currently duplicating this event across differnt interfaces. Might it be an idea to move it to the IFundingManager_v1 interface, or IModule_v1 interface so that all fee collection has access to the same event?

Zitzak (changes requested):
A few small things only. In regard to the tests, I leave it up to you if you want to adapt them to the new dev guidelines :))

src/modules/fundingManager/depositVault/interfaces/IFM_DepositVault_v1.sol
0xNuggan: I agree. I would move it to IModule_v1, since we may charge fees in different places

0xNuggan:
As MarvinG mentioned, I'd move the fee event to IModule_v1. But maybe we better do that in a separate PR? To keep things self-contained

FHieser:
> As MarvinG mentioned, I'd move the fee event to IModule_v1. But maybe we better do that in a separate PR? To keep things self-contained

In general im still a bit unsure about the implications of changes to overarching contracts and interfaces. It might be a good idea to add this as a section in our dev code of conduct.
In regards of the event moving up, we could even go as far as creating an abstract contract for the the fundingmanager that contains the fundWithdrawal as well as the fee transferal in a single go kinda like the erc20PaymentClient

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
```diff
@@ -57,6 +57,9 @@ contract FM_DepositVault_v1 is
     //--------------------------------------------------------------------------
     // Storage
 
+    /// @dev	Base Points used for percentage calculation. This value represents 100%.
+    uint internal constant BPS = 10_000;
```
FHieser, line 61: Move to end of state

FHieser (commented):
check if event moved up to IModule would change the interfaceId

Zitzak (approved):
LGTM

PR #663, opened 2024-09-15
Feat: Mint wrapper [SC-787]

fabianschu:


src/external/interfaces/IOwnable.sol
fabianschu, line 1: This is necessary because for some reason OZ is missing this interface..

src/external/token/IMintWrapper.sol
fabianschu, line 1: Mostly copied over from IERC20Issuance

0xNuggan:
Looks good to me, but I remember @marvinkruse  mentioning a concern during the call. What was it?

fabianschu:
It was about the sdk compatibility w/ respect to approvals (for decimals there is not a problem since the wrapper exposes the decimals function). The sdk handles approvals for all functinos that require an approval beforehand. In the BC case this is only Buy/Sell, so the auto-approval wouldn't work via the sdk in the current state. BUT it is also not being used for the time being, since the buy/sell happens via a safe transactions. So imo, this is fine. wdyt? @0xNuggan @marvinkruse

0xNuggan:
Yes seems fine to me for now. Can we update and merge this? @marvinkruse

marvinkruse (approved):

PR #665, opened 2024-10-01
[SC-213] Feat: Reserve Vault

FHieser:
Renamed to Token Vault

### DIFF
--- a/src/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.sol
+++ b/src/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.sol
@@ -59,19 +59,11 @@ contract FM_EXT_TokenVault_v1 is IFM_EXT_TokenVault_v1, Module_v1 {
         external
         //@note do we want to keep it that way? Special role?
         onlyOrchestratorAdmin
+        validAddress(tok)
         validAmount(amt)
         validAddress(dst)
     {
-        // if tok == address(0) then send eth //@note do we want to enable sending eth?
-        if (tok == address(0)) {
-            (bool success,) = dst.call{value: amt}("");
-            require(success);
-            emit EthWithdrawn(dst, amt);
-        }
-        // else send tokens
-        else {
-            IERC20(tok).safeTransfer(dst, amt);
-            emit TokensWithdrawn(tok, dst, amt);
-        }
+        IERC20(tok).safeTransfer(dst, amt);
+        emit TokensWithdrawn(tok, dst, amt);
     }
 }
--- a/src/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.sol
+++ b/src/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.sol
@@ -1,5 +1,5 @@
-// SPDX-License-Identifier: MIT
-pragma solidity ^0.8.17;
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity 0.8.23;
 
 // Internal Dependencies
 import {Module_v1} from "src/modules/base/Module_v1.sol";
@@ -13,7 +13,7 @@ import {ERC165Upgradeable} from
     "@oz-up/utils/introspection/ERC165Upgradeable.sol";
 
 /**
- * @title   Token Vault
+ * @title   Inverter Token Vault
  *
  * @notice  Pool holding token Reserves for later use.
  *
@@ -57,6 +57,7 @@ contract FM_EXT_TokenVault_v1 is IFM_EXT_TokenVault_v1, Module_v1 {
     /// @inheritdoc IFM_EXT_TokenVault_v1
     function withdraw(address tok, uint amt, address dst)
         external
+        virtual
         //@note do we want to keep it that way? Special role?
         onlyOrchestratorAdmin
         validAddress(tok)
--- a/src/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.sol
+++ b/src/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.sol
@@ -3,8 +3,8 @@ pragma solidity ^0.8.17;
 
 // Internal Dependencies
 import {Module_v1} from "src/modules/base/Module_v1.sol";
-import {IFM_EXT_ReservePool_v1} from
-    "src/modules/fundingManager/extensions/interfaces/IFM_EXT_ReservePool_v1.sol";
+import {IFM_EXT_TokenVault_v1} from
+    "src/modules/fundingManager/extensions/interfaces/IFM_EXT_TokenVault_v1.sol";
 
 // External Dependencies
 import {IERC20} from "@oz/token/ERC20/IERC20.sol";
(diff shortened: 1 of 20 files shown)


src/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.sol
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @dev     Funds can be withdrawn by OrchestratorAdmin
+ *
+ * @custom:security-contact security@inverter.network
+ *                          In case of any concerns or findings, please refer to our Security Policy
+ *                          at security.inverter.network or email us directly!
+ *
+ * @author  Inverter Network
+ */
+contract FM_EXT_TokenVault_v1 is IFM_EXT_TokenVault_v1, Module_v1 {
```
Zitzak, line 28: Why is this considered a FM and not a LM? It is also not implementing the FM interface as well. I would have assumed this module to be a LM

src/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.sol
```diff
@@ -0,0 +1,69 @@
...
+    // Mutating Functions
+
+    /// @inheritdoc IFM_EXT_TokenVault_v1
+    function withdraw(address tok, uint amt, address dst)
+        external
+        //@note do we want to keep it that way? Special role?
+        onlyOrchestratorAdmin
+        validAddress(tok)
+        validAmount(amt)
+        validAddress(dst)
```
Zitzak: Off topic, but I have created/used a custom modifier (sadly) within the FM_BC contracts :( Didn't know we had this in the Module

src/modules/fundingManager/extensions/interfaces/IFM_EXT_TokenVault_v1.sol
```diff
@@ -0,0 +1,36 @@
...
+    /// @notice Error thrown when the amount is zero.
+    error Module__FM_EXT_TokenVault__InvalidAmount();
+
+    //--------------------------------------------------------------------------
+    // Events
+
+    /// @notice Event emitted when eth is withdrawn.
+    /// @param dst The address to send the eth to.
+    /// @param amount The amount of eth withdrawn.
+    event EthWithdrawn(address indexed dst, uint amount);
```
Zitzak: Not being used

src/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.sol
```diff
@@ -0,0 +1,69 @@
...
+        }
+        _;
+    }
+
+    //--------------------------------------------------------------------------
+    // Mutating Functions
+
+    /// @inheritdoc IFM_EXT_TokenVault_v1
+    function withdraw(address tok, uint amt, address dst)
+        external
```
Zitzak, line 57: Might be smart to make this virtual, as the base will prob be extended in the future to add functionality

src/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.sol
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.17;
```
Zitzak:
```suggestion
pragma solidity 0.8.23;
```

To align with our other contracts

src/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.sol
```diff
@@ -0,0 +1,69 @@
...
+    "src/modules/fundingManager/extensions/interfaces/IFM_EXT_TokenVault_v1.sol";
+
+// External Dependencies
+import {IERC20} from "@oz/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@oz/token/ERC20/utils/SafeERC20.sol";
+import {ERC165Upgradeable} from
+    "@oz-up/utils/introspection/ERC165Upgradeable.sol";
+
+/**
+ * @title   Token Vault
```
Zitzak:
```suggestion
 * @title   Inverter Token Vault
```

test/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.t.sol
```diff
@@ -0,0 +1,132 @@
...
+            );
+        }
+
+        vault.withdraw(address(_token), amount, address(1));
+    }
+
+    //--------------------------------------------------------------------------
+    // Mutating Functions
+
+    function testWithdraw(address token, uint amount, address dst) public {
```
Zitzak: The test is testing multiple conditions, i.e. the success as well as the revert condition

test/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.t.sol
```diff
@@ -0,0 +1,132 @@
...
+
+import {
+    IFundingManager_v1,
+    FundingManagerV1Mock
+} from "test/utils/mocks/modules/FundingManagerV1Mock.sol";
+import {ERC20Mock} from "test/utils/mocks/ERC20Mock.sol";
+// Errors
+import {OZErrors} from "test/utils/errors/OZErrors.sol";
+
+contract FM_EXT_TokenVault_v1Test is ModuleTest {
```
Zitzak, line 39: We don't have any Gherkin in the file.

I've just checked if ChatGPT could help with that and it's pretty good, which could safe you some time. Example:
```
/* Test validAmount() modifier
    ├── Given the vault contains a specific amount of tokens
    │   └── And the amount is equal to 0
    │       └── When the withdraw() function is called
    │           └── Then the transaction should revert with Module__FM_EXT_TokenVault__InvalidAmount error
    └── Given the amount is greater than 0
        └── When the withdraw() function is called
            └── Then the transaction should succeed
*/
```

Zitzak (changes requested):

src/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.sol
FHieser: Because its an extension of a FundingManager
We did the same for the Authorizer Extension
Also for it being a logicmodule it would have to implement the ERC20PaymentClient, which it also doesnt do

test/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.t.sol
FHieser: I test both conditions here
The first one in the test itself
The second in the modifier in position / the modifier test itself

test/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.t.sol
FHieser: See comment below

test/modules/fundingManager/extensions/FM_EXT_TokenVault_v1.t.sol
Zitzak: I will be adding some gherkin to your tests. Will prob do that somewhere later today/tomorrow

Zitzak:
@FHieser Added Gherkin to the test files and changed a few other things based on our dev guidelines.

However somehow I don't get 2 tests to work, but no clue why and have already spent to long on it. Could you take a look at it please 🙏

Zitzak (approved):
LGTM

PR #666, opened 2024-10-01
[SC-809]: Feature: FM_BC_BondingSurface_Redeemable_Repayer_Seizable_v1

Zitzak:


marvinkruse:
Regarding the failing CI test
```
Failing tests:
Encountered 1 failing test in test/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1.t.sol:FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1Test
[FAIL: Module__InvalidAddress(); counterexample: calldata=0xfb2c297c00000000000000000000000003a6a84cd762d9707a21605b548aaab891562aab args=[0x03A6a84cD762D9707A21605b548aaaB891562aAb]] testSetTokenVault_worksGivenAddressIsNotZero(address) (runs: 676, μ: 18886, ~: 18886)

Encountered a total of 1 failing tests, 902 tests succeeded
```

I encountered this one before I started, so I don't think I introduced it, but who knows :D To me it looks like the test just assumes that any calls where `newVault != address(0)` will succeed, but the implementation actually also rejects any calls where `newVault == address(this)`. So I assume the case that occurred here should be exactly that.

<hr>

**Edit**:
Resolved this, should be fine now.

marvinkruse (approved):
Approved to be merged into feature branch, where it will be finalized

PR #668, opened 2024-10-08
[SC-786] Post-Merge Fixes

marvinkruse:
Just noticed two things when checking #662 after it was merged, which is why I opened this PR quickly. It had:
* Compiler Warnings
  * In the main contract, `_token` was shadowing `_token`, so I renamed it according to our new dev guidelines
* Naming of AccessMock
  * File was called `ContractName_AccessMock.sol` instead of `ContractName_Exposed.sol`
  * File was in the wrong folder (not in the same one as the test file)


Zitzak (approved):
LGTM

Zitzak:
I think the PR was actually created before we finished our dev guidelines and due to no additional reviewers, and needing the changes added in the PR to the FM base I moved forward to merging it. Thank you for changing it @marvinkruse

PR #672, opened 2024-10-14
Audit: Custom factory qacc

fabianschu:


src/external/token/MintWrapper.sol
```diff
@@ -0,0 +1,101 @@
...
+
+// External Dependencies
+import {Ownable} from "@oz/access/Ownable.sol";
+
+/**
+ * @title   Inverter Mint Wrapper
+ *
+ * @notice  Used as a wrapper around the {IERC20Issuance_v1} to manage the
+ *          permissioning around minting rights. The additional layer is
+ *          primarily used to avoid standard warnings on popular token trackers
```
omega-audits, line 17: Perhaps it is not our role here to point this out, but this seems like you are writing code to purposely mislead your users, which is not a very nice thing to do.

omega-audits: also, I am kind of wondering if this would work, given that also MintWrapper is ownable..

src/factories/custom/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,365 @@
...
+        bytes32 curveAccess = FM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1(
+            fundingManager
+        ).CURVE_INTERACTION_ROLE();
+        FM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1(fundingManager)
+            .grantModuleRole(curveAccess, beneficiary);
+
+        // assign payment pusher role to beneficiary
+        bytes32 paymentPusherRole =
+            ILM_PC_PaymentRouter_v1(paymentRouter).PAYMENT_PUSHER_ROLE();
+        IModule_v1(paymentRouter).grantModuleRole(
```
omega-audits, line 362: So this logic kind of assumes that:
- there is not more at least one module that supports the paymentRouterInterface (if not, this call will revert)
- there is not more than one module that supports the paymentRouterInterface among hte orchestrator modules (otherwise, it iseems arbitrarty to only take the first one)
- this module supports the IModule_v1 interface (or at least has `grantModuleRole`)

So all this seems to be a bit fragile.. 

The last point is easy to fix by making sure that `ILM_PC_PaymentRouter_v1` exteneds `IModule_v1`. The other two I am not sure - it seems that the right place for this logic is the configuration of the paymetRouter itself.

src/factories/custom/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,365 @@
...
+            decodedConfigParams.initialCollateralSupply,
+            decodedConfigParams.initialIssuanceSupply,
+            decodedConfigParams.realAdmin,
+            beneficiary
+        );
+
+        // assign permissions to buy/sell from curve to admin
+        bytes32 curveAccess = FM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1(
+            fundingManager
+        ).CURVE_INTERACTION_ROLE();
```
omega-audits, line 355: IF the fundingManager does not implement `CURVE_INTERACTION_ROLE` it will revert siltently - it may be a good idea ot check first and revert with a meaningful error message

src/factories/custom/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,365 @@
...
+                )
+            ) {
+                paymentRouter = modules[i];
+            }
+        }
+
+        // enable bonding curve to mint issuance token
+        deployedContracts.mintWrapper.setMinter(fundingManager, true);
+
+        // transfer initial collateral supply from realAdmin to bonding curve and mint issuance token to msg.sender
```
omega-audits: you are minting issuance tokens to `beneficiary`, not to msg.sender

src/factories/custom/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,365 @@
...
+ * @title   Inverter Restricted PIM Factory
+ *
+ * @notice  Used to deploy a PIM workflow with a restricted bonding curve with a
+ *          mechanism to sponsor the required collateral supply and an
+ *          opinionated initial configuration.
+ *
+ * @dev     More user-friendly way to deploy a PIM workflow with an restricted
+ *          bonding curve. Anyone can sponsor the required collateral supply for
+ *          a bonding curve deployment. Initial issuance token supply is minted
+ *          to the deployer. The deployer receives the role to interact with the
```
omega-audits, line 53: you might want to update the docstring, as initial issuance and the curve role is given to the `beneficiary` that is passed in the createWorkflow call

src/factories/custom/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,365 @@
...
+        IBondingCurveBase_v1 fundingManager,
+        IERC20 collateralToken,
+        ERC20Issuance_v1 issuanceToken,
+        uint initialCollateralSupply,
+        uint initialIssuanceSupply,
+        address admin,
+        address beneficiary
+    ) private {
+        uint availableFunding =
+            fundings[admin][beneficiary][address(collateralToken)];
```
omega-audits: Is this as intended? I did not write a test, but it seems to me that:

Anyone can call createPIMWorkflow and can frelly choose `admin` and `beneficiary` and `initialCollateralSupply` arguents and move collateralTokens to their new custom workflow. So anyone can effectively steal the funds.

src/factories/custom/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,365 @@
...
+    ERC2771Context,
+    IRestricted_PIM_Factory_v1
+{
+    using SafeERC20 for IERC20;
+
+    //--------------------------------------------------------------------------
+    // State Variables
+
+    // Stores address of orchestratorfactory.
+    address public orchestratorFactory;
```
omega-audits: Could mark this immutable

src/factories/custom/Restricted_PIM_Factory_v1.sol
```diff
@@ -0,0 +1,365 @@
...
+        address paymentRouter;
+        bytes4 paymentRouterInterfaceId =
+            type(ILM_PC_PaymentRouter_v1).interfaceId;
+        for (uint i; i < modules.length; ++i) {
+            if (
+                ERC165(modules[i]).supportsInterface(
+                    paymentRouterInterfaceId
+                )
+            ) {
+                paymentRouter = modules[i];
```
omega-audits, line 334: Can break here

src/external/token/MintWrapper.sol
fabianschu: I agree, this has been specifically asked for by our client so we'll stick with it for now. The underlying argument with coingecko et al about how to display dynamic tokens is something is a task for later, imo.

src/factories/custom/Restricted_PIM_Factory_v1.sol
fabianschu: You are right, this is fragile. The key purpose of this factory is to smoothen the deployment UX: have the factory do the setup for this specific use case vs requiring users to do many configuration transactions themselves which would be even more fragile. But against the use case, the fragility is ok.

src/factories/custom/Restricted_PIM_Factory_v1.sol
fabianschu: Same as here: https://github.com/InverterNetwork/contracts/pull/672#discussion_r1808532724

Since the motivation of this factory is foremost a highly use case specific UX improvement, we can live with the assumption that user inputs to the factory are correct.

src/factories/custom/Restricted_PIM_Factory_v1.sol
fabianschu: Anyone can execute the deployment for a given admin/beneficiary pair, but the beneficiary will always be the recipient of initial issuance token supply and CURVE_INTERACTION_ROLE on the newly deployed workflow, which in turn receives the initialCollateralSupply. So if Alice chooses to sponsor the bonding curve deployment for Bob, the factory makes sure that if anyone executes the deployment, Alice remains the admin of the workflow and Bob receives initial issuance supply as well as aforementioned role.

So I don't see how funds can be stolen, or am I missing something? :)

I renamed `actor` (which has been relict of previous logic) to `beneficiary` to make this clearer.
@omega-audits

src/factories/custom/Restricted_PIM_Factory_v1.sol
fabianschu: This is fine as per https://github.com/InverterNetwork/contracts/pull/672#discussion_r1808532724

src/factories/custom/Restricted_PIM_Factory_v1.sol
omega-audits: It doesn't matter who is the recipient of the tokens, since anyone can control the entire configs and modules added. For example anyone can deploy with a module transferring all the funds to themselves, or set themselves to control the authorizer etc. The tokens issues are then just worthless because the funds of the organization will be stolen.

omega-audits: So the scenario would be this:

- target of the attack is the collateralToken that is in the factory, and specifically the amount registered here `    fundings[admin][beneficiary][address(collateralToken)];` for some admin, beneficiary 
- attacker calls createPIMWorkflow with arguments a follows:
   - adin adn benificiary as in the target
   - a fundingManager configuration that allows the attacker to access the funds in teh funding manager
   - `initialCollateralSupply` equal to `fundings[admin][beneficiary][address(collateralToken)]`
- this call will transfer initialCollateralSupply collateralToken to the funding manager, where the attacker can take it

I really should write a test for this, so I may be missing some subtieley here

fabianschu: >  - a fundingManager configuration that allows the attacker to access the funds in teh funding manager
> initialCollateralSupply equal to fundings[admin][beneficiary][address(collateralToken)]

Hmm, but isn't is so that for being able to steal funds you need to be the orchestrator admin? And the factory makes sure that the admin is going to be the one who sponsors the whole bonding curve deployment?

Phrased differently: the factory makes sure who is the orchestrator admin of a workflow deployment (namely the address that added the funding in the first place). So only the address that added the funding could steal it. The only other privileges that are handed out through the factory are the CURVE_INTERACTION_ROLE and the PAYMENT_PUSHER_ROLE for the beneficiary. So only the beneficiary could potentially use the PAYMENT_PUSHER_ROLE to drain the reserve, but it is determined by the funder and is represented by a multisig which requires the funder's consent for transactions.

fabianschu: I realize something else though:

While the factory controls who becomes orchestrator admin and who gets the payment pusher role and therefore makes sure that nobody else can steal funds there is a possibility for a malicious person to mess workflows up. So someone could come and deploy a `Simple Payment Processor` instead of a `Streaming Payment Processor` (as is actually intended), so that the workflow would need to be reconfigured by the orchestrator admin to fulfill its original purpose (which involves streaming). 

So I think it makes sense to check in the factory that the expected modules are actually deployed and not some other modules. This is related to one of your [previous comments](https://github.com/InverterNetwork/contracts/pull/672#discussion_r1801118855) I guess.

src/factories/custom/Restricted_PIM_Factory_v1.sol
fabianschu: Additional info: deploying is permissioned now

PR #673, opened 2024-10-15
Audit: fix: assigns payment pusher role to beneficiary

fabianschu:


src/factories/custom/Restricted_PIM_Factory_v1.sol
```diff
@@ -30,6 +32,8 @@ import {MintWrapper} from "src/external/token/MintWrapper.sol";
 
 // External Dependencies
 import {ERC2771Context, Context} from "@oz/metatx/ERC2771Context.sol";
+import {ERC165Upgradeable} from
```
marvinkruse: Any reason we use the `Upgradeable` variant here? We only use it to check for a supported interface, so the version doesn't matter really, but as this contract isn't upgradeable to me it seems like we should use the regular one here too?

src/factories/custom/Restricted_PIM_Factory_v1.sol
```diff
@@ -250,6 +254,20 @@ contract Restricted_PIM_Factory_v1 is
         // get bonding curve / funding manager
         address fundingManager =
             address(deployedContracts.orchestrator.fundingManager());
+        // get payment router
+        address[] memory modules = deployedContracts.orchestrator.listModules();
+        address paymentRouter;
+        bytes4 paymentRouterInterfaceId =
+            type(ILM_PC_PaymentRouter_v1).interfaceId;
+        for (uint i = 0; i < modules.length; i++) {
```
marvinkruse: Slightly gas-optimized way. I know we are a bit all over the place in our codebase but this is the one we usually like to use. 

```suggestion
        for (uint i; i < modules.length; ++i) {
```

(Removed double initialization of `i` and `++i` being slightly cheaper than `i++`)

marvinkruse (commented):
Just two small things

src/factories/custom/Restricted_PIM_Factory_v1.sol
fabianschu: done

0xNuggan (approved):

PR #674, opened 2024-10-15
Payment Processor Templates

Zitzak:
## What has been done?
- Update contribution guidelines with Git Fork Workflow
- Create PP_Template_v1 template and interface
- Create PP_Template_v1 unit test

PR #677, opened 2024-10-21
Fix: Formatting & Dev Guidelines

marvinkruse:
Resolves issues from #674, like formatting and the dev guideline adherence. I just did one pass, so there might still be something.

What is still open?
* There are still multiple functions in the template contract (`processPayments`, `cancelRunningPayments`, etc.) that are not covered in the interface. Either add these to the interface directly or have the interface inherit the `IPaymentProcessor.sol` interface (and subsequently only extend the template itself from `IPP_Template_v1` as this then is an `IPaymentProcessor`).
  * after looking at how the streaming one handles it, probably the second option i listed is how we should do it
* There is no getter for the `_paymentId`, which is not important for the template itself, but it would be a good example of a variable that is definitely needed for the outside (like guardian, etc.) to be accessible, but isn't as it doesn't have a getter.
* We never explicitly spoke about whether variables and internal functions should have a `@dev` or `@notice` tag. Currently, you are using `@dev` tags for these. I've seen both used in other projects. Could make sense to keep it like this, as `@notice` is intended for end-users and internal functions/variables are arguably never seen by an end-user, but I'm not sure if it makes a difference for the doc-generating parts of the system at some point. We can keep it as `@dev` for now though I guess.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -148,7 +148,7 @@ To keep your fork in sync with the `upstream` repository, periodically fetch and
 git fetch upstream
 ```
 
-2.Merge the upstream dev branch into your local dev branch:
+2. Merge the upstream dev branch into your local dev branch:
 
 ```bash
 git checkout dev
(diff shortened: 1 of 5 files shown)


Zitzak (approved):
LGTM

PR #679, opened 2024-10-24
Chore: modifies funding logic

fabianschu:
### Changes: split sponsor (who adds funding) and workflow admin + add deployer allowlist:
* previously the sponsor was equal to the workflow admin => for legal reasons this needs to be changed
* additionally: anyone could use a funding for a deployment, deploying with an arbitrary (undesired) config

=> both is addressed in this PR

---

* new logic: sponsor (who adds funding) specifies 
  * which deployer can deploy a workflow 
  * for which beneficiary (receives initial issuance supply and CURVE_INTERACTION as well as PAYMENT_PUSHER role)
  * with which admin (is the workflow admin)
* `createPIMWorkflow` interface stays unchanged

### DIFF
--- a/src/factories/custom/Restricted_PIM_Factory_v1.sol
+++ b/src/factories/custom/Restricted_PIM_Factory_v1.sol
@@ -71,9 +71,13 @@ contract Restricted_PIM_Factory_v1 is
     address public immutable orchestratorFactory;
     // Stores available fundings.
     mapping(
-        address sponsor
+        address deployer
             => mapping(
-                address beneficiary => mapping(address token => uint amount)
+                address beneficiary
+                    => mapping(
+                        address admin
+                            => mapping(address token => Funding funding)
+                    )
             )
     ) public fundings;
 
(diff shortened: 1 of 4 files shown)


test/factories/custom/Restricted_PIM_Factory_v1.t.sol
```diff
@@ -122,112 +122,336 @@ contract Restricted_PIM_Factory_v1Test is E2ETest {
...
-        |       └── then it grants payment pusher role to beneficiary
-        |       └── then it transfers ownership of mint wrapper to admin
-        |       └── then it renounces ownership of issuance token
-        |       └── then it revokes orchestrator admin rights from factory and transfers them to admin
-        |       └── then it emits a PIMWorkflowCreated event
-        └── given that there is no funding
-        |   └── when called
-        |       └── then it reverts
-    */
+    // /* Test createPIMWorkflow
```
0xNuggan: Why this change?

0xNuggan (changes requested):
Looks good, only the one style thing

test/factories/custom/Restricted_PIM_Factory_v1.t.sol
fabianschu: makes no sense, ty

src/factories/custom/Restricted_PIM_Factory_v1.sol
```diff
@@ -125,37 +131,65 @@ contract Restricted_PIM_Factory_v1 is
...
-    {
+    function addFunding(
+        address deployer,
+        address beneficiary,
+        address admin,
+        address token,
+        uint amount
+    ) external {
+        Funding storage funding = fundings[deployer][beneficiary][admin][token];
+        if (funding.sponsor != address(0) && funding.sponsor != _msgSender()) {
```
ben-kaufman, line 142: This makes it possible to front run calls to this with 1 devimal of of the token and make adding funding fail

omega-audits: This makes it possible to front run calls to this with 1 decimal of of the token and make adding funding fail

PR #667, opened 2024-10-08
Feat: Add Topos Funding Manager + Token Vault

marvinkruse:
Contains the Topos-related contracts, including:
* FM_EXT_TokenVault_v1
* FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1
* FM_BC_BondingSurface_Redeeming_v1
* BondingSurface

Ready to be reviewed by **[Team Omega](https://teamomega.eth.limo/)**, after which it will be merged into `dev`.

PR #680, opened 2024-10-24
Sc 828 address omega review comments

FHieser:
These comments are still unadressed here:
[First](https://github.com/InverterNetwork/contracts/pull/667#discussion_r1808919560) @rororowyourboat

### DIFF
--- a/src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1.sol
+++ b/src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1.sol
@@ -212,8 +212,8 @@ contract FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1 is
     //--------------------------------------------------------------------------
     // Modifiers
 
-    modifier isBuyAndSellRestricted() {
-        _isBuyAndSellRestrictedModifier();
+    modifier checkBuyAndSellRestrictions() {
+        _checkBuyAndSellRestrictionsModifier();
         _;
     }
 
@@ -240,7 +240,7 @@ contract FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1 is
         public
         virtual
         override(BondingCurveBase_v1)
-        isBuyAndSellRestricted
+        checkBuyAndSellRestrictions
     {
         super.buyFor(_receiver, _depositAmount, _minAmountOut);
     }
@@ -268,7 +268,7 @@ contract FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1 is
         public
         virtual
         override(RedeemingBondingCurveBase_v1)
-        isBuyAndSellRestricted
+        checkBuyAndSellRestrictions
     {
         super.sellTo(_receiver, _depositAmount, _minAmountOut);
     }
@@ -541,7 +541,7 @@ contract FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1 is
 
     /// @dev    Validate if buy and sell is restricted, and if so
     ///         check if the caller has the CURVE_INTERACTION_ROLE
-    function _isBuyAndSellRestrictedModifier() internal view {
+    function _checkBuyAndSellRestrictionsModifier() internal view {
         if (buyAndSellIsRestricted) {
             _checkRoleModifier(CURVE_INTERACTION_ROLE, _msgSender());
         }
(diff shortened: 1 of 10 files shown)


src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_v1.sol
```diff
@@ -81,7 +79,7 @@ contract FM_BC_BondingSurface_Redeeming_v1 is
     // Constants
 
     /// @dev Minimum collateral reserve
-    uint public constant MIN_RESERVE = 1 ether;
+    uint public MIN_RESERVE;
```
Zitzak, line 82: Curious if it's still considered a constant as it is being set during initialization? 🤔

src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_v1.sol
```diff
@@ -119,6 +120,9 @@ contract FM_BC_BondingSurface_Redeeming_v1 is
         // Set accepted token
         _token = IERC20(_acceptedToken);
 
+        // MIN_RESERVE is in dependency to the decimals of the workflow token
```
Zitzak:
```suggestion
        // MIN_RESERVE is relational to the decimals of the workflow/colalteral token
```

src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_v1.sol
```diff
@@ -336,8 +287,13 @@ contract FM_BC_BondingSurface_Redeeming_v1 is
         ) {
             revert InvalidOrchestratorTokenWithdrawAmount();
         }
+
         token().safeTransfer(to, amount);
 
+        if (MIN_RESERVE > token().balanceOf(address(this))) {
```
Zitzak, line 293: We could also add the check for MIN_RESERVE into the if statement above, i.e. if (amount > token().balanceOf(address(this)) - projectCollateralFeeCollected + MIN_RESERVE) revert...

src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1.sol
```diff
@@ -164,12 +167,12 @@ contract FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1 is
         // Set accepted token
         _token = IERC20(_acceptedToken);
 
+        // MIN_RESERVE is in dependency to the decimals of the workflow token
```
Zitzak:
```suggestion
    // MIN_RESERVE is relational to the decimals of the workflow/colalteral token  

```

src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1.sol
```diff
@@ -335,6 +336,10 @@ contract FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1 is
         __Module_orchestrator.fundingManager().token().safeTransfer(
             _to, _amount
         );
+        if (MIN_RESERVE > token().balanceOf(address(this))) {
```
Zitzak, line 339: See comment in FM_BC_BondingSurface_Redeeming_v1

Zitzak (changes requested):

src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_v1.sol
FHieser: We decided to keep it here, as it creates more information in case of fail state
In case of the _Redeeming_Restricted_Repayer_Seizable_v1.sol we will adapt it though

src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_v1.sol
FHieser: Yeah thought about that too
Its just in the technicality a constant, at that It wont change anymore after we set it
I think I would keep it as its technically immutable, because we cant modify it without changing the code

src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1.sol
FHieser: Now that I looked into it again I decided against implementing it here, as the _getRepayableAmount doesnt directly interact with the actual amount of tokens that are still in the contract.
I think the way I implemented it now makes more sense, so I would keep it that way

Zitzak (approved):

PR #681, opened 2024-10-29
Refactor: Payment order struct [SC-309]

fabianschu:


fabianschu:
Status Quo:

- [x] new PaymentOrder Struct is applied to all clients and payment processors
- [x] default values (e.g. for streaming PP) are inconsistently defined partly on clients and on processors
- [x] encoding/decoding logic is duplicated in multiple occasions
- [ ] no clear definition of flags "blueprint" anywhere
- [x] no distinct tests that put emphasis on the new decoding/encoding logic
- [x] some refinement of events necessary
- [x] NatSpec, documentation etc.

0xNuggan:
Hey! Thanks for this. Form the driection it looks great! 
One issue I see (and maybe you are aware of it) is that in _decodePaymentOrder() you just take the decoded variables at face value. It may make sense to check against the flags there, and do the "substitute for default value at that point. Also, once we are there, I think nothing speaks against making the PaymentProcessor's decodePaymentOrder a public view function (and maybe even add it to the interface? not sure about that)

fabianschu:
> Also, once we are there, I think nothing speaks against making the PaymentProcessor's decodePaymentOrder a public view function (and maybe even add it to the interface? not sure about that)

I didn't do that because afaiu the return value will be different for every payment processor that will potentially make use of the deconding feature in the future.

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -140,9 +145,16 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
         Metadata memory metadata,
-        bytes memory /*configData*/
+        bytes memory configData
     ) external override(Module_v1) initializer {
         __Module_init(orchestrator_, metadata);
+
+        (uint _defaultStart, uint _defaultCliff, uint _defaultEnd) =
+            abi.decode(configData, (uint, uint, uint));
+
+        defaultStart = _defaultStart;
```
0xNuggan: Maybe create some setters for this, so it can be modified later? Also, we can check for _validTimes()

test/utils/mocks/modules/paymentProcessor/PP_Streaming_v1AccessMock.sol
```diff
@@ -54,4 +54,12 @@ contract PP_Streaming_v1AccessMock is PP_Streaming_v1 {
     {
         return _validPaymentToken(_token);
     }
+
+    function original_getStreamingDetails(bytes16 flags, bytes32[] memory data)
```
0xNuggan: per our brand new repository code of conduct, these functions should be called exposed_XYZ(). We'll need to check the whole repo anyway, but maybe we can make the future intern's life a bit easier :)

src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
```diff
@@ -96,6 +96,9 @@ contract LM_PC_RecurringPayments_v1 is
 
     /// @dev	Marks the beginning of the list.
     uint internal constant _SENTINEL = type(uint).max;
+    /// @dev	Flag config for payment processor indicates presence of start, end and cliff.
+    bytes16 internal constant _START_END_CLIFF_FLAG =
```
0xNuggan: This is of course a quicker way than going with the ">>"'s , but I think we should be consistent across our clients if we go with one or another. Which one do you think is best?

src/modules/paymentProcessor/PP_Streaming_v1.sol
fabianschu: Who should be able to call these setters?

src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
fabianschu: Per se I think it would be fine to use both, if the constant values are explained well (better than they are right now), but now that I look at this specific instance, I think it might make more sense to have the flags assembled "manually" in the init so that if we want to make the recurring payment processor more flexible in the future (enabling it to use different flag configs; change the config for a deployed module etc.) we don't have to touch the contract state. 

So I'm setting the flags manually now. Wdyt?

fabianschu:
After discussing with @FHieser I shifted the streaming pp related flags by one to reserve the first slot in flags for a payment order ID.

Now the second slot is `start`, the third slot is `cliff` and the fourth slot is `end`. For the streaming pp, the flags (if allset) look as follows

```
0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 0000 1110
```

src/modules/logicModule/LM_PC_KPIRewarder_v1.sol
```diff
@@ -136,19 +135,29 @@ contract LM_PC_KPIRewarder_v1 is
     // View functions
 
     /// @inheritdoc ILM_PC_KPIRewarder_v1
-    function getKPI(uint KPInum) public view returns (KPI memory) {
+    function getKPI(uint KPInum) external view returns (KPI memory) {
```
0xNuggan, line 138: Why this change? Is it part of the new CoC?

fabianschu: The spurious dragon came, so I had to do some size optimizations in the context of which I did some CoC changes.

marvinkruse:
Some things popped up for me when I went through the code. I will submit my review in regards to the code itself tomorrow (Wednesday), just wanted to leave my thoughts here so it doesn't delay potential changes.

### 1) The flags in a bytes16
This essentially means that we have 128 different "slots" of information in the data section, as 16 bytes * 8 bits per byte = 128 for the flags. It might sound like a lot, but I feel like we could quickly reach the limit here. If we just look at the example of a payment processor that works with  Sablier, we would already have the options of: 

* Vesting Type (_linear, steps, exponential, etc._)
* Cancellable (_bool_)
* Transferrable (_bool_)
* Segments (_a Sablier feature to define payment steps, each containing information about how much is streamed and when_)
* ...

That's just an example, I can think of a few more protocols that would look similar. So if we would want to give our modules the full capability to make use of all Sablier functions, this would already take up 3-4 slots, depending on what we hard-code or recycle. The payment processors that are developed in the future will probably cover various different use cases (streaming, cross-chain, nfts, shares, conditional, etc.) and each takes up slots, so I can see us reaching that limit rather quickly, personally.

I'd probably just go with bytes32, which doesn't completely remove the problem but at least gives us more room to be safe(r) until we go to v2 in the future and revamp the whole thing potentially. **Or did I miss a specific reason why bytes16 is the preferred choice here?**

### 2) Data being bytes32
More of a question here: is using bytes32 the best choice and would the use of a "normal" bytes array be too big of a complexity overhead? 
My initial reaction is just: we can't even put an IPFS hash into these without splitting it and using two slots (_the older IPFS CIDv0 is 46 bytes, the newer CIDv1 base32 hash 59 bytes_). URLs would also only work if they are quite short. For everything else, this is fine of course, just my initial thought. 

I can see that working with `bytes32` just makes a lot more sense and is a lot easier, as it would otherwise be a two-dimensional bytes-array and probably be a lot more complex to deal with, which might not be worth the benefits. Was just curious - was that your main reasoning? I'd probably keep it like this if my assumption about the complexity is correct, we'll just need to work around the limitations, which might not be a problem after all!

fabianschu:
@marvinkruse Thanks for putting thoughts into this.

Re 1: I agree, I don't see a real advantage here for using bytes32 (not even gas), but maybe I'm overseeing something, @0xNuggan ? If I'm not missing anything I'd suggest to go with your proposition and change this to bytes32.

Re 2: I wasn't really thinking about offchain-related data, therefore using bytes32 with little complexity seemed natural to me, but I agree that for anything that is not meant to be processed by the evm the length restriction might a bit unpractical. My personal feeling is that most params will be used for onchain stuff and those exceptions that don't could be caught with reserving more than one slot for that specific param, but I'm not sure about it. Do you have a preference @marvinkruse ? And also curious what @0xNuggan thinks here.

0xNuggan:
That's a great point, @marvinkruse ! I'm fine with going with bytes32, I don't think there was a specific technical reason for the bytes16 choice. 
For the second point, I'd go with keeping it simple for now (since this is also a "bridge" solution which can be revised for V2 proper). The limitation of reserving two slots for longer parameters doesn't seem that critical to me

0xNuggan:
Ok guys, I think the rest of the implementation is ready. I'd still like to polish the documentation text in IPaymentClientBase, but the rest should be ready for review

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
FHieser, line 1: Wouldnt it be a good idea to make the PaymentRouter as a general "ERC20PaymentClient"?
Doesnt have to happen in this PR, but allowing it create custom PaymentOrders by submitting the data and flag themselves?

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
```diff
@@ -127,4 +155,32 @@ contract LM_PC_PaymentRouter_v1 is
             IERC20PaymentClientBase_v1(address(this))
         );
     }
+
+    /// @dev	Creates the flags and data for a payment order.
+    /// @param  start_ The start date of the streaming period (optional).
+    /// @param  cliff_ The duration of the cliff period.
+    /// @param  end_ The ending of the streaming period.
+    function _assemblePaymentConfig(uint start_, uint cliff_, uint end_)
```
FHieser: You use a function to setup the PaymentConfig here but in the Recurring Payment you dont
Do we want to generalize the assembling into an abstract function maybe?
Just so we agree on a style

src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
```diff
@@ -137,6 +138,16 @@ contract LM_PC_RecurringPayments_v1 is
         }
 
         emit EpochLengthSet(newEpochLength);
+
+        // Set the flags for the PaymentOrders
+        uint8 numOfFlags = 3; // The Module will use 3 flags
```
FHieser: Same for the setting of flags
Is this the way we go forward or do we want to enforce this via an abstract function?

FHieser (commented):
The PaymentOrder stuff looks good
Left some comments of how we want to structure this for future developments

src/modules/paymentProcessor/PP_Simple_v1.sol
FHieser, line 1: Valid PaymentOrder needs to check that the target chain is the same as originchain
This contract doesnt allow for crosschain payments

src/modules/paymentProcessor/PP_Streaming_v1.sol
FHieser, line 1: Valid PaymentOrder needs to check that the target chain is the same as originchain
This contract doesnt allow for crosschain payments

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -86,12 +86,31 @@ abstract contract ERC20PaymentClientBase_v1 is
...
     /// @dev	Storage gap for future upgrades.
-    uint[50] private __gap;
+    uint[48] private __gap;
 
     //--------------------------------------------------------------------------
     // Internal Mutating Functions
 
+    /// @dev	Initializes the staking contract.
+    /// @param  flags_ The flags, represented as an array of uint8 containing the flag IDs between 0 and 255
+    function __ERC20PaymentClientBase_v1_init(uint8[] memory flags_)
```
FHieser, line 105: where test?

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -186,38 +239,50 @@ abstract contract ERC20PaymentClientBase_v1 is
...
             revert Module__ERC20PaymentClientBase__CallerNotAuthorized();
         }
 
         // reduce outstanding token amount by the given amount
-        _outstandingTokenAmounts[token] -= amount;
+        _outstandingTokenAmounts[token_] -= amount_;
+    }
+
+    /// @inheritdoc IERC20PaymentClientBase_v1
+    function getFlags() public view returns (bytes32 flags_) {
```
FHieser, line 282: test?

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -186,38 +239,50 @@ abstract contract ERC20PaymentClientBase_v1 is
...
+        return (_flags);
+    }
+
+    /// @inheritdoc IERC20PaymentClientBase_v1
+    function getAmountOfFlags() public view returns (uint8 numOfFlags_) {
```
FHieser: test?

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -301,4 +366,21 @@ abstract contract ERC20PaymentClientBase_v1 is
     {
         return __Module_orchestrator.paymentProcessor() == who;
     }
+
+    // @dev	Returns the payment configuration from a list of supplied flag values. Can be overriden to add additional validation steps
+    function _assemblePaymentConfig(bytes32[] memory flagValues_)
```
FHieser, line 375: test?

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
```diff
@@ -60,6 +60,24 @@ contract LM_PC_PaymentRouter_v1 is
...
+    // Initializer
+    function init(
+        IOrchestrator_v1 orchestrator_,
+        Metadata memory metadata,
+        bytes memory configData
+    ) external override(Module_v1) initializer {
+        __Module_init(orchestrator_, metadata);
+
+        // Set the flags for the PaymentOrders
+        uint8[] memory flags = new uint8[](3); // The Module will use 3 flags
```
FHieser, line 73: Did you test the init for the Flags?

src/modules/logicModule/LM_PC_Bounties_v1.sol
```diff
@@ -519,14 +520,18 @@ contract LM_PC_Bounties_v1 is ILM_PC_Bounties_v1, ERC20PaymentClientBase_v1 {
         for (uint i; i < length;) {
             contrib = contribs[i];
 
+            (bytes32 flags, bytes32[] memory data) =
+                _assemblePaymentConfig(new bytes32[](0)); // No additional payment data
```
FHieser, line 525: Didnt we agree that we could leave this out?
At least thats the way I remembered it from friday

FHieser (changes requested):

src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
```diff
@@ -297,6 +307,22 @@ contract LM_PC_RecurringPayments_v1 is
...
                 if (epochsNotTriggered > 0) {
+                    // assemble data array
+
+                    bytes32 flags;
+                    bytes32[] memory data;
+
+                    {
+                        bytes32[] memory paymentParameters = new bytes32[](3);
+                        paymentParameters[0] = bytes32(block.timestamp);
+                        paymentParameters[1] = bytes32(0);
```
FHieser: You dont need flag 2 here right?
the parameter is always zero so it should have standard values :thinking:

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
0xNuggan: I mean, both this and getAmountOfFlags are tested in the setFlags() test...

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
FHieser: See comment above

src/modules/logicModule/LM_PC_Bounties_v1.sol
0xNuggan: No, we ended up deciding to do it "properly"

src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
0xNuggan: You are right. Removed it from the contract.

FHieser (approved):

0xNuggan:
Resolved the conflicts, should be mergeable now. For some reason, there were some additional bonding curve tests that were not present in dev. I removed them since they are not relevant for this PR, but it may be good to look if they re-appear once we merge the other open PRs. Here's the link to the change (File is test/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.t.sol): 
https://github.com/InverterNetwork/contracts/pull/681/commits/9e1b87f3b421c8dabf5c1c2fd1cc15b48133ff2d#diff-438b97d4c0c47277e23ea487eca056b9e1f0f6356b537bb8dc868bca94463cfc

src/modules/logicModule/LM_PC_Bounties_v1.sol
```diff
@@ -275,6 +275,7 @@ contract LM_PC_Bounties_v1 is ILM_PC_Bounties_v1, ERC20PaymentClientBase_v1 {
         bytes memory
     ) external override(Module_v1) initializer {
         __Module_init(orchestrator_, metadata);
+        __ERC20PaymentClientBase_v1_init(new uint8[](0)); // This module does not use any PaymentOrder flags
```
marvinkruse: Putting it into one line violates our 80-character limit, so I'd put it before.

```suggestion
        // This module does not use any PaymentOrder flags.
        __ERC20PaymentClientBase_v1_init(new uint8[](0)); 
```

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -121,6 +146,30 @@ abstract contract ERC20PaymentClientBase_v1 is
         }
     }
 
+    /// @dev    Sets the flags for the PaymentOrders.
+    /// @param  numOfFlags_ The number of flags.
+    /// @param  flags_ The flags, represented as an array of uint8 containing the flag IDs between 0 and 255
```
marvinkruse: Line break and dot.
```suggestion
    /// @param  flags_ The flags, represented as an array of uint8 containing 
    ///         the flag IDs between 0 and 255.
```

test/utils/mocks/modules/paymentClient/ERC20PaymentClientBaseV1AccessMock.sol
```diff
@@ -80,4 +86,17 @@ contract ERC20PaymentClientBaseV1AccessMock is ERC20PaymentClientBase_v1 {
     function set_outstandingTokenAmount(address token, uint amount) external {
         _outstandingTokenAmounts[token] = amount;
     }
+
+    function direct_setFlags(uint8 numOfFlags_, uint8[] memory flags_)
```
marvinkruse: Should be `exposed_`, but I haven't checked the rest of this test, if we use `direct` here still we can keep it until we refactor.

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -86,12 +86,31 @@ abstract contract ERC20PaymentClientBase_v1 is
...
+
     /// @dev	Storage gap for future upgrades.
-    uint[50] private __gap;
+    uint[48] private __gap;
 
     //--------------------------------------------------------------------------
     // Internal Mutating Functions
 
+    /// @dev	Initializes the staking contract.
+    /// @param  flags_ The flags, represented as an array of uint8 containing the flag IDs between 0 and 255
```
marvinkruse:
```suggestion
    /// @param  flags_ The flags, represented as an array of uint8 containing
    ///         the flag IDs between 0 and 255.
```

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -121,6 +146,30 @@ abstract contract ERC20PaymentClientBase_v1 is
...
+        }
+
+        _numOfFlags = numOfFlags_;
+
+        _flags = 0;
+        for (uint8 i = 0; i < numOfFlags_; i++) {
+            _flags |= bytes32((1 << flags_[i]));
+        }
+
+        emit FlagsSet(_numOfFlags, _flags);
```
marvinkruse: It should be cheaper to use the `numOfFlags_` calldata value here instead of storage, as that just triggers an `MLOAD` (3 gas) instead of an `SLOAD` (2000 gas for the first read, then 100 for any follow ups).

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
marvinkruse: Naming is a bit inconsistent. The variable is called `_numOfFlags` and then this naming needs to be used throughout. I don't think having a getter that is called slightly different compared to the actual variable is a good idea.

I'd personally probably call the variable itself `_flagCount` as well, as I find that a lot clearer compared to an abbreviation-based "numOfFlags", but yea - let's at least try to be consistent in the naming between variables and functions 🙏

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -301,4 +366,21 @@ abstract contract ERC20PaymentClientBase_v1 is
     {
         return __Module_orchestrator.paymentProcessor() == who;
     }
+
+    // @dev	Returns the payment configuration from a list of supplied flag values. Can be overriden to add additional validation steps
```
marvinkruse: Missing a `/` and wrong line length...

```suggestion
    /// @dev	Returns the payment configuration from a list of supplied flag
    ///         values. Can be overriden to add additional validation steps.
```

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -5,24 +5,64 @@ pragma solidity ^0.8.0;
 import {IPaymentProcessor_v1} from
     "src/modules/paymentProcessor/IPaymentProcessor_v1.sol";
 
+/**
+ * @title   Inverter ERC20 Payment Client Base Interface
```
marvinkruse, line 9: Line length in the whole natspec section here is too long in nearly every line 😅

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -5,24 +5,64 @@ pragma solidity ^0.8.0;
...
+ *
+ *
+ * @author  Inverter Network
+ */
 interface IERC20PaymentClientBase_v1 {
+    //--------------------------------------------------------------------------
+    // MASTER LIST OF PAYMENT ORDER FLAGS
+
+    /*
+    | Flag | Variable type | Name       | Description                              |
```
marvinkruse: The whole box is too wide and the row with the `...` doesn't align at the end.

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -5,24 +5,64 @@ pragma solidity ^0.8.0;
...
+    |------|---------------|------------|------------------------------------------|
+    */
+
     //--------------------------------------------------------------------------
     // Structs
 
     /// @notice Struct used to store information about a payment order.
     /// @param  recipient The recipient of the payment.
-    /// @param  paymentToken The token in which to pay.
+    /// @param  paymentToken The token in which to pay. Assumed to always be on the local chain_id.
```
marvinkruse:
```suggestion
    /// @param  paymentToken The token in which to pay. Assumed to always 
    ///         be on the local chain.
```

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -5,24 +5,64 @@ pragma solidity ^0.8.0;
...
     /// @param  amount The amount of tokens to pay.
-    /// @param  start Timestamp at which the payment should start.
-    /// @param  cliff Duration of the payment cliff.
-    /// @param  end Timestamp at which the payment should be fulfilled.
+    /// @param  originChainId The chain id of the origin.
```
marvinkruse:
```suggestion
    /// @param  originChainId The id of the origin chain.
```

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -5,24 +5,64 @@ pragma solidity ^0.8.0;
...
+    /// @param  targetChainId The chain id on which to find the recipient address.
```
marvinkruse:
```suggestion
    /// @param  targetChainId The id of the target chain.
```

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -5,24 +5,64 @@ pragma solidity ^0.8.0;
...
+    /// @param  flags Flags that contain information on the information contained in the data array.
```
marvinkruse:
```suggestion
    /// @param  flags Flags that indicate which information the data array
    ///         contains.
```

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -5,24 +5,64 @@ pragma solidity ^0.8.0;
...
+    /// @param  data Array of bytes32 that contain information on the payment order.
```
marvinkruse:
```suggestion
    /// @param  data Array of additional data regarding the payment order.
```

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -50,11 +90,13 @@ interface IERC20PaymentClientBase_v1 {
...
-    /// @notice Given end invalid.
-    error Module__ERC20PaymentClientBase__Invalidend();
+    /// @notice Given mismatch between flag count and supplied array length.
+    error Module__ERC20PaymentClientBase__MismatchBetweenFlagCountAndArrayLength(
+        uint8 numOfFlags, uint arrayLength
+    );
 
-    /// @notice Given arrays' length mismatch.
-    error Module__ERC20PaymentClientBase__ArrayLengthMismatch();
+    /// @notice Given number of flags exceeds limit
```
marvinkruse:
```suggestion
    /// @notice Given number of flags exceeds the limit.
```

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -63,10 +105,25 @@ interface IERC20PaymentClientBase_v1 {
     /// @param  recipient The address that will receive the payment.
     /// @param  token The token in which to pay.
     /// @param  amount The amount of tokens the payment consists of.
+    /// @param  originChainId The chain id the order originates from.
+    /// @param  targetChainId The chain id to which the payment order will be sent.
+    /// @param  flags List of flags that are employed by the payment order.
+    /// @param  data Array of bytes32 that contain the values for the payment order's used flags.
```
marvinkruse:
```suggestion
    /// @param  originChainId The id of the origin chain.
    /// @param  targetChainId The id of the target chain.
    /// @param  flags Flags that indicate additional data used by the payment
    ///         order.
    /// @param  data Array of additional data regarding the payment order.
```

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -63,10 +105,25 @@ interface IERC20PaymentClientBase_v1 {
...
+        uint amount,
+        uint originChainId,
+        uint targetChainId,
+        bytes32 flags,
+        bytes32[] data
     );
 
+    /// @notice Emitted when the flags are set.
+    /// @param numOfFlags The number of flags.
+    /// @param newFlags The flags.
```
marvinkruse:
```suggestion
    /// @param  numOfFlags The number of flags set.
    /// @param  newFlags The newly set flags.
```

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -86,12 +86,31 @@ abstract contract ERC20PaymentClientBase_v1 is
...
+    bytes32 internal _flags;
+
     /// @dev	Storage gap for future upgrades.
-    uint[50] private __gap;
+    uint[48] private __gap;
 
     //--------------------------------------------------------------------------
     // Internal Mutating Functions
 
+    /// @dev	Initializes the staking contract.
```
marvinkruse, line 102: Wrong natspec comment here.

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -5,24 +5,64 @@ pragma solidity ^0.8.0;
...
+ * @author  Inverter Network
+ */
 interface IERC20PaymentClientBase_v1 {
+    //--------------------------------------------------------------------------
+    // MASTER LIST OF PAYMENT ORDER FLAGS
+
+    /*
+    | Flag | Variable type | Name       | Description                              |
+    |------|---------------|------------|------------------------------------------|
+    | 0    | bytes32       | orderID    | The ID of the order.                     |
```
marvinkruse: What is the purpose of this id? It can not be used to identify the actual payment order, as payment orders from different payment clients would otherwise be able to use the same order id (as it's set in the payment client, not the processor).

Not really sure what the idea is here?

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -5,24 +5,64 @@ pragma solidity ^0.8.0;
...
+ *
+ * @notice  Enables modules within the Inverter Network to create and manage payment orders
+ *          that can be processed by authorized payment processors, ensuring efficient
+ *          and secure transactions. Refer to the implementations contract for more details.
+ *
+ * @dev     STRUCTURING OF THE FLAGS AND DATA FIELDS
+ *          The PaymentOrder struct implements a flag system to manage the information payloads
+ *          received by the payment processor. It is comprised of a bytes32 value that indicates
+ *          the number of flags that are active, and a bytes32[] value that stores the corresponding values.
+ *          and an array of bytes32[], which stores the corresponding values.
```
marvinkruse: line is double

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -5,24 +5,64 @@ pragma solidity ^0.8.0;
...
+ *
+ *          For example:    If the value of 'flags' is '0xB, that would correspond to a bit representation of '0000 [...] 0000 1011'.
```
marvinkruse: it's the other way around. The actual value is `0000 [...] 0000 1011` (stored as bits) and the hex representation is `0xB`. Also not sure why the hex representation is mentioned here at all?

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -86,12 +86,31 @@ abstract contract ERC20PaymentClientBase_v1 is
     /// @dev	The current cumulative amount of tokens outstanding.
     mapping(address => uint) internal _outstandingTokenAmounts;
 
+    /// @dev    Number of Payment Processor flags.
```
marvinkruse:
```suggestion
    /// @dev    The number of payment processor flags used by this payment
    ///         client.
```

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -86,12 +86,31 @@ abstract contract ERC20PaymentClientBase_v1 is
...
+    uint8 internal _numOfFlags;
+
+    /// @dev    Payment processor flags.
```
marvinkruse:
```suggestion
    /// @dev    The payment processor flags used by this payment client.
```

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -75,26 +132,38 @@ interface IERC20PaymentClientBase_v1 {
...
     /// @dev	Payment Client will reduce the total amount of tokens it will stock up by the given amount.
     /// @dev	This has to be called by a paymentProcessor.
-    /// @param  token The token in which the payment was made.
-    /// @param  amount amount of tokens that have been paid out.
-    function amountPaid(address token, uint amount) external;
+    /// @param  token_ The token in which the payment was made.
+    /// @param  amount_ amount of tokens that have been paid out.
+    function amountPaid(address token_, uint amount_) external;
+
+    /// @notice Returns the number of flags and the flags of the PaymentOrders this client will create.
```
marvinkruse: Wrong description
```suggestion
    /// @notice Returns the flags used when creating payment orders in this
    ///         client.
```

src/modules/paymentProcessor/IPaymentProcessor_v1.sol
```diff
@@ -31,17 +31,19 @@ interface IPaymentProcessor_v1 {
     /// @param  recipient The address that will receive the payment.
     /// @param  paymentToken The address of the token that will be used for the payment.
     /// @param  amount The amount of tokens the payment consists of.
-    /// @param  start Timestamp at which the payment should start being paid out.
-    /// @param  cliff Duration of the cliff period.
-    /// @param  end Timestamp at which the payment should finished being paid out.
+    /// @param  originChainId The chain id of the origin.
+    /// @param  targetChainId The chain id on which to find the recipient address.
+    /// @param  flags Flags that contain information on the information contained in the data array.
+    /// @param  data Array of bytes32 that contain information on the payment order.
```
marvinkruse:
```suggestion
    /// @param  originChainId The id of the origin chain.
    /// @param  targetChainId The id of the target chain.
    /// @param  flags Flags that indicate additional data used by the payment
    ///         order.
    /// @param  data Array of additional data regarding the payment order.
```

src/modules/paymentProcessor/PP_Simple_v1.sol
```diff
@@ -258,4 +262,14 @@ contract PP_Simple_v1 is Module_v1, IPaymentProcessor_v1 {
         );
         return (success && data.length != 0 && _token.code.length != 0);
     }
+
+    function _validOriginAndDestinationChain(
+        uint originChainId_,
+        uint tartgetChainId_
```
marvinkruse:
```suggestion
        uint targetChainId_
```

src/modules/paymentProcessor/PP_Simple_v1.sol
```diff
@@ -258,4 +262,14 @@ contract PP_Simple_v1 is Module_v1, IPaymentProcessor_v1 {
...
+    ) internal view returns (bool) {
+        return (
+            (originChainId_ == tartgetChainId_)
```
marvinkruse:
```suggestion
            (originChainId_ == targetChainId_)
```

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -915,4 +939,51 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
         );
         return (success && data.length != 0 && _token.code.length != 0);
     }
+
+    function _validOriginAndDestinationChain(
+        uint originChainId_,
+        uint tartgetChainId_
```
marvinkruse:
```suggestion
        uint targetChainId_
```

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -915,4 +939,51 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
+    ) internal view returns (bool) {
+        return (
+            (originChainId_ == tartgetChainId_)
```
marvinkruse:
```suggestion
            (originChainId_ == targetChainId_)
```

src/modules/paymentProcessor/PP_Simple_v1.sol
```diff
@@ -258,4 +262,14 @@ contract PP_Simple_v1 is Module_v1, IPaymentProcessor_v1 {
         );
         return (success && data.length != 0 && _token.code.length != 0);
     }
+
+    function _validOriginAndDestinationChain(
```
marvinkruse: Why the mismatch between calling it "destination" here and "target" 2 lines below?

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -915,4 +939,51 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
+        if (hasCliff) dataIdx += 1;
+
+        bool hasEnd = (uint(flags) & (1 << 3)) != 0;
+        end = hasEnd ? uint(data[dataIdx]) : defaultEnd;
+    }
+
+    /// @dev Sets the default start time, cliff and end times for new payment orders
+    /// @param newStart_ The new default start time
+    /// @param newCliff_ The new default cliff time
+    /// @param newEnd_ The new default end time
```
marvinkruse:
```suggestion
    /// @dev    Sets the default start time, cliff and end times for new
    ///         payment orders.
    /// @param  newStart_ The new default start time.
    /// @param  newCliff_ The new default cliff duration.
    /// @param  newEnd_ The new default end time.
```

src/modules/paymentProcessor/interfaces/IPP_Streaming_v1.sol
```diff
@@ -152,6 +132,14 @@ interface IPP_Streaming_v1 is IPaymentProcessor_v1 {
         address paymentClient, address paymentReceiver
     );
 
+    /// @notice the default start, cliff and end times are invalid.
+    /// @param  start The start time.
+    /// @param  cliff The cliff time.
```
marvinkruse:
```suggestion
    /// @notice The default start, cliff and end times are invalid.
    /// @param  start The start time.
    /// @param  cliff The cliff duration.
```

marvinkruse (changes requested):
The changes to the dependencies need to be resolved still, we don't want to update this for no reason :D

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
FHieser: Renamed it to flagCount

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
FHieser: I had a discussion with marvink about this:
The problem this field tries to solve is to (in case it is necessary) make payment orders unique from an outside perspective. The duplication is a problem, where we currently dont want to enforce a concrete solution, as this could be tackled from multiple directions (direct id in the paymentProcessor, hashedId in the PaymentClient). I added a recommondation/Disclaimer going with the latter (the hashid).

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
FHieser: Wont adapt as it breaks style

src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
```diff
@@ -137,6 +138,14 @@ contract LM_PC_RecurringPayments_v1 is
         }
 
         emit EpochLengthSet(newEpochLength);
+
+        // Set the flags for the PaymentOrders
```
marvinkruse, line 142: double comment
```suggestion
```

marvinkruse (approved):

PR #685, opened 2024-11-08
Doc: Funding Manager Template

DogaOztuzun:
Template for Funding Manager has been added.

It contains a simple scenario with deposit.
It contains a simple test scenario for usage of exposed.


src/templates/modules/IFM_Template_v1.sol
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import {IFundingManager_v1} from "@fm/IFundingManager_v1.sol";
+
+/**
+ * @title   Inverter Template Funding Manager
```
FHieser, line 7: Funding Manager Template Interface

src/templates/modules/IFM_Template_v1.sol
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @notice  Basic template funding manager used as base for developing new
```
FHieser: This is the interface for the Funding Manager Template.

src/templates/modules/IFM_Template_v1.sol
```diff
@@ -0,0 +1,67 @@
...
+ *          funding managers.
+ *
+ * @dev     This contract is used to showcase a basic setup for a funding
```
FHieser, line 12: For the Interfaces that were created for Topos I usually removed this section if it didnt have any additional information

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,189 @@
...
+    {
+        return _depositedAmounts[user_];
+    }
+
+    /// @inheritdoc IFundingManager_v1
+    function token() external view override returns (IERC20) {
+        return _orchestratorToken;
+    }
+
+    //--------------------------------------------------------------------------
```
FHieser: The Sectioning is still an open topic in the code of conduct, so this could still change

src/templates/modules/IFM_Template_v1.sol
```diff
@@ -0,0 +1,67 @@
...
+    // Public (Getter)
+
+    /// @notice Returns the deposited balance of a specific address.
+    /// @return user Address of the user.
+    function getDepositedAmount(address user_) external view returns (uint);
+
+    //--------------------------------------------------------------------------
+    // Public (Mutating)
+
+    /// @notice Deposits tokens to the funding manager.
```
FHieser, line 69: missing Param

src/templates/modules/IFM_Template_v1.sol
```diff
@@ -0,0 +1,67 @@
...
+
+    /// @notice Token receiver is not valid.
+    error Module__FM_Template__ReceiverNotValid();
+
+    //--------------------------------------------------------------------------
+    // Public (Getter)
+
+    /// @notice Returns the deposited balance of a specific address.
+    /// @return user Address of the user.
+    function getDepositedAmount(address user_) external view returns (uint);
```
FHieser: same here

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,189 @@
...
+        virtual
+        override(Module_v1)
+        returns (bool)
+    {
+        return interfaceId_ == type(IFM_Template_v1).interfaceId
+            || interfaceId_ == type(IFundingManager_v1).interfaceId
+            || super.supportsInterface(interfaceId_);
+    }
+
+    using SafeERC20 for IERC20;
```
FHieser: Wrong positioning

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,189 @@
...
+    //--------------------------------------------------------------------------
+    // Constants
+
+    // Add constants here
+
+    //--------------------------------------------------------------------------
+    // State
+
+    /// @dev    Mapping of user addresses to their deposited token amounts.
+    mapping(address => uint) private _depositedAmounts;
```
FHieser: internal instead of private

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,189 @@
...
+
+    IERC20 private _orchestratorToken;
```
FHieser: same here

FHieser: Also missing comment
The commenting is still an open topic in the code of conduct, so this could still change

FHieser (changes requested):
All functions in contracts should be virtual

src/templates/modules/IFM_Template_v1.sol
DogaOztuzun: Updated

src/templates/modules/IFM_Template_v1.sol
DogaOztuzun: Updated

src/templates/modules/IFM_Template_v1.sol
DogaOztuzun: I kept them as its.

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,192 @@
...
+pragma solidity 0.8.23;
+
+// Internal
+import {IOrchestrator_v1} from
+    "src/orchestrator/interfaces/IOrchestrator_v1.sol";
+import {ERC165Upgradeable, Module_v1} from "src/modules/base/Module_v1.sol";
+import {IFM_Template_v1} from "./IFM_Template_v1.sol";
+import {IFundingManager_v1} from "@fm/IFundingManager_v1.sol";
+
+// External Libraries
```
marvinkruse:
```suggestion
// External
```

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,192 @@
...
+    //--------------------------------------------------------------------------
+    // Constants
+
+    // Add constants here
+
+    //--------------------------------------------------------------------------
+    // State
+
+    /// @dev    Mapping of user addresses to their deposited token amounts.
+    mapping(address => uint) internal _depositedAmounts;
```
marvinkruse: We want to name mappings so it's immediately clear what a mapping does internally. For example here that could be:

```suggestion
    mapping(address user => uint amount) internal _depositedAmounts;
```

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,192 @@
...
+
+    /// @dev    The orchestrator token.
+    IERC20 internal _orchestratorToken;
+
```
marvinkruse, line 78: Missing gap for future upgrades, like so:

```suggestion

    /// @dev    Storage gap for future upgrades.
    uint[50] private __gap;
    
```

src/templates/modules/FM_Template_v1.sol
DogaOztuzun: Updated to equals

src/templates/modules/FM_Template_v1.sol
DogaOztuzun: updated with comment

src/templates/modules/IFM_Template_v1.sol
marvinkruse: Also: Currently it's saying "user" is the return value, while that's the input. So the docs are just mixed up. Also the return value is not named here, but should be called `user_`, so all in all this should be something like:
```
    /// @notice Returns the deposited balance of a specific address.
    /// @param  user_ Address of the user.
    /// @return amount_ The amount that the user deposited.
    function getDepositedAmount(address user_) external view returns (uint amount_);
```

src/templates/modules/FM_Template_v1.sol
DogaOztuzun: Updated

src/templates/modules/FM_Template_v1.sol
DogaOztuzun: Updated

src/templates/modules/FM_Template_v1.sol
DogaOztuzun: Updated

src/templates/modules/IFM_Template_v1.sol
FHieser: Still needs to be updated

FHieser (changes requested):
One comment still missing

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,196 @@
...
+ *                          In case of any concerns or findings, please refer
+ *                          to our Security Policy at security.inverter.network
+ *                          or email us directly!
+ *
+ * @custom:version 1.0.0
+ *
+ * @author  Inverter Network
+ */
+contract FM_Template_v1 is IFM_Template_v1, Module_v1 {
+    //==========================================================================
```
marvinkruse: These separators need a space between the `//` and the `===` (applies to all of them). Still keeping it at 80 characters total length though with all whitespace etc. counted.

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,196 @@
...
+
+    //==========================================================================
+    // Constants
+
+    // Add constants here
+
+    //==========================================================================
+    // State
+
+    /// @dev    Mapping of user addresses to their deposited token amounts.
```
marvinkruse: This should be `@notice`, as it's not developer-specific information.

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,196 @@
...
+    mapping(address user => uint amount) internal _depositedAmounts;
+
+    /// @dev    The orchestrator token.
```
marvinkruse: This should be `@notice`, as it's not developer-specific information.

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,196 @@
...
+
+    //==========================================================================
+    // Public (Getters)
+
+    /// @inheritdoc IFM_Template_v1
+    function getDepositedAmount(address user_)
+        external
+        view
+        virtual
+        override
```
marvinkruse: Are you overriding anything here? Same for all other functions: override shouldn't be used unless we are overriding anything. The important part is the virtual keyword.

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,196 @@
...
+        returns (uint amount_)
+    {
+        uint amount_ = _depositedAmounts[user_];
+        return amount_;
+    }
+
+    /// @inheritdoc IFundingManager_v1
+    function token() external view override returns (IERC20) {
```
marvinkruse: Unnamed return value

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,196 @@
...
+
+    /// @inheritdoc IFM_Template_v1
+    function deposit(uint amount_) external virtual override {
+        // Validate parameters.
+        if (amount_ == 0) {
+            revert Module__FM_Template_InvalidAmount();
+        }
+
+        // Emit event.
+        emit Deposited(_msgSender(), amount_);
```
marvinkruse: We usually want to have events at the end, unless there is a reason (i.e. events that want to reference before- and after-values). But here, I'd move it to the very end.

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,196 @@
...
+        _depositedAmounts[_msgSender()] += amount_;
+
+        // Transfer tokens.
+        _orchestratorToken.safeTransferFrom(
+            _msgSender(), address(this), amount_
+        );
+    }
+
+    /// @inheritdoc IFundingManager_v1
+    function transferOrchestratorToken(address to, uint amount)
```
marvinkruse, line 154: Input parameters should be `to_` and `amount_` (underscores).

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,196 @@
...
+        external
+        virtual
+        override
+        onlyPaymentClient
+    {
+        // Only the payment client can call this function.
```
marvinkruse: This would be a `@dev` note in then Natspec, not just a regular comment somewhere in the code.

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,196 @@
...
+
+        // Validate parameters.
+        _validateOrchestratorTokenTransfer(to, amount);
+
+        // Emit event.
+        emit TransferOrchestratorToken(to, amount);
```
marvinkruse: Same with the event here: move to end.

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,196 @@
...
+        // This value is an example value used to showcase the setters/getters
+        // and internal functions/state formatting style.
+        (address orchestratorTokenAddress) = abi.decode(configData_, (address));
+
+        // Set init state.
+        _orchestratorToken = IERC20(orchestratorTokenAddress);
+    }
+
+    //==========================================================================
+    // Public (Getters)
```
marvinkruse: In the dev guidelines we styled this as "Public - Getters"

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,196 @@
...
+        return amount_;
+    }
+
+    /// @inheritdoc IFundingManager_v1
+    function token() external view override returns (IERC20) {
+        return _orchestratorToken;
+    }
+
+    //==========================================================================
+    // Public (Mutating)
```
marvinkruse: `Public - Mutating`

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,196 @@
...
+        emit TransferOrchestratorToken(to, amount);
+
+        // Transfer tokens.
+        _orchestratorToken.safeTransfer(to, amount);
+    }
+
+    //==========================================================================
+    // Internal
+
+    // Add internal functions here
```
marvinkruse: Probably just a leftover, as we do have an internal function here this comment doesn't need to be here.

src/templates/modules/FM_Template_v1.sol
```diff
@@ -0,0 +1,196 @@
...
+            revert Module__FM_Template_InvalidAmount();
+        }
+
+        if (_depositedAmounts[_msgSender()] < amount_) {
+            revert Module__FM_Template_InvalidAmount();
+        }
+    }
+
+    //==========================================================================
+    // Internal override
```
marvinkruse: Maybe something like "Overridden Internal Functions" would sound a bit better?

src/templates/modules/IFM_Template_v1.sol
```diff
@@ -0,0 +1,72 @@
...
+    // Errors
+
+    /// @notice Amount can not be zero.
+    error Module__FM_Template_InvalidAmount();
+
+    /// @notice Token receiver is not valid.
+    error Module__FM_Template__ReceiverNotValid();
+
+    //==========================================================================
+    // Public (Getter)
```
marvinkruse: Public - Getters

src/templates/modules/IFM_Template_v1.sol
```diff
@@ -0,0 +1,72 @@
...
+    /// @notice Returns the deposited balance of a specific address.
+    /// @param  user_ The address of the user.
+    /// @return amount_ Deposited amount of the user.
+    function getDepositedAmount(address user_)
+        external
+        view
+        returns (uint amount_);
+
+    //==========================================================================
+    // Public (Mutating)
```
marvinkruse: Public - Mutating

src/templates/tests/unit/FM_Template_v1_Exposed.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import {FM_Template_v1} from "src/templates/modules/FM_Template_v1.sol";
+
+// Access Mock of the FM_Template_v1 contract for Testing.
+contract FM_Template_v1_Exposed is FM_Template_v1 {
+    // Use the `exposed_` prefix for functions to expose internal contract for testing purposes only.
```
marvinkruse:
```suggestion
    // Use the `exposed_` prefix for functions to expose internal functions for testing purposes only.
```

marvinkruse (changes requested):
Some more changes sir 🙏

src/templates/modules/IFM_Template_v1.sol
marvinkruse: I disagree here @FHieser (also tagging @DogaOztuzun), it's an interface but the title is still the normal one "Inverter Template Funding Manager". It's not a best practice to name it as "Interface" here.

src/templates/modules/IFM_Template_v1.sol
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import {IFundingManager_v1} from "@fm/IFundingManager_v1.sol";
+
+/**
+ * @title   Inverter Template Funding Manager Inteface
```
marvinkruse: See my comment here:
https://github.com/InverterNetwork/contracts/pull/685#discussion_r1834442553

src/templates/modules/FM_Template_v1.sol
DogaOztuzun: It's coming as its from the Interface. I don't want to touch there atm.

FHieser (approved):

marvinkruse (approved):

PR #686, opened 2024-11-08
Crosschain PP Contract addition - DRAFT

leeftk:
# Connext Bridge Implementation & Testing Structure

## Overview
This PR implements the core structure for the Connext Bridge payment processor, including inheritance patterns, base contract functionality, and initial test coverage.

## Key Changes
- ✅ Established contract inheritance structure
- ✅ Implemented `executeBridge` function in base contract
- ✅ Created exposed version of `executeBridge` for testing purposes
- ✅ Separated bridge logic into dedicated `Connext_Logic` contract
- ✅ Added mock contracts for Connext bridge testing
- ✅ Implemented base contract tests (all passing)

### Contract Structure
- Base contract contains core bridge functionality
- Added interface for Base contracts
- Exposed contract exposed executeBridge function for testing
- Created `PP_Connext_Crosschain_v1` `IPP_Connext_Crosschain_v1`  and `Connext_Logic`
- Logic for bridge separated into `Connext_Logic` contract for better maintainability
- `PP_Connext_Crosschain_v1` implements `executeBridge` using the separated logic imported from`Connext_Logic`

### Testing Infrastructure
- Added mocks mocking the Everclear(Connext) bridge
- Added interfaces and exposed them
- Added test file for`PP_Connext_Crosschain_v1`, currently no tests in there but everything compiles
- Base tests implemented and passing(need to expand on these these are just testing init atm)

## Current Challenges
🚧 Setting up tests for `PP_Connext_Bridge_v1` contract
- Currently investigating optimal testing approach
- May require architectural discussion for test structure

## TODO - Delivery 11/13
- [ ] Add comprehensive comments to test implementations
- [ ] Refactor and restructure the code
- [ ] Expand test coverage for edge cases
- [ ] Resolve `PP_Connext_Crosschain_v1` testing challenges
- [ ] Add documentation for test helpers and mocks
- [ ] Add more tests for base add tests for PP_Connext

## Next Steps
1. Continue investigation of `PP_Connext_Crosschain_v1` testing approach
2. Schedule discussion if current testing challenges persist
3. Implement additional test cases once base structure is confirmed

Helpful feedback:

- Setting up the test for our logic PP_Connext_Crosschain_v1t.sol
- Are the inheritance patterns correct?
- Is the setup for the CrosschainBase_v1.t.sol correct?
- What else should we be testing for the base contract besides that it returns empty bytes?
- Does our file structure(tests included) make sense the way it is? Any changes?

*Note: Moving forward I suspect the PRs will be a lot smaller than this so that its easier to give feedback and avoid the waterfall method. This one ended up being this long due to adding the process in this week and us already having written some of this prior to that.

P.S. The code needs cleaning and refactoring, for now, we wanted to get the inheritance pattern correct and set up our contracts. Zuhaib has a separate branch we're he's already fixing a lot of these things but we wanted to keep that separate to make the feedback easier.


Zitzak:
I changed the branch to merge into to `dev` @leeftk

Zitzak:
Also could you set the PR to be a draft PR. Assuming here that it is being opened for feedback more than actual merging yet @leeftk

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
```diff
@@ -0,0 +1,94 @@
...
+        returns (bool)
+    {
+        return interfaceId_ == type(ICrossChainBase_v1).interfaceId
+            || super.supportsInterface(interfaceId_);
+    }
+    //--------------------------------------------------------------------------
+    // State
+
+    /// @dev Mapping of payment ID to bridge transfer return data
+    mapping(uint => bytes) internal _bridgeData;
```
Zitzak: How do we make sure that the implementation contract, inheriting from the base, is using this `bytes` storage? Not sure about this, but we could make it private and enforce a setter for it (maybe). Similar as what is done with ERC20 abstracts and the token balance

Zitzak:
@leeftk It might be a good idea if we plan a call for you to lead me, and @FHieser through the code, maybe as a sort of onboarding.

Also I've noticed that there are a lot of template elements in the code still. As I think this is what you meant with refactoring the code I did not comment on them :)

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
Zitzak: Also curious if you guys looked into common return values from bridge protocol to conclude this to be the right data to be stored.

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
```diff
@@ -0,0 +1,94 @@
...
+
+    /// @notice Execute the cross-chain bridge transfer
+    /// @dev Override this function to implement specific bridge logic
+    /// @param order The payment order containing all necessary transfer details
+    /// @return bridgeData Arbitrary data returned by the bridge implementation
+    function _executeBridgeTransfer(
+        IERC20PaymentClientBase_v1.PaymentOrder memory order,
+        bytes memory executionData
+    ) internal virtual returns (bytes memory) {
+        emit BridgeTransferExecuted(executionData);
```
Zitzak: Curious why did you guys go with implementing a function here, instead of just defining it's interface

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
```diff
@@ -0,0 +1,94 @@
...
+ *          - Use of the ERC165Upgradeable contract to check for interface support.
+ *
+ * @custom:security-contact security@inverter.network
+ *                          In case of any concerns or findings, please refer to our Security Policy
+ *                          at security.inverter.network or email us directly!
+ *
+ * @author  Inverter Network
+ */
+
+contract CrosschainBase_v1 is ICrossChainBase_v1, Module_v1 {
```
Zitzak: Would assume this to be an abstract contract, as we don't want this contract to work as a standalone

src/modules/paymentProcessor/bridging/PP_Connext_Crosschain_v1.sol
```diff
@@ -0,0 +1,58 @@
...
+import {ICrossChainBase_v1} from
+    "src/modules/paymentProcessor/bridging/abstracts/ICrosschainBase_v1.sol";
+import {ConnextBridgeLogic} from "./ConnextBridgeLogic.sol";
+import {IERC20PaymentClientBase_v1} from
+    "@lm/interfaces/IERC20PaymentClientBase_v1.sol";
+
+contract PP_Connext_Crosschain_v1 is CrosschainBase_v1 {
+    ConnextBridgeLogic public connextBridgeLogic;
+
+    constructor(uint chainId_, address connextBridgeLogic_)
```
Zitzak:
```suggestion
 /// @inheritdoc Module_v1
    function init(
        IOrchestrator_v1 orchestrator_,
        Metadata memory metadata,
        bytes memory configData
    ) external override(Module_v1) initializer {
        __Module_init(orchestrator_, metadata);

        (uint chainId_, address connextBridgeLogic_) = abi.decode(configData, (uint, address));
        
        // set chainId
        
          connextBridgeLogic = ConnextBridgeLogic(connextBridgeLogic_);
    }

```

Zitzak: The modules are upgradeable so they should work through initialization. Also each module gets its parameters through the `bytes configdata` in the init. So in this case the `Module_v1` init is overridden as it is inherited in the `CrosschainBase_v1`

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
```diff
@@ -0,0 +1,94 @@
...
+    bytes public executionData;
+
+    /// @dev    Payout amount multiplier.
+    uint internal _payoutAmountMultiplier;
+
+    /// @dev    The number of payment orders.
+    uint internal _paymentId;
+
+    /// @dev    The chain id of the current chain.
+    uint internal _chainId;
```
Zitzak: We can get the current chainID from the [block](https://docs.soliditylang.org/en/latest/units-and-global-variables.html#block-and-transaction-properties). Not sure however if some versions of ETH forks might not provide this value.

Maybe a setter could get called which could check `block.chainid` and does a revert if the value is invalid. This function could then be called from the implementation during init

src/modules/paymentProcessor/bridging/ConnextBridgeLogic.sol
```diff
@@ -0,0 +1,67 @@
...
+
+interface IEverclearSpoke {
+    function newIntent(
+        uint32[] memory destinations,
+        address to,
+        address inputAsset,
+        address outputAsset,
+        uint amount,
+        uint24 maxFee,
+        uint48 ttl,
```
Zitzak: What does ttl stand for :D

src/modules/paymentProcessor/bridging/ConnextBridgeLogic.sol
```diff
@@ -0,0 +1,67 @@
...
+        everClearSpoke = IEverclearSpoke(_everclearSpoke);
+        weth = _weth;
+    }
+
+    function xcall(
+        IERC20PaymentClientBase_v1.PaymentOrder memory order,
+        bytes memory executionData
+    ) external payable returns (bytes32 intentId) {
+        // Decode any additional parameters from executionData
+        (uint maxFee, uint ttl) = abi.decode(executionData, (uint, uint));
```
Zitzak: Do we need to validate if the values are valid, i.e. are we able to find out if something would revert on the bridge side given something went wrong with the value configuration?

src/modules/paymentProcessor/bridging/ConnextBridgeLogic.sol
```diff
@@ -0,0 +1,67 @@
...
+        address inputAsset,
+        address outputAsset,
+        uint amount,
+        uint24 maxFee,
+        uint48 ttl,
+        bytes memory data
+    ) external returns (bytes32 intentId, uint amountOut);
+}
+
+contract ConnextBridgeLogic is CrosschainBase_v1 {
```
Zitzak: Why did you went the route of inheriting from the crosschain base? This would make this contract a module (as the base inherits from `Module_v1`?

It seems to me that this logic is very Connext/Everclear specific so it could be added to the Connext Implementation

Zitzak: Just went back to the diagram to understand if this is something you outlined in it.

My assumption was that the `ConnextBridgeLogic` is something provided by Connext and not a contract we create ourselves

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
leeftk: Different bridges return different types of data:
- Hyperlane returns bytes32 (messageId)
- Connext returns uint (intentID)
- LayerZero returns nothing

The `_bridgeData` mapping isn't actually necessary because:
1. The return values are primarily used for tracking/monitoring on the source chain
2. Each bridge implementation can handle its return data differently based on its specific needs
3. The destination chain execution isn't dependent on these return values

Regarding enforcing usage through a setter - since the return data isn't critical for cross-chain functionality, we can leave it up to the implementing contracts to decide if and how they want to handle bridge return data. This gives more flexibility to handle the varying return types from different bridge protocols. So for now I'm thinking we remove this.

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
leeftk: This is wrong, I uncommented this yesterday while formatting the code and realized I never changed it back. Will change this and push it up. Nice catch.

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
leeftk: Yes this is correct, we changed this in our newest PR but will also adjust here.

src/modules/paymentProcessor/bridging/PP_Connext_Crosschain_v1.sol
leeftk: Ahhh okay this is nice catch as well, didn't realize this but yes we should be using init and not the constructor. @zzzuhaibmohd please see this as you are making changes in the other PR and this is relevant.

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
leeftk: This makes sense we don't need this.

src/modules/paymentProcessor/bridging/ConnextBridgeLogic.sol
leeftk: Time to live aka a deadline check for a solver to fill an intent. This helps avoid slippage attacks with solvers filling intents that may be stale or intentionally aren't filled to get a better price for themselves.

src/modules/paymentProcessor/bridging/ConnextBridgeLogic.sol
leeftk: Yes 100% we do, we need to check that ttl is not zero and the fee isn't greater than max fee. Nice point out, ideally we're also testing for this and the unhappy paths.

src/modules/paymentProcessor/bridging/ConnextBridgeLogic.sol
leeftk: I think this was an oversight from reading the diagram. You're right - inheriting from CrosschainBase_v1 was incorrect as it would make ConnextBridgeLogic a module which is not the goal, this should be solely a contract containing the logic. However my question is maybe the logic should go in PP_CrossChain_V1 instead?

This bridge-specific logic should instead live in PP_Connext_Bridge implementation since:
1. It's Connext-specific functionality
2. It shouldn't be a separate module
3. It follows our architecture pattern better where bridge implementations handle their own specific logic

I'll move this logic into PP_Connext_Bridge and remove the CrosschainBase_v1 inheritance. Thanks for catching this - think I'm finally starting to catch on with regard to the inheritance patterns but still digesting.

leeftk:
I think next steps here is updating some of the obvious things that were mentioned by @Zitzak. I'm going to push those changes to this branch. Moving forward the PRs will be smaller and a bit more clear with what needs review. Please bear with us as we're trying to figure this out on the fly and learning as we go. Thanks for the review I'll push up the changes and lyk when it's ready for another look.

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
leeftk: This is addressed in the new commits added.

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
leeftk: Addressed in new commits

src/modules/paymentProcessor/bridging/ConnextBridgeLogic.sol
leeftk: Addressed in new commits this contract was removed fully after the new UML diagram was sent over. Please verify now Inheritance chain is correct.

leeftk:
## New Changes Overview

I've addressed most of the requested changes in this PR, with a few items still pending. Key updates include:

## Architecture Changes
- Restructured the inheritance chain to better align with the provided UML diagram
- Separated concerns by creating distinct contracts:
  - Created a base contract for common crosschain functionality
  - Established a separate contract for crosschain payment processor
  - Removed the standalone Connext bridge logic contract
  - Integrated Connext bridge logic directly into PP_Crosschain

## Design Improvements
- Enhanced abstraction of the crosschain base functionality, following the diagram specifications
- Simplified the overall contract structure

## Next Steps
- Please review the inheritance structure against the UML diagram
- A few minor items are still pending and will be addressed in follow-up PRs

Let's focus on the inheritance chain for now and ensure everything is where it's supposed to be. I commented out all the tests so we can make things a bit cleaner as we have tests coming in a different branch that Zuhaib is already working on. Let me know if you have any questions.

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
```diff
@@ -0,0 +1,88 @@
...
+
+    /// @notice Execute the cross-chain bridge transfer
+    /// @dev Override this function to implement specific bridge logic
+    /// @param order The payment order containing all necessary transfer details
+    /// @return bridgeData Arbitrary data returned by the bridge implementation
+    function _executeBridgeTransfer(
+        IERC20PaymentClientBase_v1.PaymentOrder memory order,
+        bytes memory executionData
+    ) internal virtual returns (bytes memory) {
+        return bytes("");
```
FHieser: function has an implementation, but doesnt need it here.
If you remove it and replace the {...} section of the function with a semicolon ; then we enforce the function implementation downstream.

src/modules/paymentProcessor/bridging/PP_Connext_Crosschain_v1.sol
```diff
@@ -0,0 +1,118 @@
...
+    "src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol";
+import {ICrossChainBase_v1} from
+    "src/modules/paymentProcessor/bridging/abstracts/ICrosschainBase_v1.sol";
+import {IERC20PaymentClientBase_v1} from
+    "@lm/interfaces/IERC20PaymentClientBase_v1.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {PP_Crosschain_v1} from
+    "src/modules/paymentProcessor/bridging/abstracts/PP_Crosschain_v1.sol";
+
+interface IEverclearSpoke {
```
FHieser: Please move these interfaces into separate files and put them into the interface folder

Zitzak:
The inheritance chain looks good as it currently is @leeftk. I just reviewed it together with @FHieser, as he will be supporting your development efforts in the coming week.

One thing which I think would be good to do at this point is to move the files into the correct folders.
- `src->modules->pp->abstracts = ` all of the abstracts you created, i.e. the CrosschainBase & PP_Crosschain
- `src->modules->pp->interfaces = ` all of the interfaces created. This includes the external interfaces as @FHieser commented
- `src->modules->pp = ` the implementation can stay in the root of the paymentProcessor directory.

Additionally you can remove the directory `bridging`

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    function getChainId() external view virtual returns (uint) {
+        return _chainId;
+    }
+
+    /// @notice Process payments for a given payment client
+    /// @param client The payment client to process payments for
+    function processPayments(IERC20PaymentClientBase_v1 client)
```
FHieser: This function shouldnt be part of this contract

leeftk: Okay, I removed this but am now curious if this is also true for the state variables and events. Alternatively, what should go in this contract, just `_executeBridgeTransfer`?

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
leeftk: Fixed in new commit

src/modules/paymentProcessor/bridging/ConnextBridgeLogic.sol
leeftk: Addressed in new PR.

src/modules/paymentProcessor/bridging/PP_Connext_Crosschain_v1.sol
leeftk: Resolved this in the new changes.

src/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.sol
FHieser: Yeah thats a good point
I would say that the state in this case can be removed too
The events still fit the purpose and ideally should be used by inherited contract,
Although i am currently not sure if they should be enforced.
Something we could do here, would be to already write the modifiers, that fit to these events,
like a clientIsValid modifier for example.
In general the question to ask is:
Does this functionality fit for the purpose of **this** contract?

leeftk: hmmm okay something to ponder, give a bit. We may need to spar a bit for me to understand this conceptually but I think I grasp the idea. Will make changes and send them over in a bit.

src/modules/paymentProcessor/bridging/PP_Connext_Crosschain_v1.sol
leeftk: @FHieser I need some help on understanding this `PP_Connext_Crosschain_v1.sol` inherits 'PP_Crosschain' which then inherits 'CrosschainBase_v1'. How do the initializers work here especially when dealing with this nested inheritance pattern?

FHieser: Yes that is a really good question. Took me a bit to solve it too to solve it in this case.
First what you have to do to solve it:
CrosschainBase -> Remove init
PP_CrossChain -> Remove init
PP_Connext_Crosschain -> Use `override(Module_v1) ` (You'll have to import it)

FHieser: Now to the explanation, starting with the bad part:
The initialization concept we currently use straight up wasnt properly designed for multi contract inheritance 😅
The initial thought process that one might follow is, that we have to pass along the configdata through the inheritance tree until it lands at the correct place. But as we dont have a standard defined here, of how this would actually looks like, this will get confusing very fast. We initially solved this by creating a different __Module_init, that can be called separately, but this solution is probably not ideal too. So essentially, you stumbled upon something that we didnt define properly, which we will do.
Now to the good part:
We are lucky in this case, that only the Connext_Crosschain actually needs to set data during initialization, which the solution that I posted above basically uses. We just say ignore all the abstract contracts, as they dont need any initialization data in the first place. 
Nice to get lucky I guess, but this isnt feasable in the long run. I will get together with my colleagues, to get a proper concept together, hopefully this week.

leeftk: Awesome thanks! Well glad that us running into these things is helping you guys on the other end. I'm curious if the other team will also run into a similar issue. Will start a discussion in the thread around initialization and we can discuss it there as well. Thanks for the response here. Also Zuhaib fixed this so Im resolving here.

leeftk:
Resolving the conflicts and then merging this to update the feature branch. I know we still have one item pending with structure but lets leave that for the tests PR so that we can get this one closed.

src/modules/paymentProcessor/abstracts/CrosschainBase_v1.sol
```diff
@@ -0,0 +1,62 @@
...
+    // Virtual Functions
+
+    /// @notice Execute the cross-chain bridge transfer
+    /// @dev Override this function to implement specific bridge logic
+    /// @param order The payment order containing all necessary transfer details
+    /// @return bridgeData Arbitrary data returned by the bridge implementation
+    function _executeBridgeTransfer(
+        IERC20PaymentClientBase_v1.PaymentOrder memory order,
+        bytes memory executionData
+    ) internal virtual returns (bytes memory) {}
```
Zitzak: Function should only be the interface still. Might be that you guys left it to focus on the structure onle

test/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1.t.sol
```diff
@@ -0,0 +1,86 @@
+//SPDX-License-Identifier: LGPL-3.0-only
```
Zitzak: Contract should live at `test/modules/paymentProcessor/abstracts/CrosschainBase_v1.t.sol`, same location as for the src folder

test/modules/paymentProcessor/bridging/PP_Connext_Bridge.t.sol
```diff
@@ -0,0 +1,73 @@
+//SPDX-License-Identifier: LGPL-3.0-only
```
Zitzak: Contract should live `test/modules/paymentProcessor/PP_Connext_Bridge.t.sol` same as the src folder

test/modules/paymentProcessor/bridging/abstracts/CrosschainBase_v1_Exposed.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
Zitzak: Should be in `test/modules/paymentProcessor/abstracts/CrosschainBase_v1_Exposed.sol` same folder as the test file

test/modules/paymentProcessor/bridging/abstracts/Interfaces/IEverClearSpoke.sol
```diff
@@ -0,0 +1,116 @@
+// SPDX-License-Identifier: MIT
```
Zitzak: SHould be in folder `test/modules/paymentProcessor/Interfaces/IEverClearSpoke.sol`

test/modules/paymentProcessor/bridging/abstracts/mocks/Mock_EverclearPayment.sol
```diff
@@ -0,0 +1,70 @@
+// SPDX-License-Identifier: MIT
```
Zitzak: Can be in `test/utils/mocks/external/Mock_EverclearPayment.sol

Zitzak (changes requested):
The overall structure looks good @leeftk. Only comments I left are for moving the test files into the right directories.

I have not done an overall review as this PR is only for the architecture setup

Zitzak:
The CI failing has something to do with our RPC endpoint being a secret of the Inverter Github Org, and somehow the CI doesn't access when a PR is created from outside. We are looking into it fyi @leeftk

src/modules/paymentProcessor/abstracts/CrosschainBase_v1.sol
leeftk: Yea I had a lot of doubts around this which is why i asked in the slack but maybe its something we can clarify when we pair tomorrow. Closing for now as I adjusted this.

leeftk: Actually this doesn't align with the Mermaid diagram you sent, in the diagram it shows the function should be in the CrosschainBase and not only in the interface, also adding it to the interface and attempting to expose it throws an error as interfaces can only have external functions not internal function, so this renders the exposed contract useless as we wouldn't need that wrapper. Need more clarity here lets save this for the call.


https://www.mermaidchart.com/raw/6ea07035-6e3f-4c33-a146-8213381eb6c1?theme=light&version=v0.1&format=svg

leeftk:
Pushed up all the changes and resolved all the comments besides one. Think we need to review the inheritance chain together on a call cause seems like there's either an issue in the diagram or the way Im implementing it. We can discuss more on the call tomorrow.

src/modules/paymentProcessor/abstracts/CrosschainBase_v1.sol
leeftk: Fixed this and pushed it up.

leeftk:
All comments here have been addressed, we just need another review and then hopefully a merge once everything is checked again :)

test/modules/paymentProcessor/interfaces/IEverClearSpoke.sol
```diff
@@ -0,0 +1,116 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/**
+ * @title IEverclearSpoke
+ * @notice Interface for the Everclear spoke contract
+ */
+interface IEverclearSpoke {
```
Zitzak, line 8: I think this one could be removed right? We can import the everclear interface from the src folder if we need it for testing

test/modules/paymentProcessor/bridging/abstracts/Interfaces/IEverClearSpoke.sol
Zitzak: Only now seeing that it isn't used I believe

Zitzak (approved):
One file can probably be removed but this can be done in the next PR ser @leeftk

test/modules/paymentProcessor/interfaces/IEverClearSpoke.sol
Zitzak: Can be done in the next PR

PR #690, opened 2024-11-15
Fix: Resolve Module Address Collisions

marvinkruse:
**Why does this PR exist?**
I noticed that the original idea - when we switched to create2-based deployments in the factories - was to allow users to deploy workflows at the same address across networks. What we did not think about back then: the `ModuleFactory_v1` is not just called by end users, but mainly be the `OrchestratorFactory_v1`, which makes the following salt calculation not unique per user:

```Solidity
function _createSalt() internal returns (bytes32) {
    return keccak256(
        abi.encodePacked(_msgSender(), _deploymentNonces[_msgSender()]++)
    );
}
```

So essentially, the system only works for `OrchestratorFactory_v1`, where the user calls the contract.

**How did I fix this?**
We decided, that it's enough if we guarantee folks that their Orchestrator can be at the same address across networks, and the modules don't matter really. So to resolve the underlying issue, I added the `chainId` into the salt calculation in the `ModuleFactory_v1`. This makes it unique at all times and the address can never be griefed (also not on accident).

### DIFF
--- a/src/factories/ModuleFactory_v1.sol
+++ b/src/factories/ModuleFactory_v1.sol
@@ -291,7 +291,9 @@ contract ModuleFactory_v1 is
     ///         nonce is an increasing number for each user.
     function _createSalt() internal returns (bytes32) {
         return keccak256(
-            abi.encodePacked(_msgSender(), _deploymentNonces[_msgSender()]++)
+            abi.encodePacked(
+                _msgSender(), block.chainid, _deploymentNonces[_msgSender()]++
+            )
         );
     }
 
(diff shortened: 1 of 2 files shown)


FHieser (approved):

FHieser:
9boukq

FHieser (changes requested):
Need tests

Zitzak:
How to move forward with this PR? How does having these changes un-merged effect our deployments? @marvinkruse @FHieser @0xNuggan

FHieser:
@marvinkruse what is the state of this PR?

marvinkruse:
Closing for now, will reopen if we decide to use this so that we can have a cleaner stack of PRs for now.

PR #695, opened 2024-11-27
DRAFT - FM_ExternalPriceSetter, Oracle, blacklist contracts and interfaces implemented

vendrell46:
The scope of this PR is to review the proposed implementation, considering that we DO NOT consider it complete, but rather a work in progress.

We want to ensure we are going in the right direction by getting feedback from the team.


vendrell46:
selected the wrong branch to merge into

PR #697, opened 2024-11-27
FIX: feat wrapper mint transfer

Zitzak:
## What still needs to be done?
- Abstract fee collection like was done within the [Topos branch](https://github.com/InverterNetwork/contracts/blob/feature/ToposFundingManager/src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol#L271) (I thought we added this already in the past 🤔)
- We might want to abstract all of the fee related code into functions in general for the buy/sell internal functions. We don't need to create interfaces to be implemented in downstream contracts, but virtual functions which can be overridden
- Tests for the new function I added
- If collateral is not in the contract by design and the function to transfer collateral is overridden then this would fail due to the [following condition](https://github.com/InverterNetwork/contracts/blob/dev/src/modules/fundingManager/bondingCurve/abstracts/RedeemingBondingCurveBase_v1.sol#L243)

### DIFF
--- a/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
+++ b/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
@@ -603,7 +603,7 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
     // Overridden Internal Functions
 
     /// @notice Handles incoming collateral tokens by transferring them
-    ///         to the provider.
+    ///         from the provider.
     /// @param  _provider The address that provides the collateral tokens.
     ///         will be receive.
     /// @param  _amount The amount of collateral tokens to handle.
--- a/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
+++ b/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
@@ -604,9 +604,9 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
 
     /// @notice Handles incoming collateral tokens by transferring them
     ///         to the provider.
-    /// @param _provider The address that provides the collateral tokens.
-    ///                  will be receive.
-    /// @param _amount   The amount of collateral tokens to handle.
+    /// @param  _provider The address that provides the collateral tokens.
+    ///         will be receive.
+    /// @param  _amount The amount of collateral tokens to handle.
     function _handleCollateralTokensBeforeBuy(address _provider, uint _amount)
         internal
         virtual
--- a/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
+++ b/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
@@ -602,6 +602,21 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
     // -------------------------------------------------------------------------
     // Overridden Internal Functions
 
+    /// @notice Handles incoming collateral tokens by transferring them
+    ///         to the provider.
+    /// @param _provider The address that provides the collateral tokens.
+    ///                  will be receive.
+    /// @param _amount   The amount of collateral tokens to handle.
+    function _handleCollateralTokensBeforeBuy(address _provider, uint _amount)
+        internal
+        virtual
+        override
+    {
+        __Module_orchestrator.fundingManager().token().safeTransferFrom(
+            _provider, address(this), _amount
+        );
+    }
+
     /// @notice Handles issuance tokens by minting them to the receiver.
     /// @param  _receiver The address that will receive the bought tokens.
     /// @param  _issuanceTokenAmount The amount of issuance tokens to handle.
@@ -620,19 +635,5 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
         uint _collateralTokenAmount
     ) internal virtual override {
         token().safeTransfer(_receiver, _collateralTokenAmount);
-    }
-
-    /// @notice Handles incoming collateral tokens by transferring them
-    ///         to the provider.
-    /// @param _provider The address that provides the collateral tokens.
-    ///                  will be receive.
-    /// @param _amount   The amount of collateral tokens to handle.
-    function _handleCollateralTokensBeforeBuy(address _provider, uint _amount)
-        internal
-        virtual
-        override
-    {
-        IERC20 collateralToken = __Module_orchestrator.fundingManager().token();
-        collateralToken.safeTransferFrom(_provider, address(this), _amount);
     }
 }
--- a/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
+++ b/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
@@ -623,11 +623,11 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
     }
 
     /// @notice Handles incoming collateral tokens by transferring them to the receiver.
-    /// @param  _sender The address from which the collateral tokens will be send.
+    /// @param  _provider The address that provides the collateral tokens.
     /// @param  _receiver The address to which the collateral tokens will be receive.
     /// @param  _amount The amount of collateral tokens to handle.
     function _handleCollateralTokensBeforeBuy(
-        address _sender,
+        address _provider,
         address _receiver,
         uint _amount
     ) internal virtual override {
--- a/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
+++ b/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
@@ -605,7 +605,6 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
     /// @notice Handles incoming collateral tokens by transferring them
     ///         from the provider.
     /// @param  _provider The address that provides the collateral tokens.
-    ///         will be receive.
     /// @param  _amount The amount of collateral tokens to handle.
     function _handleCollateralTokensBeforeBuy(address _provider, uint _amount)
         internal
--- a/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
+++ b/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
@@ -621,4 +621,17 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
     ) internal virtual override {
         token().safeTransfer(_receiver, _collateralTokenAmount);
     }
+
+    /// @notice Handles incoming collateral tokens by transferring them to the receiver.
+    /// @param  _sender The address from which the collateral tokens will be send.
+    /// @param  _receiver The address to which the collateral tokens will be receive.
+    /// @param  _amount The amount of collateral tokens to handle.
+    function _handleCollateralTokensBeforeBuy(
+        address _sender,
+        address _receiver,
+        uint _amount
+    ) internal virtual override {
+        IERC20 collateralToken = __Module_orchestrator.fundingManager().token();
+        collateralToken.safeTransferFrom(_sender, _receiver, _amount);
+    }
 }
--- a/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
+++ b/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
@@ -66,7 +66,7 @@ import {SafeERC20} from "@oz/token/ERC20/utils/SafeERC20.sol";
  *                          to our Security Policy at security.inverter.network
  *                          or email us directly!
  *
- * @custom:version 1.1.1
+ * @custom:version 1.1.2
  *
  * @author  Inverter Network
  */
--- a/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
+++ b/src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
@@ -622,16 +622,17 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
         token().safeTransfer(_receiver, _collateralTokenAmount);
     }
 
-    /// @notice Handles incoming collateral tokens by transferring them to the receiver.
-    /// @param  _provider The address that provides the collateral tokens.
-    /// @param  _receiver The address to which the collateral tokens will be receive.
-    /// @param  _amount The amount of collateral tokens to handle.
-    function _handleCollateralTokensBeforeBuy(
-        address _provider,
-        address _receiver,
-        uint _amount
-    ) internal virtual override {
+    /// @notice Handles incoming collateral tokens by transferring them
+    ///         to the provider.
+    /// @param _provider The address that provides the collateral tokens.
+    ///                  will be receive.
+    /// @param _amount   The amount of collateral tokens to handle.
+    function _handleCollateralTokensBeforeBuy(address _provider, uint _amount)
+        internal
+        virtual
+        override
+    {
         IERC20 collateralToken = __Module_orchestrator.fundingManager().token();
-        collateralToken.safeTransferFrom(_sender, _receiver, _amount);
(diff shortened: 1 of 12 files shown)


src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -238,8 +238,8 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
         // Cache Collateral Token
         IERC20 collateralToken = __Module_orchestrator.fundingManager().token();
 
-        // Transfer collateral, confirming that correct amount == allowance
-        collateralToken.safeTransferFrom(
+        // Handle collateral transfer
```
FHieser:
```suggestion
        // Handle collateral tokens before buy
```

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -621,4 +621,17 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
     ) internal virtual override {
         token().safeTransfer(_receiver, _collateralTokenAmount);
     }
+
+    /// @notice Handles incoming collateral tokens by transferring them to the receiver.
+    /// @param  _sender The address from which the collateral tokens will be send.
+    /// @param  _receiver The address to which the collateral tokens will be receive.
+    /// @param  _amount The amount of collateral tokens to handle.
+    function _handleCollateralTokensBeforeBuy(
+        address _sender,
```
FHieser:
```suggestion
        address _provider,
```

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -306,6 +306,16 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
         uint _issuanceTokenAmount
     ) internal virtual;
 
+    /// @notice Virtual function to handle collateral tokens before a buy.
+    /// @param  _sender The address from which the collateral tokens will be send.
+    /// @param  _receiver The address to which the collateral tokens will be receive.
+    /// @param  _amount The amount of collateral tokens to handle.
+    function _handleCollateralTokensBeforeBuy(
+        address _sender,
```
FHieser:
```suggestion
        address _provider,
```

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
FHieser: We dont know if this is a transfer that will happening,
could be a proof of ownership for example to receive tokens

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -621,4 +621,17 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
     ) internal virtual override {
         token().safeTransfer(_receiver, _collateralTokenAmount);
     }
+
+    /// @notice Handles incoming collateral tokens by transferring them to the receiver.
+    /// @param  _sender The address from which the collateral tokens will be send.
```
FHieser:
```suggestion
    /// @param  _provider The address that provides the collateral tokens.
```

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -238,8 +238,8 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
         // Cache Collateral Token
         IERC20 collateralToken = __Module_orchestrator.fundingManager().token();
 
-        // Transfer collateral, confirming that correct amount == allowance
-        collateralToken.safeTransferFrom(
+        // Handle collateral transfer
+        _handleCollateralTokensBeforeBuy(
             _msgSender(), address(this), _depositAmount
```
FHieser:
```suggestion
            _msgSender(), _receiver, _depositAmount
```

FHieser: the BondingCurveBase isnt the receiver of the buy function.
address (this) can just be used in the _handleCollateralTokensBeforeBuy function directly
We can either leave _receiver as part of the function call in (for potential added info, although we dont need it in this case or we can remove from the function altogether

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -306,6 +306,16 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
         uint _issuanceTokenAmount
     ) internal virtual;
 
+    /// @notice Virtual function to handle collateral tokens before a buy.
+    /// @param  _sender The address from which the collateral tokens will be send.
+    /// @param  _receiver The address to which the collateral tokens will be receive.
+    /// @param  _amount The amount of collateral tokens to handle.
+    function _handleCollateralTokensBeforeBuy(
+        address _sender,
+        address _receiver,
```
FHieser:
```suggestion
        address,
```

FHieser: See comment above

test/e2e/fundingManager/BondingCurveTokenRescueE2E.t.sol
```diff
@@ -51,13 +51,13 @@ contract BondingCurveTokenRescueE2E is E2ETest {
...
         // BancorFormula 'formula' is instantiated in the E2EModuleRegistry
 
-        IBondingCurveBase_v1.IssuanceToken memory issuanceToken_properties =
-        IBondingCurveBase_v1.IssuanceToken({
-            name: "Bonding Curve Token",
-            symbol: "BCT",
-            decimals: 18,
-            maxSupply: type(uint).max - 1
-        });
+        // IBondingCurveBase_v1.IssuanceToken memory issuanceToken_properties =
```
FHieser: Outcommented
needs fix

FHieser (changes requested):
If you do that you might as well do the same thing for the issuance token
also tests are missing

test/e2e/fundingManager/BondingCurveTokenRescueE2E.t.sol
Zitzak: This has been commented because of the compiler warning. As further up the file test file other parts were also out commented I went with this approach

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -306,6 +306,16 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
         uint _issuanceTokenAmount
     ) internal virtual;
 
+    /// @notice Virtual function to handle collateral tokens before a buy.
+    /// @param  _sender The address from which the collateral tokens will be send.
```
marvinkruse:
```suggestion
    /// @param  _sender The address from which the collateral tokens will be sent.
```

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -306,6 +306,16 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
...
+    /// @param  _receiver The address to which the collateral tokens will be receive.
```
marvinkruse:
```suggestion
    /// @param  _receiver The address that will receive the collateral tokens.
```

src/modules/fundingManager/bondingCurve/abstracts/RedeemingBondingCurveBase_v1.sol
```diff
@@ -280,6 +280,8 @@ abstract contract RedeemingBondingCurveBase_v1 is
     }
 
     /// @notice Virtual function to handle collateral tokens after a successful sell.
+    /// @dev    The downstream contract is responsible for implementing checks to ensure the contract
+    ///         has the right balance.
```
marvinkruse: Line too long, 80 characters including all whitespaces etc.
```suggestion
    /// @dev    The downstream contract is responsible for implementing checks
    ///         to ensure the contract has the right balance.
```

test/modules/fundingManager/bondingCurve/utils/mocks/RedeemingBondingCurveBaseV1Mock.sol
```diff
@@ -85,14 +85,24 @@ contract RedeemingBondingCurveBaseV1Mock is RedeemingBondingCurveBase_v1 {
...
         virtual
         override
     {
-        distributeCollateralTokenFunctionCalled++;
+        distributeCollateralTokenAfterSellFunctionCalled++;
+    }
+
+    uint public distributeCollateralTokenBeforeBuyFunctionCalled;
+
+        function _handleCollateralTokensBeforeBuy(
```
marvinkruse: Looks indented?

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
marvinkruse: I'd call it `_provider` to be consistent. It's clear what is being provided, and otherwise we'd need to call the rest also similarly to be consistent (collateralTokenReceiver, collateralTokenAmount), which we do not do here (and shouldn't).

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -621,4 +621,17 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
...
+    /// @param  _receiver The address to which the collateral tokens will be receive.
```
marvinkruse: Not 100% sure as I only check on GitHub directly, but to me this looks like its >80 characters in total.

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
marvinkruse: This comment (in the code) is superfluous and doesn't add any information.

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
marvinkruse: This doesn't seem correct to me. The `_receiver` references the receiver of the bought tokens (i.e. issuance tokens in this case), not the collateral tokens. With this change, the collateral tokens would never reach the contract, which they should! Or am I blind? @FHieser

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
marvinkruse: I believe (again, not sure because of GitHub UI) that this might be over the 80 character limit (by 2 characters), so needs to be changed.

src/modules/fundingManager/bondingCurve/abstracts/RedeemingBondingCurveBase_v1.sol
marvinkruse: Then we can also adapt the line below, which is also too long.

test/e2e/fundingManager/BondingCurveTokenRescueE2E.t.sol
marvinkruse: ah, it's never used, we define the issuance token twice and never use the first one. in that case I'd just remove it, no need to keep this around, the git history does this for us anyways.

marvinkruse (changes requested):

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
FHieser: The idea was to adapt the _handleCollateralTokensBeforeBuy to use address(this) internally and take the receiver of the buy function as bonus information in case we need it somehow
But after talking to MarvinG about it we agreed to remove the receiver altogether

FHieser:
@marvinkruse @Zitzak could you please rereview?

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -306,6 +304,14 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
         uint _issuanceTokenAmount
     ) internal virtual;
 
+    /// @notice Virtual function to handle collateral tokens before a buy.
+    /// @param  _provider The address from which the collateral tokens
+    ///                   will be sent.
+    /// @param  _amount   The amount of collateral tokens to handle.
```
marvinkruse:
```suggestion
    /// @param  _provider The address from which the collateral tokens
    ///         will be sent.
    /// @param  _amount The amount of collateral tokens to handle.
```

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -602,6 +602,21 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
     // -------------------------------------------------------------------------
     // Overridden Internal Functions
 
+    /// @notice Handles incoming collateral tokens by transferring them
+    ///         to the provider.
+    /// @param _provider The address that provides the collateral tokens.
+    ///                  will be receive.
+    /// @param _amount   The amount of collateral tokens to handle.
```
marvinkruse:
```suggestion
    /// @param  _provider The address that provides the collateral tokens.
    ///         will be receive.
    /// @param  _amount The amount of collateral tokens to handle.
```

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
```diff
@@ -306,6 +304,14 @@ abstract contract BondingCurveBase_v1 is IBondingCurveBase_v1, Module_v1 {
         uint _issuanceTokenAmount
     ) internal virtual;
 
+    /// @notice Virtual function to handle collateral tokens before a buy.
+    /// @param  _provider The address from which the collateral tokens
```
marvinkruse, line 307: Why `_provider` instead of just `_sender`? "Provider" to me has a different connotation, but may just be me. Technically of course they "provide" the tokens, but sounds like a donation/provision mechanism instead of just a sending of funds?

src/modules/fundingManager/bondingCurve/abstracts/RedeemingBondingCurveBase_v1.sol
```diff
@@ -280,6 +280,8 @@ abstract contract RedeemingBondingCurveBase_v1 is
     }
 
     /// @notice Virtual function to handle collateral tokens after a successful sell.
+    /// @dev    The downstream contract is responsible for implementing checks
+    ///         to ensure the contract has the right balance.
     /// @param  _receiver  The address for which the collateral tokens will be handled.
```
marvinkruse: Just saw this as well, I know you didn't do it, but as we touch this NatSpec section we may just fix that.
```suggestion
    /// @param  _receiver The address for which the collateral tokens will be handled.
```

marvinkruse (changes requested):
Some small comments, otherwise good I think.

src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
FHieser: Thats the idea,
by abstracting it we actually enable the implementation of potential provision mechanisms instead of "just" sending funds :D

Zitzak:
Is this one ready for another review @FHieser ?

FHieser:
Yes your right @Zitzak 
Already wrote @marvinkruse

FHieser (approved):
9cu7gg

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -602,6 +602,21 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
     // -------------------------------------------------------------------------
     // Overridden Internal Functions
 
+    /// @notice Handles incoming collateral tokens by transferring them
+    ///         to the provider.
```
marvinkruse:
```suggestion
    /// @notice Handles incoming collateral tokens by transferring them
    ///         from the provider.
```

marvinkruse (changes requested):

marvinkruse (approved):

PR #698, opened 2024-11-28
Fix: Account for wrapped Issuance Tokens

marvinkruse:
The current issue is that `getIssuanceToken` in the bonding-curve-based funding managers does not return the issuance token if a wrapper is used. To address this I added a fallback in the function to attempt the retrieval of an underlying token in the case of it existing. If it doesn't exist, we return the existing storage value.

This solves the issue for now, as we can not get rid of the wrappers for existing workflows.


FHieser (approved):

0xNuggan:
LGTM @fabianschu I think this would remove the need to do changes on the indexer

FHieser:
@marvinkruse  what is the state of this PR?

PR #700, opened 2024-12-05
Sc 848 task copy fee adaption from topos to dev

FHieser:
Abstracts the collection of the projectfee into its own function
Directly copied from the topos branch #667

### DIFF
--- a/test/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.t.sol
+++ b/test/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.t.sol
@@ -1214,23 +1214,27 @@ contract BondingCurveBaseV1Test is ModuleTest {
 
     /*  Test internal _projectFeeCollected()
         └── When the function _projectFeeCollected() gets called
+            └── Then the ProjectCollateralFeeAdded Event will be emitted
             └── Then the _workflowFeeAmount will be added to the projectCollateralFeeCollected
     */
-    function testInternalProjectFeeCollected_works(uint _workflowFeeAmount)
+    function testInternalProjectFeeCollected_works(uint _projectFeeAmount)
         public
     {
         uint currentFeeCollected =
             bondingCurveFundingManager.projectCollateralFeeCollected();
 
+        vm.expectEmit(true, true, true, true);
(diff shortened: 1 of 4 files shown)


test/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.t.sol
```diff
@@ -1212,6 +1212,28 @@ contract BondingCurveBaseV1Test is ModuleTest {
...
+            └── Then the _workflowFeeAmount will be added to the projectCollateralFeeCollected
+    */
+    function testInternalProjectFeeCollected_works(uint _workflowFeeAmount)
+        public
+    {
+        uint currentFeeCollected =
+            bondingCurveFundingManager.projectCollateralFeeCollected();
+
+        // Execute Tx
+        bondingCurveFundingManager.exposed_projectFeeCollected(
```
Zitzak, line 1230: Could you please add the emit tests, as well as add it to the gherkin

Zitzak (changes requested):

FHieser:
@Zitzak could you pls rereview

Zitzak (approved):

marvinkruse:
Just saw this was merged without me being a reviewer, three things:

(1) Why is the downstream contract not version upgraded?
If `BondingCurveBase_v1` changes, the version of all contracts inheriting it also needs to change, i.e. `FM_BC_Bancor_Redeeming_VirtualSupply_v1` and `FM_BC_Restricted_Bancor_Redeeming_VirtualSupply_v1`, as the code of these also changed subsequently, speaking from a contract standpoint and it not the same anymore, so it needs changing.

(2) The `_projectFeeCollected` function Natspec should have `@notice`, not just `@dev`

(3) The name of the merged commit is not as we agreed.

Please mark me as a reviewer for **any** PRs that go into **dev** until we are all up to speed on using the Inverter Standard, otherwise it's just double work. (@FHieser @Zitzak)

marvinkruse:
This merge-to-dev-commit should be reverted, or did you guys already rebase too many other PRs?

Zitzak:
I did merge it into the feature branch for the External price funding manager. However if you would like to change it than please go ahead and I force push it to the feature branch @marvinkruse

PR #703, opened 2024-12-20
[Audit] Revised Payment Order Struct

marvinkruse:
This PR contains changes related to our revised Payment Order Struct. A payment order is a formalized set of information used by payment clients and are sent to a payment processor, who acts based upon the information contained in the order.

**What have we done**
We revamped the struct and made a distinction between base information (i.e. amount, token, chainId, etc.) and additional information, where we saw that the additional information needs to allow for certain changes over time. We ultimately arrived at this struct:

```Solidity
struct PaymentOrder {
    address recipient;    
    address paymentToken;
    uint amount;
    uint originChainId;
    uint targetChainId;        
    bytes32 flags;
    bytes32[] data;
}
```
The way the `flags` and `data` fields work is as follows:
The `flags` field encodes which information is contained in the `data` array. The way this happens is similar to a bit-mask, where each of the bits in the `bytes32` variable represents one potential data type. We currently have these slots defined:
```
    | Flag | Variable type | Name       | Description                         |
    |------|---------------|------------|-------------------------------------|
    | 0    | bytes32       | orderID    | ID of the order.                    |
    | 1    | uint256       | start      | Start date of the streaming period. | 
    | 2    | uint256       | cliff      | Duration of the cliff period.       |
    | 3    | uint256       | end        | Due Date of the order               |
    | ...  | ...           | ...        | (yet unassigned)                    |
    | 255  | .             | .          | (Max Value).                        | 
    |------|---------------|------------|-------------------------------------|
```
So in this example, if we want to make use of the streamed payment processor and want to encode the `start`, `cliff`, and `end` times, we would set the `flags` value to be `0000 [...] 0000 1110`, so the flags at position 1, 2 and 3 and set to 1, while the rest is set to 0. If the used payment processor allows for `start`, `cliff`, and `end` values, it will check whether these slots are set (i.e. are 1) and - if that is the case - extract these values from the array.

The data array only contains values for slots where the flag is 1, so for the example above (`0000 [...] 0000 1110`) it would contain three data entries (length of 3), where the first one is from flag 1 and so on.

The documentation around this and the definition of the available slots can be found in `IERC20PaymentClientBase_v1.sol`.

### DIFF
--- a/src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
+++ b/src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
@@ -69,8 +69,8 @@ contract LM_PC_PaymentRouter_v1 is
     ) external override(Module_v1) initializer {
         __Module_init(orchestrator_, metadata);
 
-        // Set the flags for the PaymentOrders
-        uint8[] memory flags = new uint8[](3); // The Module will use 3 flags
+        // Set the flags for the PaymentOrders (this module uses 3 flags).
+        uint8[] memory flags = new uint8[](3);
         flags[0] = 1; // start, flag_ID 1
         flags[1] = 2; // cliff, flag_ID 2
         flags[2] = 3; // end, flag_ID 3
--- a/src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
+++ b/src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
@@ -61,6 +61,24 @@ contract LM_PC_PaymentRouter_v1 is
     bytes32 public constant PAYMENT_PUSHER_ROLE = "PAYMENT_PUSHER";
 
     //--------------------------------------------------------------------------
+    // Initializer
+    function init(
+        IOrchestrator_v1 orchestrator_,
+        Metadata memory metadata,
+        bytes memory configData
+    ) external override(Module_v1) initializer {
+        __Module_init(orchestrator_, metadata);
+
+        // Set the flags for the PaymentOrders
+        uint8[] memory flags = new uint8[](3); // The Module will use 3 flags
+        flags[0] = 1; // start, flag_ID 1
+        flags[1] = 2; // cliff, flag_ID 2
+        flags[2] = 3; // end, flag_ID 3
+
+        __ERC20PaymentClientBase_v1_init(flags);
+    }
+
+    //--------------------------------------------------------------------------
     // Mutating Functions
 
     /// @inheritdoc ILM_PC_PaymentRouter_v1
@@ -72,16 +90,29 @@ contract LM_PC_PaymentRouter_v1 is
         uint cliff,
         uint end
     ) public onlyModuleRole(PAYMENT_PUSHER_ROLE) {
-        _addPaymentOrder(
-            PaymentOrder(
-                recipient,
-                paymentToken,
-                amount,
-                start == 0 ? block.timestamp : start,
-                cliff,
-                end
-            )
-        );
+        bytes32 flags;
+        bytes32[] memory data;
+
+        {
+            bytes32[] memory paymentParameters = new bytes32[](3);
+            paymentParameters[0] = bytes32(start);
+            paymentParameters[1] = bytes32(cliff);
+            paymentParameters[2] = bytes32(end);
+
+            (flags, data) = _assemblePaymentConfig(paymentParameters);
+        }
+
+        PaymentOrder memory order = PaymentOrder({
+            recipient: recipient,
+            paymentToken: paymentToken,
+            amount: amount,
+            originChainId: block.chainid,
+            targetChainId: block.chainid,
+            flags: flags,
+            data: data
+        });
+
+        _addPaymentOrder(order);
 
         // call PaymentProcessor
         __Module_orchestrator.paymentProcessor().processPayments(
@@ -105,20 +136,33 @@ contract LM_PC_PaymentRouter_v1 is
                 || paymentTokens.length != numOfOrders
                 || amounts.length != numOfOrders
         ) {
-            revert Module__ERC20PaymentClientBase__ArrayLengthMismatch();
+            revert Module__LM_PC_PaymentRouter_v1__ArrayLengthMismatch();
+        }
+
+        bytes32 flags;
+        bytes32[] memory data;
+
+        {
+            bytes32[] memory paymentParameters = new bytes32[](3);
+            paymentParameters[0] = bytes32(start);
+            paymentParameters[1] = bytes32(cliff);
+            paymentParameters[2] = bytes32(end);
+
+            (flags, data) = _assemblePaymentConfig(paymentParameters);
         }
 
         // Loop through the arrays and add Payments
         for (uint8 i = 0; i < numOfOrders; i++) {
             _addPaymentOrder(
-                PaymentOrder(
-                    recipients[i],
-                    paymentTokens[i],
-                    amounts[i],
-                    start == 0 ? block.timestamp : start,
-                    cliff,
-                    end
-                )
+                PaymentOrder({
+                    recipient: recipients[i],
+                    paymentToken: paymentTokens[i],
+                    amount: amounts[i],
+                    originChainId: block.chainid,
+                    targetChainId: block.chainid,
+                    flags: flags,
+                    data: data
+                })
             );
         }
 
--- a/src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
+++ b/src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
@@ -95,9 +95,9 @@ contract LM_PC_PaymentRouter_v1 is
 
         {
             bytes32[] memory paymentParameters = new bytes32[](3);
-            paymentParameters[0] = bytes32(start);
+            paymentParameters[0] = bytes32(start == 0 ? block.timestamp : start);
             paymentParameters[1] = bytes32(cliff);
-            paymentParameters[2] = bytes32(end);
+            paymentParameters[2] = bytes32(end == 0 ? block.timestamp : end);
 
             (flags, data) = _assemblePaymentConfig(paymentParameters);
         }
--- a/src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
+++ b/src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
@@ -60,6 +60,10 @@ contract LM_PC_PaymentRouter_v1 is
     /// @dev	The role that allows the pushing of payments.
     bytes32 public constant PAYMENT_PUSHER_ROLE = "PAYMENT_PUSHER";
 
+    uint8 public constant FLAG_START = 1;
+    uint8 public constant FLAG_CLIFF = 2;
+    uint8 public constant FLAG_END = 3;
+
     //--------------------------------------------------------------------------
     // Initializer
     function init(
@@ -70,10 +74,10 @@ contract LM_PC_PaymentRouter_v1 is
         __Module_init(orchestrator_, metadata);
 
         // Set the flags for the PaymentOrders (this module uses 3 flags).
-        uint8[] memory flags = new uint8[](3);
-        flags[0] = 1; // start, flag_ID 1
-        flags[1] = 2; // cliff, flag_ID 2
-        flags[2] = 3; // end, flag_ID 3
+        bytes32 flags;
+        flags |= bytes32(1 << FLAG_START);
+        flags |= bytes32(1 << FLAG_CLIFF);
+        flags |= bytes32(1 << FLAG_END);
 
         __ERC20PaymentClientBase_v1_init(flags);
     }
--- a/src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
+++ b/src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
@@ -112,6 +112,10 @@ contract LM_PC_RecurringPayments_v1 is
     /// @dev	List of RecurringPayment id's.
     LinkedIdList.List _paymentList;
 
+    /// @dev	Flags used for the payment order.
+    uint8 constant FLAG_START = 1;
+    uint8 constant FLAG_END = 3;
+
     /// @dev	Storage gap for future upgrades.
     uint[50] private __gap;
 
@@ -139,10 +143,9 @@ contract LM_PC_RecurringPayments_v1 is
 
         emit EpochLengthSet(newEpochLength);
 
-        // Set the flags for the PaymentOrders (this module uses 2 flags).
-        uint8[] memory flags = new uint8[](2);
-        flags[0] = 1; // start, flag_ID 1
-        flags[1] = 3; // end, flag_ID 3
+        bytes32 flags;
+        flags |= bytes32(1 << FLAG_START);
+        flags |= bytes32(1 << FLAG_END);
 
         __ERC20PaymentClientBase_v1_init(flags);
     }
--- a/src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
+++ b/src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
@@ -139,8 +139,8 @@ contract LM_PC_RecurringPayments_v1 is
 
         emit EpochLengthSet(newEpochLength);
 
-        // Set the flags for the PaymentOrders
-        uint8[] memory flags = new uint8[](2); // The Module will use 2 flags
+        // Set the flags for the PaymentOrders (this module uses 2 flags).
+        uint8[] memory flags = new uint8[](2);
         flags[0] = 1; // start, flag_ID 1
         flags[1] = 3; // end, flag_ID 3
 
--- a/src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
+++ b/src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
@@ -128,6 +128,7 @@ contract LM_PC_RecurringPayments_v1 is
         // Set empty list of RecurringPayment
         _paymentList.init();
 
+        // Set the epoch Data
         uint newEpochLength = abi.decode(configData, (uint));
         epochLength = newEpochLength;
 
@@ -137,6 +138,13 @@ contract LM_PC_RecurringPayments_v1 is
         }
 
         emit EpochLengthSet(newEpochLength);
+
+        // Set the flags for the PaymentOrders
+        uint8[] memory flags = new uint8[](2); // The Module will use 2 flags
+        flags[0] = 1; // start, flag_ID 1
+        flags[1] = 3; // end, flag_ID 3
+
+        __ERC20PaymentClientBase_v1_init(flags);
     }
 
     //--------------------------------------------------------------------------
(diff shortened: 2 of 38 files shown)


src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
```diff
@@ -72,16 +90,29 @@ contract LM_PC_PaymentRouter_v1 is
...
-                cliff,
-                end
-            )
-        );
+        bytes32 flags;
+        bytes32[] memory data;
+
+        {
+            bytes32[] memory paymentParameters = new bytes32[](3);
+            paymentParameters[0] = bytes32(start);
```
omega-audits: It doesn't do the check for start == 0 ? block.timestamp anymore, not sure if it was removed intentionally as it still seems useful

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -86,12 +86,33 @@ abstract contract ERC20PaymentClientBase_v1 is
...
 
+    /// @dev	Initializes the staking contract.
+    /// @param  flags_ The flags, represented as an array of uint8 containing
+    ///         the flag IDs between 0 and 255.
+    function __ERC20PaymentClientBase_v1_init(uint8[] memory flags_)
+        internal
+        onlyInitializing
+    {
+        uint amountOfFlags = flags_.length;
+        if (amountOfFlags > type(uint8).max) {
```
omega-audits: It'd make sense to move this check to the _setFlags as it is a requirement any time it will be used.
Also it does not seem necessary to pass the flags_.length since it's always checked to be the flags length in the _setFlags, which is just unnecessary duplication then

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -915,4 +937,52 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
+
+    function _getStreamingDetails(bytes32 flags, bytes32[] memory data)
+        internal
+        view
+        returns (uint start, uint cliff, uint end)
+    {
+        uint dataIdx = 0;
+
+        bool hasStart = (uint(flags) & (1 << 1)) != 0;
+        start = hasStart ? uint(data[dataIdx]) : defaultStart;
```
omega-audits: If default start and end are 0 you may want to have them as block.timestamp, it seems like this option could be useful as you would usually want the default to be whatever block.timestamp is, not a specific date.

0xNuggan: Thanks fo the tip! While it makes sense, we've chosen to go against it since being able to default variables to 0 (and  therefore create valid payment Orders, even if they are immediately claimable), seems to me better than introducing block.timestamp values that the user might not expect.
I've created a linear issue to track this, in case we choose to look for a different way to handle it.

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v1.sol
```diff
@@ -5,24 +5,83 @@ pragma solidity ^0.8.0;
...
+    | 3    | uint256       | end        | Due Date of the order               |
+    | ...  | ...           | ...        | (yet unassigned)                    |
+    | 255  | .             | .          | (Max Value).                        | 
+    |------|---------------|------------|-------------------------------------|
+    */
+
+    /*
+    | Flag | Name       | Disclaimer                                          |
+    |------|------------|-----------------------------------------------------|
+    | 0    | orderID    | The order id should be a hashed value of an         |
```
Zitzak: It's missing the [comment update](https://github.com/InverterNetwork/contracts/pull/707) merged into dev, which reverted this comment. We prob need to rebase/merge from dev.

Just something I noticed while working on another branch

marvinkruse: Yes, good reminder: please rebase from dev so this is up to date. I can also do this if you want (@0xNuggan) - lmk!

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -101,13 +101,14 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
     ///         client => paymentReceiver => arrayOfStreamIdsWithPendingPayment(uint[]).
     mapping(address => mapping(address => uint[])) private activeStreams;
 
-    /// @dev    Default start, cliff and end times for new payment orders.
-    uint private defaultStart;
-    uint private defaultCliff;
-    uint private defaultEnd;
+    /// @dev    The flags that this PaymentProcessor uses
+    bytes32 internal constant processor_flags =
+        0x000000000000000000000000000000000000000000000000000000000000000e; // 1110
```
marvinkruse: I'd not add stuff at the end as the line is too ling already anyways and it would not be in the docs otherwise if we generate these from the NatSpec.

```suggestion
    /// @dev    The flags that this PaymentProcessor uses.
    ///         Contains the value 1110.
    bytes32 internal constant processor_flags =
        0x000000000000000000000000000000000000000000000000000000000000000e;
```

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -101,13 +101,14 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
+    /// @dev    The default values for those flags, in ascending order: start, cliff, end
```
marvinkruse: Line too long and a missing dot at the end.
```suggestion
    /// @dev    The default values for those flags, in ascending order: start,
    ///         cliff, end.
```

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -101,13 +101,14 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
+    uint[3] private defaultValues;
 
     /// @dev	Storage gap for future upgrades.
-    uint[47] private __gap;
+    uint[46] private __gap;
```
marvinkruse: Any reason why this was decreased? The `processor_flags` are a constant and thus do not use any storage slots, so we only have the 3-length-array instead of three individual variables, which takes the same storage space, so to me 47 would be the right value here still.
```suggestion
    uint[47] private __gap;
```

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -101,13 +101,14 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
     ///         client => paymentReceiver => arrayOfStreamIdsWithPendingPayment(uint[]).
     mapping(address => mapping(address => uint[])) private activeStreams;
 
-    /// @dev    Default start, cliff and end times for new payment orders.
-    uint private defaultStart;
-    uint private defaultCliff;
-    uint private defaultEnd;
+    /// @dev    The flags that this PaymentProcessor uses
+    bytes32 internal constant processor_flags =
```
marvinkruse: Any reason why this is named `processor_flags` instead of `processorFlags` or (as it's a `constant`) even `PROCESSOR_FLAGS`?

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -948,23 +957,48 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
 
-    function _getStreamingDetails(bytes32 flags, bytes32[] memory data)
-        internal
-        view
-        returns (uint start, uint cliff, uint end)
-    {
-        uint dataIdx = 0;
+    function _getStreamingDetails(
+        bytes32 order_flags,
+        bytes32[] memory order_data
```
marvinkruse: Why are the parameters using snake case and not camel case?
Also: the underscores at the end are missing as well.

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -948,23 +957,48 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
+    ) internal view returns (uint start_, uint cliff_, uint end_) {
+        uint[3] memory return_data;
+
+        uint8 position_in_order_data = 0;
```
marvinkruse: Again: using snake case

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -948,23 +957,48 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
+        uint8 position_in_return_data = 0;
 
-        bool hasStart = (uint(flags) & (1 << 1)) != 0;
-        start = hasStart ? uint(data[dataIdx]) : defaultStart;
-        if (hasStart) dataIdx += 1;
+        for (uint i = 0; i < 256; i++) {
+            if (position_in_return_data == return_data.length) {
+                // we have either:
+                // - reached the end of the order_data array
+                // - already checked for all the values this PaymentProcessor will need
```
marvinkruse: Line is > 80 characters

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -948,23 +957,48 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
+                //      ==> exit loop
+                break;
+            }
 
-        bool hasCliff = (uint(flags) & (1 << 2)) != 0;
-        cliff = hasCliff ? uint(data[dataIdx]) : defaultCliff;
-        if (hasCliff) dataIdx += 1;
+            //uint order_bit = (uint(order_flags) & (1 << i));
```
marvinkruse: Missing whitespaces between `//` and the comment.
What are these comments as well? What do these do here?

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -948,23 +957,48 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
+            bool order_bit = (uint(order_flags) & (1 << i)) != 0;
+            //uint processor_bit = (uint(processor_flags) & (1 << i));
+            bool processor_bit = (uint(processor_flags) & (1 << i)) != 0;
+            if (order_bit == true && processor_bit == false) {
+                // the PaymentProcessor does not use that value -> skip that data slot in the order
```
marvinkruse: Line length (throughout the function, just leaving it here again)

test/modules/paymentProcessor/PP_Streaming_v1.t.sol
```diff
@@ -1947,11 +1947,16 @@ contract PP_StreamingV1Test is ModuleTest {
     function test__getStreamingDetails(bytes32 flags, bytes32[] memory data)
         public
     {
+        bool hasOrderId = false; // the processor doesn't use this, but it could be part of recieved data
```
marvinkruse:
```suggestion
        // the processor doesn't use this, but it could be part of received data.
        bool hasOrderId = false;
```

test/modules/paymentProcessor/PP_Streaming_v1.t.sol
```diff
@@ -1979,6 +1984,7 @@ contract PP_StreamingV1Test is ModuleTest {
             paymentProcessor.exposed_getStreamingDetails(flags, data);
 
         uint idx = 0;
+        if (hasOrderId) idx++; // if there was an orderId, we start comparing results one index further
```
marvinkruse:
```suggestion
        // if there was an orderId, we start comparing results one index further.
        if (hasOrderId) idx++;
```

marvinkruse (commented):
Thank you for adressing my overall comment of the data array ordering thing I mentioned, to me it looks like that was fully addressed. We probably just need to make sure that each payment processor does it like you did it here (i.e. applies the same principle). May be worth to look into creating a library or a base here for payment processors that handles this?

marvinkruse:
Tagging @0xNuggan - as this is my PR I can't explicitly tag my comments as requests for changes, so just letting you know via this comment that I submitted my review. 

Feel free to close any comments that you feel like you properly addressed and just leave anything open you feel like I should look at again/you have comments on/etc.

0xNuggan:
Thanks for the review! Addressed the comments

0xNuggan:
About moving the array logic into a library I agree, but I can't think of a good place right now where we shoudl put it. But worth keeping in mind.

src/modules/logicModule/LM_PC_PaymentRouter_v1.sol
```diff
@@ -60,6 +60,24 @@ contract LM_PC_PaymentRouter_v1 is
...
+        Metadata memory metadata,
+        bytes memory configData
+    ) external override(Module_v1) initializer {
+        __Module_init(orchestrator_, metadata);
+
+        // Set the flags for the PaymentOrders
+        uint8[] memory flags = new uint8[](3); // The Module will use 3 flags
+        flags[0] = 1; // start, flag_ID 1
+        flags[1] = 2; // cliff, flag_ID 2
+        flags[2] = 3; // end, flag_ID 3
```
marvinkruse:
```suggestion
        // Set the flags for the PaymentOrders (this module uses 3 flags).
        uint8[] memory flags = new uint8[](3);
        flags[0] = 1; // start, flag_ID 1
        flags[1] = 2; // cliff, flag_ID 2
        flags[2] = 3; // end, flag_ID 3
```

src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
```diff
@@ -137,6 +138,13 @@ contract LM_PC_RecurringPayments_v1 is
         }
 
         emit EpochLengthSet(newEpochLength);
+
+        // Set the flags for the PaymentOrders
+        uint8[] memory flags = new uint8[](2); // The Module will use 2 flags
+        flags[0] = 1; // start, flag_ID 1
+        flags[1] = 3; // end, flag_ID 3
```
marvinkruse:
```suggestion
        // Set the flags for the PaymentOrders (this module uses 2 flags).
        uint8[] memory flags = new uint8[](2);
        flags[0] = 1; // start, flag_ID 1
        flags[1] = 3; // end, flag_ID 3
```

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -101,8 +101,17 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
     ///         client => paymentReceiver => arrayOfStreamIdsWithPendingPayment(uint[]).
     mapping(address => mapping(address => uint[])) private activeStreams;
 
+    /// @dev    The flags that this PaymentProcessor uses
```
marvinkruse:
```suggestion
    /// @dev    The flags that this PaymentProcessor uses.
```

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -101,8 +101,17 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
+    ///         Contains the value 1110.
+    bytes32 internal constant PROCESSOR_FLAGS =
+        0x000000000000000000000000000000000000000000000000000000000000000e;
+
+    /// @dev    The default values for those flags, in ascending order: start
+    //          cliff, end
```
marvinkruse:
```suggestion
    //          cliff, end.
```

marvinkruse (commented):
Looks good, just small typos.

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -121,6 +144,26 @@ abstract contract ERC20PaymentClientBase_v1 is
...
+        }
+
+        _flagCount = uint8(amountOfFlags);
+
+        _flags = 0;
+        for (uint8 i = 0; i < amountOfFlags; i++) {
+            _flags |= bytes32((1 << flags_[i]));
+        }
+
+        emit FlagsSet(_flagCount, _flags);
```
omega-audits: Better to use amountOfFlags instead of reading from storage

src/modules/logicModule/abstracts/ERC20PaymentClientBase_v1.sol
```diff
@@ -86,12 +86,29 @@ abstract contract ERC20PaymentClientBase_v1 is
...
-    uint[50] private __gap;
+    uint[48] private __gap;
 
     //--------------------------------------------------------------------------
     // Internal Mutating Functions
 
+    /// @dev	Initializes the staking contract.
+    /// @param  flags_ The flags, represented as an array of uint8 containing
+    ///         the flag IDs between 0 and 255.
+    function __ERC20PaymentClientBase_v1_init(uint8[] memory flags_)
```
jellegerbrandy: So in your system you are now using 2 representations for setting the flags:

- flags are uint8[] type that is used (here in the setter function)
- flags are a bitmaps of type bytes32 (used everywhere elese, also in public getters adn events)

I'd just use a single represeentation (i.e. just use bitmaps)

src/modules/logicModule/LM_PC_RecurringPayments_v1.sol
```diff
@@ -137,6 +138,13 @@ contract LM_PC_RecurringPayments_v1 is
         }
 
         emit EpochLengthSet(newEpochLength);
+
+        // Set the flags for the PaymentOrders (this module uses 2 flags).
+        uint8[] memory flags = new uint8[](2);
+        flags[0] = 1; // start, flag_ID 1
```
jellegerbrandy: since you will use these values ("0", "1", "2" etc) in different places with a fixed meaning, it would be a good idea to introduce constants here for better readability and a lower chance of making mistakes. So you'd have a `const FLAG_ORDERID = 0` etc, and you could call this here `__ERC20PaymentClientBase_v1_init([FLAG_START, FLAG_END])`

src/modules/paymentProcessor/PP_Streaming_v1.sol
```diff
@@ -915,4 +949,79 @@ contract PP_Streaming_v1 is Module_v1, IPP_Streaming_v1 {
...
+            bool processorBit = (uint(PROCESSOR_FLAGS) & (1 << i)) != 0;
+
+            if (orderBit == true && processorBit == false) {
+                // the P_P does not use that value
+                //      ==> skip that data slot in the order
+                positionInOrderData++;
+            }
+            if (orderBit == false && processorBit == true) {
+                // the P_P needs the value, but it's missing in the order
+                //      ==> use default value
```
jellegerbrandy, line 997: Would it not be better (safer) if this is an error? This after all is not a case where the order does not provide a value, it is a case where the "order type" is different from what is expected (I mean that if the flag value is different, this is not the same as providing a "0" value and have the PP fill that in with a default value)
For example, if the order says "This is an order with no start and end date", then it is it is better that a PP that is made for streaming orders says "Ok, you probably made a istake sending me this order" rather than have it say "O, ok, then I will just make it a streaming order which starts in 17 days and lasts 189 hours"

0xNuggan: It's intended behavior, since we want to give the PaymentProcessor to build a valid PaymentOrder using its defaults. I've added a check that the resulting PaymentOrder after "reconstruction" is valid.

0xNuggan:
Closing this in favor or #717

PR #704, opened 2024-12-20
[Audit] FM_PC_ExternalPrice_Redeeming, LM_ManualExternalPriceSetter_v1, 

Zitzak:
## What needs to be audit.
The contracts that need to be audited are
**Funding Manager**

- src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol

**Logic Module**

- src/modules/logicModule/LM_Oracle_Permissioned_v1.sol

**Payment Processor**

- src/modules/paymentProcessor/PP_Queue_v1.sol
- src/modules/paymentProcessor/interfaces/IPP_Queue_ManualExecution_v1.sol

**Blacklist Token**

- src/external/token/ERC20Issuance_Blacklist_v1.sol


src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol
```diff
@@ -0,0 +1,214 @@
...
+    /// @notice Decimals of the collateral token (e.g., USDC with 6 decimals).
+    /// @dev    This is the token used to pay/buy with.
+    uint8 private _collateralTokenDecimals;
+
+    /// @notice Decimals of the issuance token (e.g., ISS with 18 decimals).
+    /// @dev    This is the token being bought/sold.
+    uint8 private _issuanceTokenDecimals;
+
+    /// @dev    Storage gap for upgradeable contracts.
+    uint[50] private __gap;
```
omega-audits: The convention is to put the gap at the end of the file, so you never accidentally put anything after it. However that is just the convetion and though I'd reccomend to follow it, it's not a necessity.

src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol
```diff
@@ -0,0 +1,214 @@
...
+        // Store token decimals for price normalization.
+        _collateralTokenDecimals = IERC20Metadata(collateralToken).decimals();
+        _issuanceTokenDecimals = IERC20Metadata(issuanceToken).decimals();
+    }
+
+    // -------------------------------------------------------------------------
+    // External Functions
+
+    /// @inheritdoc ILM_ManualExternalPriceSetter_v1
+    function setIssuancePrice(uint price_)
```
omega-audits: Since issuance and redemption price are likely to move together, you could consider either uniting the setters from them, or adding a setter which atomically updates both, as in some cases you may need the updates to happen atomically.

omega-audits: For example if buy price was 2, and sell price was 1, but now they're adjusted to 4 buy and 3 sell, you don't want the sell price updated before the buy price, so you either have to wait until new buy price is confirmed, or do the change atomically.

src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol
```diff
@@ -0,0 +1,214 @@
...
+
+    /// @inheritdoc ILM_ManualExternalPriceSetter_v1
+    function setRedemptionPrice(uint price_)
+        external
+        onlyModuleRole(PRICE_SETTER_ROLE)
+    {
+        if (price_ == 0) revert Module__LM_ExternalPriceSetter__InvalidPrice();
+
+        // Normalize price to internal decimal precision.
+        _redemptionPrice = _normalizePrice(price_, _issuanceTokenDecimals);
```
omega-audits: According to the docs above, this should be the _collateralTokenDecimals, not _issuanceTokenDecimals.

src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol
```diff
@@ -0,0 +1,214 @@
...
+        emit RedemptionPriceSet(price_, block.timestamp);
```
omega-audits: You could usually get the block time of the event when reading it, which makes it a waste of gas to also emit it here (and in the other events as well)

src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol
```diff
@@ -0,0 +1,214 @@
...
+
+    /// @notice Gets current price for token issuance (buying tokens).
+    /// @return price_ Current price in 18 decimals (collateral tokens per 1
+    ///         issuance token).
+    /// @dev    Example: If price is 2 USDC/ISS, returns 2e18 (2 USDC needed for
+    ///         1 ISS).
+    function getPriceForIssuance() external view returns (uint) {
+        if (_issuancePrice == 0) {
+            revert Module__LM_ExternalPriceSetter__InvalidPrice();
+        }
```
omega-audits: This can never happen since the setter disallows it

src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol
```diff
@@ -0,0 +1,214 @@
...
+
+    /// @notice Gets current price for token redemption (selling tokens).
+    /// @return price_ Current price in 18 decimals (collateral tokens per 1
+    ///         issuance token).
+    /// @dev    Example: If price is 1.9 USDC/ISS, returns 1.9e18 (1.9 USDC
+    ///         received for 1 ISS).
+    function getPriceForRedemption() external view returns (uint) {
+        if (_redemptionPrice == 0) {
+            revert Module__LM_ExternalPriceSetter__InvalidPrice();
+        }
```
omega-audits: This can never happen since the setter disallows it

src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol
```diff
@@ -0,0 +1,214 @@
...
+
+    //--------------------------------------------------------------------------
+    // Internal Functions
+
+    /// @notice Normalizes a price from token decimals to internal decimals.
+    /// @param  price_ The price to normalize.
+    /// @param  tokenDecimals_ The decimals of the token the price is
+    ///         denominated in.
+    /// @return The normalized price with INTERNAL_DECIMALS precision.
+    function _normalizePrice(uint price_, uint8 tokenDecimals_)
```
omega-audits: Why do all this in the first place though? Price could just be defined in the collateral token, since this is what the price actually is (issuance and redemption are both denominated in the collateral token, ie. Buy 1 EUR for 1.1 USD, sell 1 EUR for 1.05 USD, both prices are denominated in USD, and it makes sense to just keep prices as USD all the way through).

src/external/token/ERC20Issuance_Blacklist_v1.sol
```diff
@@ -0,0 +1,212 @@
...
+    // -------------------------------------------------------------------------
+    // View Functions
+
+    /// @inheritdoc	IERC20Issuance_Blacklist_v1
+    function isBlacklisted(address account_) public view returns (bool) {
+        return _blacklist[account_];
+    }
+
+    /// @inheritdoc	IERC20Issuance_Blacklist_v1
+    function isBlacklistManager(address account_) public view returns (bool) {
```
omega-audits: Can be marked external. Or the mapping itself could just be public

src/external/token/ERC20Issuance_Blacklist_v1.sol
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /// @inheritdoc IERC20Issuance_Blacklist_v1
+    function removeFromBlacklist(address account_)
+        public
+        onlyBlacklistManager
+    {
+        if (account_ == address(0)) {
+            revert ERC20Issuance_Blacklist_ZeroAddress();
+        }
```
omega-audits: This will fail anyway in isBlacklisted since address 0 can't be blacklisted

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
```diff
@@ -0,0 +1,769 @@
...
+        external
+        onlyPaymentClient
+    {
+        token().safeTransfer(to_, amount_);
+
+        emit TransferOrchestratorToken(to_, amount_);
+    }
+
+    /// @inheritdoc IFM_PC_ExternalPrice_Redeeming_v1
+    function deductProcessedRedemptionAmount(uint processedRedemptionAmount_)
```
omega-audits: This does not seem to be implemented currently in any of the contracts

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
```diff
@@ -0,0 +1,769 @@
...
+    /// @dev    Maximum allowed project fee percentage for buying tokens.
+    uint private _maxBuyFee;
+
+    /// @notice Order ID counter for tracking individual orders.
+    /// @dev    Unique identifier for the current order being processed.
+    uint private _orderId;
+
+    /// @notice Counter for generating unique order IDs.
+    /// @dev    Keeps track of the next available order ID to ensure uniqueness.
+    uint private _nextOrderId;
```
omega-audits: Why do you need both? It seems they serve the same purpose and one can be known from the other.

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
```diff
@@ -0,0 +1,769 @@
...
+
+        // Update open redemption amount.
+        _addToOpenRedemptionAmount(collateralRedeemAmount_);
+
+        // Calculate redemption amount.
+        uint redemptionAmount_ = collateralRedeemAmount_ - issuanceFeeAmount_;
+
+        // Create and add payment order.
+        PaymentOrder memory order = PaymentOrder({
+            recipient: _msgSender(),
```
omega-audits: Recipient should be the receiver_ not msg.sender

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
```diff
@@ -0,0 +1,769 @@
...
+            receiver_,
+            depositAmount_,
+            _oracle.getPriceForRedemption(),
+            collateralRedeemAmount_,
+            sellFee,
+            issuanceFeeAmount_,
+            redemptionAmount_,
+            address(token()),
+            block.timestamp,
+            RedemptionState.PROCESSING
```
omega-audits: The other states don't seem to really be used

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
```diff
@@ -0,0 +1,769 @@
...
+        }
+
+        // Revert when the redeem amount is lower than minimum amount the user
+        // expects.
+        if (collateralRedeemAmount < _minAmountOut) {
+            revert Module__BondingCurveBase__InsufficientOutputAmount();
+        }
+
+        // Use virtual function to handle collateral tokens.
+        _handleCollateralTokensAfterSell(_receiver, collateralRedeemAmount);
```
omega-audits: You could just remove this call here

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
```diff
@@ -0,0 +1,769 @@
...
+        uint issuanceFeeAmount_
+    ) internal {
+        // Generate new order ID.
+        _orderId = _nextOrderId++;
+
+        // Update open redemption amount.
+        _addToOpenRedemptionAmount(collateralRedeemAmount_);
+
+        // Calculate redemption amount.
+        uint redemptionAmount_ = collateralRedeemAmount_ - issuanceFeeAmount_;
```
omega-audits: The collateralRedeemAmount_ is already post fee, which means this essentially deducts the fee twice and is wrong

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
```diff
@@ -0,0 +1,769 @@
...
+        // Require that enough collateral tokens are held to cover the project
+        // collateral fee.
+        if (
+            projectCollateralFeeCollected
+                > collateralToken.balanceOf(address(this))
+        ) {
+            revert
+                Module__RedeemingBondingCurveBase__InsufficientCollateralForProjectFee(
+            );
+        }
```
omega-audits: Could just remove this as it is unused in this implementation

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
```diff
@@ -0,0 +1,769 @@
...
+        if (collateralRedeemAmount < _minAmountOut) {
+            revert Module__BondingCurveBase__InsufficientOutputAmount();
+        }
+
+        // Use virtual function to handle collateral tokens.
+        _handleCollateralTokensAfterSell(_receiver, collateralRedeemAmount);
+
+        // Create and emit the order.
+        _createAndEmitOrder(
+            _receiver, _depositAmount, collateralRedeemAmount, issuanceFeeAmount
```
omega-audits: issuanceFeeAmount is just one of the fees charged, it's unclear why it does not pass all fees combined

src/external/token/ERC20Issuance_Blacklist_v1.sol
omega-audits: the whole zero address check here is a bit useless as it will not matter if the zero address is balcklisted or not

src/external/token/ERC20Issuance_Blacklist_v1.sol
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /// @inheritdoc IERC20Issuance_Blacklist_v1
+    function addToBlacklistBatched(address[] memory accounts_)
+        external
+        onlyBlacklistManager
+    {
+        uint totalAccounts = accounts_.length;
+        if (totalAccounts > BATCH_LIMIT) {
+            revert ERC20Issuance_Blacklist_BatchLimitExceeded(
```
omega-audits, line 147: what is the point of limiting the batch to this arbitrary number. If the manager wants to add 400 addresses, is that a problem?

src/external/token/ERC20Issuance_Blacklist_v1.sol
```diff
@@ -0,0 +1,212 @@
...
+        }
+    }
+
+    /// @inheritdoc IERC20Issuance_Blacklist_v1
+    function addToBlacklistBatched(address[] memory accounts_)
+        external
+        onlyBlacklistManager
+    {
+        uint totalAccounts = accounts_.length;
+        if (totalAccounts > BATCH_LIMIT) {
```
omega-audits, line 165: what is the purpose of this logic? Is there a problem i fthe blacklist manager adds move than BATCH_LIMIT addresses to the blacklist?

src/external/token/ERC20Issuance_Blacklist_v1.sol
```diff
@@ -0,0 +1,212 @@
...
+        if (isBlacklisted(account_)) {
+            _blacklist[account_] = false;
+            emit RemovedFromBlacklist(account_);
+        }
+    }
+
+    /// @inheritdoc IERC20Issuance_Blacklist_v1
+    function addToBlacklistBatched(address[] memory accounts_)
+        external
+        onlyBlacklistManager
```
omega-audits, line 162: can be removed, as addToBlacklist already has this check

src/external/token/ERC20Issuance_Blacklist_v1.sol
```diff
@@ -0,0 +1,212 @@
...
+        }
+    }
+
+    /// @inheritdoc IERC20Issuance_Blacklist_v1
+    function removeFromBlacklistBatched(address[] memory accounts_)
+        external
+        onlyBlacklistManager
+    {
+        uint totalAccounts = accounts_.length;
+        if (totalAccounts > BATCH_LIMIT) {
```
omega-audits, line 182: see comment above - it is unclear what the point of this extra check is

src/external/token/ERC20Issuance_Blacklist_v1.sol
```diff
@@ -0,0 +1,212 @@
...
+        }
+        for (uint i; i < totalAccounts; ++i) {
+            addToBlacklist(accounts_[i]);
+        }
+    }
+
+    /// @inheritdoc IERC20Issuance_Blacklist_v1
+    function removeFromBlacklistBatched(address[] memory accounts_)
+        external
+        onlyBlacklistManager
```
omega-audits, line 179: can be removed as it is already checkd in removeFromBlacklist

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
```diff
@@ -0,0 +1,769 @@
...
+
+    // -------------------------------------------------------------------------
+    // Public Functions
+
+    /// @inheritdoc BondingCurveBase_v1
+    function buy(uint collateralAmount_, uint minAmountOut_)
+        public
+        override(BondingCurveBase_v1)
+        onlyModuleRole(WHITELIST_ROLE)
+        buyingIsEnabled
```
omega-audits: buyingisenabled is also checked in super.buyFor, so it could be removed here

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
omega-audits: This makes the _openRedemptionAmount variable basically useless, so it is probably a bug

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
omega-audits: (jelle:) I thnk it is because the protocolFeeAmount, projectFeeAmount are taken from the collateral, and the issaunceFee from the issuance token.

src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol
```diff
@@ -0,0 +1,214 @@
...
+ *                          to our Security Policy at security.inverter.network
+ *                          or email us directly!
+ *
+ * @custom:version  v1.0.0
+ *
+ * @custom:standard-version v1.0.0
+ *
+ * @author  Zealynx Security
+ */
+contract LM_ManualExternalPriceSetter_v1 is
```
omega-audits: there seems to be no test touching this contract

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
```diff
@@ -0,0 +1,769 @@
+// SPDX-License-Identifier: LGPL-3.0-only
```
omega-audits: There are no tests covering this contract, and we believe we have found some errors that would definitely have been caught with a test.

src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol
vendrell46: I'm confident the implementation is correct as is. 

For redemption, we're pricing 'how many collateral tokens you get for 1 issuance token', so the input price should be in issuance token decimals, as stated in the documentation. This maintains consistency with the pricing model where 1 unit is always denominated in the issuance token.

src/external/token/ERC20Issuance_Blacklist_v1.sol
vendrell46: It's merely a DoS protection. While the chances are low, it is still worth having it. 

If a blacklist manager needs to add more than BATCH_LIMIT addresses, they can simply split it into multiple transactions.

src/external/token/ERC20Issuance_Blacklist_v1.sol
vendrell46: same as above

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
vendrell46: mainly we saw it as having two different purposes even though you are right that we could derive one from the other. But having them both we have that:

- `_orderId` represents the currently active/processing order.
- `_nextOrderId` is an incrementing counter that ensures each new order gets a unique ID.

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
vendrell46: from my point of view, we need it because the function is public and could be called directly.

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
vendrell46: good catch

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
vendrell46: It will be used when integrated with a different module

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
vendrell46: makes sense

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
vendrell46: true

src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol
Zitzak: Currently discussing internally. Will be following up on it

src/external/token/ERC20Issuance_Blacklist_v1.sol
Zitzak: Small context @omega-audits, we make all state variable private by default and add explicit getters for it. Code has been changed as you suggested

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
Zitzak: The FM_PC contract will be coupled with a queue based Payment Processor, which is almost up for review. This contract will be calling the function mentioned above. However I do see an issue with it now if a different PP is paired with this FM. Will be addressed differently

src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol
Zitzak: Feeds into the conversation about using the collateral price for all calculation [see comment](https://github.com/InverterNetwork/contracts/pull/704#discussion_r1897340470)

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
Zitzak: I see the point in regards to gas optimization. The `nextOrderId` will be removed

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
Zitzak: An indexer will be used to track state changes and redemption orders which will be visualized within our admin panel.

However the `block.timestamp` can probably be removed as recommended elsewhere, as those can be retrieved from the event

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
Zitzak: You are right (jelle), we charge different fees here. The project fee can only be charged from the collateral while the protocol fee can be charge from issuance and collateral.

The event emitting the fee amount in the `_createAndEmitOrder` function serves the purpose of giving the project information they need for their users, as well as their admin panel.

**However** this question did reveal to me that what we should have been passing into the function is not the `issuanceFeeAmount` (which is part of the protocol fee), but the `projectFeeAmount`. So turns out to still be a great catch @omega-audits

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
Zitzak: I have to agree here with @omega-audits, as calling this function directly, as you're mentioning, will direct to the base `buyFor` where the modifier is located.

Could you please expand a bit on your point @vendrell46. Thank you

src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol
Zitzak: Tests have been written but were on a different branch. We decided to move forward with the audit (without the tests being on the feature branch) due to the time crunch. However the tests will be in the feature branch for the second audit

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
Zitzak: See comments above. Nonetheless, some could have indeed be found in the testing that was done

src/external/token/ERC20Issuance_Blacklist_v1.sol
leeftk: I don't actually think this is necessary tbh, I believe this can be removed as the manager themselves can just submit  addresses in batches themselves if they end up hitting the gas limit with too many.

src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
leeftk:
```
    function buyFor(address _receiver, uint _depositAmount, uint _minAmountOut)
        public
        virtual
        override(BondingCurveBase_v1)
        validReceiver(_receiver)
        buyingIsEnabled
    {
        (uint amountIssued, uint collateralFeeAmount) =
            _buyOrder(_receiver, _depositAmount, _minAmountOut);
        _addVirtualIssuanceAmount(amountIssued);
        _addVirtualCollateralAmount(_depositAmount - collateralFeeAmount);
    }
    ```
    
    This is correct this is redundant and should be removed.

File: src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol (hunk 6 above)
Comment (Zitzak) [reply src/modules/logicModule/LM_ManualExternalPriceSetter_v1.sol]: Just finding out that this could be triggered in the situation where the contract has been initialized, but no price has been set yet. Then it would revert instead of returning 0. Is this something we should keep, or maybe enforce setting an initial price when initializing the module. Wdyt? @leeftk @vendrell46

File: src/external/token/ERC20Issuance_Blacklist_v1.sol (hunk 19 above)
Comment (omega-audits) [reply src/external/token/ERC20Issuance_Blacklist_v1.sol:165]: Yes that is what we'd recommend

File: src/modules/paymentProcessor/PP_Queue_v1.sol [hunk 26]
```diff
@@ -0,0 +1,952 @@
...
+        return data_.length == requiredDataLength
+            && (flagsValue & (1 << FLAG_ORDER_ID)) != 0;
+    }
+
+    /// @notice Internal function to check whether the client is valid.
+    /// @param  client_ Address to validate.
+    function _ensureValidClient(address client_) internal view virtual {
+        if (client_ == address(0)) {
+            revert Module__PP_Queue_InvalidClientAddress(client_);
+        }
```
omega-audits: If client has to be the _msgSender it will never be address(0) anyway (and if it is it'll revert)

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
```diff
@@ -0,0 +1,1040 @@
...
+            targetChainId: block.chainid,
+            flags: flags,
+            data: data
+        });
+
+        // Add order to payment client.
+        _addPaymentOrder(order);
+
+        // Process payments through the payment processor.
+        __Module_orchestrator.paymentProcessor().processPayments(
```
omega-audits: The benefit of a queue is to execute multiple orders at once, but this will just process each order when it's added, which makes the whole queue logic unused.

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,952 @@
...
+
+        (orders,,) = client_.collectPaymentOrders();
+
+        uint orderLength = orders.length;
+        for (uint i; i < orderLength; ++i) {
+            // Add order to order queue
+            _addPaymentOrderToQueue(orders[i], address(client_));
+        }
+        // Execute Order Queue.
+        _executePaymentQueue(address(client_));
```
omega-audits, line 389: The benefit of a queue is to execute multiple orders at once, but this will execute the orders as soon as they are added to the queue. It would make more sense to separate the execution logic from adding to the queue

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,952 @@
...
+        virtual
+        clientIsValid(client_)
+    {
+        uint firstId = _queue[client_].getNextId(LinkedIdList._SENTINEL);
+        if (firstId == LinkedIdList._SENTINEL) {
+            revert Module__PP_Queue_EmptyQueue();
+        }
+
+        uint processedCount;
+        while (firstId != LinkedIdList._SENTINEL && _processNextOrder(client_))
```
omega-audits: If firstId != LinkedIdList._SENTINEL then _processNextOrder will return false, so the entire tracking here of firstId is essentially unnecessary, the function could be written as:
```
        uint processedCount;
        while (_processNextOrder(client_))
        {
            ++processedCount;
        }

        emit PaymentQueueExecuted(_msgSender(), client_, processedCount);
```

omega-audits: Another point, _processNextOrder may return false in various cases, but not in all of them it's needed to stop the execution of the rest of the queue. For example, if the payment fails for a specific order, it doesn't mean you should stop here and not execute the rest of the queue. Or if for example there's an order which is too big and cannot be paid at the moment, this will cause also smaller orders further down the queue to not be paid (which may or may not be the intention here). You should consider continuing to execute orders in the queue if there's no reason to stop for all.

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,952 @@
...
+        internal
+        view
+        virtual
+        returns (bool validPaymentReceiver_)
+    {
+        return !(
+            receiver_ == address(0) || receiver_ == _msgSender()
+                || receiver_ == address(this)
+                || receiver_ == address(orchestrator())
+                || receiver_ == address(orchestrator().fundingManager().token())
```
omega-audits, line 857: This is essentially making an external call for each time you validate. If this restriction doesn't have a specific important reason I'd suggest to remove it. And if it does I'd recommend to save the token in storage here instead of reading it every time.

src/modules/paymentProcessor/PP_Queue_v1.sol
omega-audits: This is especially true since there can be different payment tokens for different orders (in theory also if from the same client)

omega-audits: And another point, if the execution halts due to any of the reasons for _processNextOrder to return false, there's not option to continue the execution until the client processes more payments. It would make sense to allow users to also trigger payments as there seems to be no reason to force them to wait from new payments to be processed in order to be able to retry processing their own

src/modules/paymentProcessor/PP_Queue_ManualExecution_v1.sol
```diff
@@ -0,0 +1,101 @@
...
+            // Add order to order queue.
+            _addPaymentOrderToQueue(orders[i], address(client_));
+        }
+    }
+
+    /// @inheritdoc IPP_Queue_ManualExecution_v1
+    function executePaymentQueue(IERC20PaymentClientBase_v2 client_)
+        external
+        virtual
+        clientIsValid(address(client_))
```
omega-audits: _executePaymentQueue already checks clientIsValid. Though it's unclear to me why would you require the client itself only to trigger its payment and not allow it to be triggered also externally. If a user is willing to pay the gas for the execution, I see no security reason to prevent them from calling execute (assuming proper validation of the client being called)

src/modules/paymentProcessor/PP_Queue_ManualExecution_v1.sol
```diff
@@ -0,0 +1,101 @@
...
+ *                          our Security Policy at security.inverter.network or
+ *                          email us directly!
+ *
+ * @custom:version  v1.0.0
+ *
+ * @custom:standard-version v1.0.0
+ *
+ * @author  Zealynx Security
+ */
+contract PP_Queue_ManualExecution_v1 is
```
omega-audits, line 98: For a payment queue, it seems to me more logical to have the default contract be manual execution, then have an extension contract offering automatic execution. That would both save code duplication you have in processPayments, and would make more sense in the context of queue, as the goal is usually to wait and execute multiple payments at once.

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
```diff
@@ -0,0 +1,1040 @@
...
+            _orderId,
+            _msgSender(),
+            receiver_,
+            depositAmount_,
+            _oracle.getPriceForRedemption(),
+            sellFee,
+            projectCollateralFeeAmount_,
+            collateralRedeemAmount_,
+            address(token()),
+            RedemptionState.PENDING
```
omega-audits, line 744: Depending on the payment processor, the previous call to `processPayments` may have executed the order (and emitted a correspoding event with a PROCESSED redemption state). It is better to emit this even t _before_ execution

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
```diff
@@ -0,0 +1,1040 @@
...
+        virtual
+        onlyModuleRole(QUEUE_EXECUTOR_ROLE)
+    {
+        (bool success,) = address(__Module_orchestrator.paymentProcessor()).call(
+            abi.encodeWithSignature(
+                "executePaymentQueue(address)", address(this)
+            )
+        );
+        if (!success) {
+            revert Module__FM_PC_ExternalPrice_Redeeming_QueueExecutionFailed();
```
omega-audits: it is never a good idea to suppress error messages - you probably want to include the reason for the revert in the Error in some way

Zitzak:
@Zitzak 
- Remove `setSellFee` public function as we don't add any feature when overriding
- Remove `setBuyFee` public function as we don't add any feature when overriding

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
Zitzak: As we don't make a hard coupling between specific module, i.e. this FM and the Queue base PP, we decided to built the FM in a way that it would also work if someone would couple for example the PP_Simple_v1. For that reason, we call process payments, and if it isn't the `PP_Queue_ManualExecution_v1`, the PP would just go ahead and process each payment order. Hope this clarifies it. @omega-audits

src/modules/paymentProcessor/PP_Queue_v1.sol
Zitzak: I let @vendrell46 react to this as well in regard to your first comment. I have a few points to add to your comment:
- The Queue is first in, first out. So if the queue stops due to an order being too big then that is expected behaviour
- You are right that the `_processNextOrder` function would stop when an order fails. The possibility of an order failing has been added later to the contract, and apparently the tests have not caught this sadly.
- You are right in regard to the case that a client could have different payment tokens. Could you give a suggestions how you would stick to the "first in, first out" requirements while still accounting for multiple tokens? 
- Given the issue with the queue execution stopping when a transfer has failed is resolved we should be good in respect to the queue execution. The workflow owners should have the control over when to execute the queue
@omega-audits

src/modules/paymentProcessor/PP_Queue_v1.sol
Zitzak: You are right here, however this is how we've done it in the other PPs as well, and we aim to be on networks with low gas costs. However going to bring this up internally. Thank you

src/modules/paymentProcessor/PP_Queue_ManualExecution_v1.sol
Zitzak: Agree on the `clientIsValid` check. Thank you. In regard to the question around why it can only be triggered through the client; the manual queue PP is coupled with a FM which may or may not hold assets to execute the redemption queue. The general flow would look something like this:
1. Tokens are minted in the FM, collateral withdrawn
2. Issued tokens are redeemed, but instead of a direct payout, a redemption queue order is created in the PP
3. Collateral reserve is deposited back into the FM (which also is the Payment Client) in preparation to executing the  queue
4. The queue is manually executed through the FM function `executeRedemptionQueue()` after which the collateral is pulled from the FM and processed for each order

Hope this clarifies some of the thinking behind the setup @omega-audits

src/modules/paymentProcessor/PP_Queue_ManualExecution_v1.sol
Zitzak: So.. I think that makes sense in hindsight to do the naming and "default" queue based PP manual. I will check if it makes sense for us now to rename everything as the client has already started testing, and it has been integrated already in the Indexer etc. However fair point. 

In terms of code duplication, I would suspect that it is about the same

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
Zitzak: Good point! We will move it. Thank you

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
Zitzak: The idea here was that the only reason it would fail is because the contract being called doesn't have the correct function. However we can add the `bytes data` from the call to the custom error

src/modules/paymentProcessor/PP_Queue_v1.sol
vendrell46: > If firstId != LinkedIdList._SENTINEL then _processNextOrder will return false, so the entire tracking here of firstId is essentially unnecessary

You are absolutely right, good catch. I will modify it as you suggested, @omega-audits

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
jellegerbrandy: So this is of course not a bug, but our observation here is that if the PP has a queue, this queue is not used if you call processPrayments each time you add a payment order. In other words, our point is that this is not a good use of the queue-based PP.

jellegerbrandy: That said, if this is clear to all of us, we can close this comment :)

src/modules/paymentProcessor/PP_Queue_ManualExecution_v1.sol
omega-audits: In this case though, why does it try to immediately execute each payment added to the queue (as mentioned in another comment above here)?

src/modules/paymentProcessor/PP_Queue_v1.sol
Zitzak: We have done this with the manual version ser. Assuming that this was commented before you've seen that version :)

src/modules/paymentProcessor/PP_Queue_v1.sol
Zitzak: Has been addressed

src/modules/paymentProcessor/PP_Queue_ManualExecution_v1.sol
Zitzak: I believe you confuse the manual contract with the automatic one @omega-audits? The manual version which is referenced in the comment does not execute the payments right away

src/modules/paymentProcessor/PP_Queue_ManualExecution_v1.sol
Zitzak: We will leave the name as is for now ser @omega-audits

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
Zitzak: Has been addressed

src/modules/paymentProcessor/PP_Queue_v1.sol
Zitzak: We will leave it as is for now ser

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
Zitzak: has been addressed

src/external/token/ERC20IssuanceUpgradeable_Blacklist_v1.sol
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    // -------------------------------------------------------------------------
+    // External Functions
+
+    /// @inheritdoc IERC20Issuance_Blacklist_v1
+    function addToBlacklist(address account_)
+        public
+        virtual
+        onlyBlacklistManager
```
omega-audits, line 151: It could be slightly more gas efficient to have an internal _ addToBlacklist without the modifier and use it both here and in the batched version, and add the onlyBlacklistManager to the batched function, so that the modifier is only executed once and not on each iteration. Same for the remove logic.

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,1066 @@
...
+        if (protocolFeePercentage >= BPS) {
+            revert Module__PP_Queue_FeeAmountToHigh(protocolFeePercentage);
+        }
+
+        // Calculate protocol fee amount if applicable
+        if (protocolFeePercentage > 0) {
+            feeAmount_ = totalAmount_ * protocolFeePercentage / BPS;
+            // Revert if calculated protocol fee amount rounded down to zero
+            if (feeAmount_ == 0) {
+                revert Module__PP_Queue_InvalidFeeAmount(feeAmount_);
```
omega-audits: Are you sure you'd like to revert here in this case? 

If a payment order was for a very small amount, this could get the whole queue completely stuck, also for removing the payment which also uses this function (which is another point, when this function is used in cancelling a payment, there's no reason to take fees)

omega-audits: So I'd recommend:
1. Remove this revert check entirely
2. When cancelling a payment, pass a flag to the  tryPaymentTransfer to avoid calling this function

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
```diff
@@ -0,0 +1,1119 @@
...
+            bytes32[] memory paymentParameters = new bytes32[](1);
+            paymentParameters[0] = bytes32(_orderId);
+
+            (flags, data) = _assemblePaymentConfig(paymentParameters);
+        }
+        // Create and add payment order.
+        PaymentOrder memory order = PaymentOrder({
+            recipient: receiver_,
+            paymentToken: address(token()),
+            amount: netCollateralRedeemAmount_ + protocolCollateralFeeAmount_,
```
omega-audits, line 724: So the Payment Processor will now be recalculating the fee, but using the redeemAmount + protocol fee as the total amount, instead of redeemAmount + protocol fee + project fee.

This will result in a potential slight underpayment of the fees.

src/modules/paymentProcessor/PP_Queue_v1.sol
Zitzak: Good point ser @omega-audits. I'll remove the revert and implement the flag as suggested

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
Zitzak: The project fee is never deposited back into the FM as total redeem amount for executing the queue. Because of that, we don't have the project here in the payment order amount ser @omega-audits . Does that make sense?

omega-audits: No but what I mean is that the protocol fee is calculated here based on net redeem amount + protocol fee + project fee. So if total amount is 100, and the fees are 2 percent each, it'll pass 98 as the amount to the payment order, which is fine, but the problem is then that the payment processor is re calculating the protocol fee, but it'll do it as 2% of 98, instead of 2% of 100.

Zitzak: Makes sense, thank you

src/external/token/ERC20IssuanceUpgradeable_Blacklist_v1.sol
Zitzak: Agree

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
Zitzak: Discussed the issue internally and understood that a similar problem surfaced not long ago when dealing with a different client project, i.e. the fee being calculated twice in different locations which results in an underpayment of fee. The solution decided to move forward with in that case was to adjust the fee percentage upwards such that the fee amount collected would be as expected.

As we are currently working on a new FeeManager contract, which will enable a better mechanism of collecting fee across the workflows, we decided to go with the same manual adjustment approach for now.

We're aware that this solution is a temporary, manual bandaid and not an actual solution. However as no client or token receiver is negatively impacted by the current setup, and we are currently working on the new FeeManager and there is no clean solution we decided to move ahead with the way it is now. @omega-audits

src/external/token/ERC20Issuance_Blacklist_v1.sol
```diff
@@ -254,4 +247,32 @@ contract ERC20Issuance_Blacklist_v1 is
...
+    }
+
+    /// @notice Internal function to remove an address from the blacklist.
+    /// @dev    This function only updates the blacklist state if the address is
+    ///         currently blacklisted to prevent unnecessary state changes.
+    /// @param  account_ Address to remove from the blacklist.
+    function _removeFromBlacklist(address account_) internal virtual {
+        if (account_ == address(0)) {
+            revert ERC20Issuance_Blacklist_ZeroAddress();
+        }
```
omega-audits: This is unnecessary since the add already checks so if account passed here is 0 nothing will happen

omega-audits (approved):

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,1081 @@
...
+        }
+
+        // Update the order state to CANCELLED.
+        _updateOrderState(orderId_, address(client_), RedemptionState.CANCELLED);
+
+        // Remove the cancelled order from the queue.
+        _removeFromQueue(orderId_, address(client_));
+
+        // Try to transfer the amount to the treasury.
+        success_ = _tryPaymentTransfer(
```
jellegerbrandy, line 480: The state of the order will be canceled and the order removed from the queue also if the payment failed. Is that desired behavior?

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,1081 @@
...
+    /// @param	amount_ The amount to transfer.
+    /// @return	success_ True if the transfer was successful.
+    function _lowLevelTransfer(
+        address token_,
+        address client_,
+        address recipient_,
+        uint amount_
+    ) internal virtual returns (bool success_) {
+        // Make a low-level call to the token contract to execute transferFrom
+        (bool success, bytes memory data) = token_.call(
```
jellegerbrandy, line 586: A `try... catch.. ` pattern would do the same thing, but you do not need to parse the resuts of a low-level call yourself

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,1081 @@
...
+    {
+        uint firstId = _queue[client_].getNextId(LinkedIdList._SENTINEL);
+        if (firstId == LinkedIdList._SENTINEL) {
+            return false;
+        }
+
+        QueuedOrder storage order = _orders[client_][firstId];
+
+        // Skip if order is not in PENDING state.
+        if (order.state_ != RedemptionState.PENDING) {
```
jellegerbrandy, line 517: 1. afaics thsi should never happen, as each time the state is changed from pending, the order is removed from the queue. So perhaps not fail silently here? 
2. If I am wrong and this happens, why return `false`? This will cause the loop in _executePaymentQueue to exit, and so the operator will need to call processPayments a second time to process the rest of the queue.

src/modules/paymentProcessor/PP_Queue_v1.sol
Zitzak: Good point. In practise this should never 'fail' as we're sending it to project owners canceled orders treasury. Meaning that this address would have to be blacklisted by the project for it to fail. However it makes sense to add a revert in case this would happen anyway

Zitzak: Good point. In practise this should never 'fail' as we're sending it to project owners canceled orders treasury. Meaning that this address would have to be blacklisted by the project for it to fail. However it makes sense to add a revert in case this would happen anyway

src/modules/paymentProcessor/PP_Queue_v1.sol
Zitzak: So it's also being tested with an address which doesn't have the transfer interface. I tried all kind of ways with `try/catch` but I didn't get it to work with the EVM revert which resulted from it. Eventually a colleague found this explanation for it not working, so I will stick with the low level call for now :)
image (3)

src/modules/paymentProcessor/PP_Queue_v1.sol
Zitzak: You're right it shouldn't happen, so going to revert instead of Fail here

jellegerbrandy (approved):
lgtm!

.vscode/settings.json
FHieser, line 1: Are these really part of the PR?

src/modules/logicModule/LM_Oracle_Permissioned_v1.sol
```diff
@@ -0,0 +1,244 @@
...
+    /// @notice Internal function to set the redemption price
+    /// @param price_ The price to set
+    function _setRedemptionPrice(uint price_) internal virtual {
+        if (price_ == 0) revert Module__LM_ExternalPriceSetter__InvalidPrice();
+        _redemptionPrice = price_;
+        emit RedemptionPriceSet(price_, _msgSender());
+    }
+
+    /// @dev    Storage gap for upgradeable contracts.
+    uint[50] private __gap;
```
FHieser: Why is this gap at the end of the file?
As it is a state modification I would argue that it fits better at the end of the State section

src/modules/logicModule/LM_Oracle_Permissioned_v1.sol
```diff
@@ -0,0 +1,244 @@
...
+    function setIssuanceAndRedemptionPrice(
+        uint issuancePrice_,
+        uint redemptionPrice_
+    ) external virtual onlyModuleRole(PRICE_SETTER_ROLE) {
+        _setIssuancePrice(issuancePrice_);
+        _setRedemptionPrice(redemptionPrice_);
+    }
+
+    /// @inheritdoc ILM_Oracle_Permissioned_v1
+    function getCollateralTokenDecimals()
```
FHieser, line 168: Would love to see a separation between Getter and Mutating Functions here

FHieser: Getter being first and then Mutating

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
```diff
@@ -0,0 +1,1119 @@
...
+    function setOracleAddress(address oracle_)
+        external
+        virtual
+        onlyOrchestratorAdmin
+    {
+        _setOracleAddress(oracle_);
+    }
+
+    /// @inheritdoc IFM_PC_Oracle_Redeeming_v1
+    function getBuyFee() public view virtual returns (uint buyFee_) {
```
FHieser: Seperation from view and not view functions section please

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
```diff
@@ -0,0 +1,1119 @@
...
+        super.sellTo(receiver_, depositAmount_, minAmountOut_);
+    }
+
+    /// @inheritdoc IERC20PaymentClientBase_v2
+    function amountPaid(address token_, uint amount_)
+        public
+        virtual
+        override(ERC20PaymentClientBase_v2, IERC20PaymentClientBase_v2)
+    {
+        _deductFromOpenRedemptionAmount(amount_);
```
FHieser, line 610: Can you explain to me what exactly is the difference between Amount paid and Open Redemption amount?

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
```diff
@@ -0,0 +1,1119 @@
...
+            processPaymentsSelector
+        );
+
+        // Get issuance fee and treasury addresses from fee manager.
+        (issuanceFeePercentage_, issuanceTreasury_) =
+            _getFeeManagerIssuanceFeeData(selector_);
+    }
+
+    /// @dev    Storage gap for future upgrades.
+    uint[50] private __gap;
```
FHieser: Gap here

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
FHieser, line 1: I think I would have loved to see a general implementation here first (only the Oracle Functionality)
And then a more specific solution, kinda like we did it for Topos.
I know that is to late for this implementation now, but I think it would be great if you guys could write an outlier of which functionalities could be abstracted into the general and specific implementation respectively, to make it easier at a later point of time to get into it again

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
FHieser: I did not dig into this fully, but my first perception is, that they could be the same value essentially

FHieser (commented):
It took me more time to review this than anticipated.
Will give my feedback for the Queue PP tomorrow

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
leeftk: I think this is great feedback for the modules that we are currently building if I understood correctly. I'm constantly trying to figure out what can be abstracted during development but would be nice to review this beforehand so we go into the work already thinking of the abstracting away as much as possible.

src/modules/logicModule/LM_Oracle_Permissioned_v1.sol
leeftk: For some reason these keep getting added, maybe we need to add this to the .gitignore, but we had the same issue on the other module. Will look for @zitzak guidance on this as it seems to be recurring.

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
leeftk: Checked the OZ docs and @FHieser is correct this can be placed with the other storage variables in the file though I have seen a lot of projects place it at the end like this, not sure why though as there's no specific reason

leeftk: Here's an example.
https://github.com/forta-network/forta-contracts/blob/ad2a47a28cf0dd1913caf28b0f498f84aceec34e/contracts/vesting/VestingWalletV1.sol

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,1089 @@
...
+        external
+        view
+        virtual
+        clientIsValid(address(client_))
+    {
+        return;
+    }
+
+    /// @inheritdoc IPaymentProcessor_v2
+    function unclaimable(
```
FHieser: View Function in Mutating Section

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,1089 @@
...
+    ) external virtual {
+        if (unclaimable(client_, token_, receiver_) == 0) {
+            revert Module__PaymentProcessor__NothingToClaim(client_, receiver_);
+        }
+
+        _claimPreviouslyUnclaimable(client_, token_, receiver_);
+    }
+
+    /// @inheritdoc IPaymentProcessor_v2
+    function validPaymentOrder(
```
FHieser: same here

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,1089 @@
...
+        }
+
+        // Update the order state to CANCELLED.
+        _updateOrderState(orderId_, address(client_), RedemptionState.CANCELLED);
+
+        // Remove the cancelled order from the queue.
+        _removeFromQueue(orderId_, address(client_));
+
+        // Try to transfer the amount to the treasury.
+        success_ = _tryPaymentTransfer(
```
FHieser, line 480: There is an argument that could be made here, for actually taking a protocol fee here, as tokens get actually routed through the workflow itself and land in a private treasury. Do you have any opinions of this @Zitzak @marvinkruse @0xNuggan

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,1089 @@
...
+        if (protocolFeePercentage > 0) {
+            feeAmount_ = totalAmount_ * protocolFeePercentage / BPS;
+        }
+
+        // Calculate the net amount after deducting the protocol fee.
+        netAmount_ = totalAmount_ - feeAmount_;
+    }
+
+    /// @dev    Gap for possible future upgrades.
+    uint[50] private __gap;
```
FHieser: Gap

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,1089 @@
...
+            revert Module__PP_Queue_OnlyCallableByClient();
+        }
+    }
+
+    /// @notice Adds to the unclaimable amount for a specific payment.
+    /// @param  client_ The client address.
+    /// @param  token_ The token address.
+    /// @param  receiver_ The receiver address.
+    /// @param  amount_ The amount to add.
+    function _addToUnclaimableAmount(
```
FHieser, line 1002: Is this function used anywhere?

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,1089 @@
...
+        address client_,
+        address token_,
+        address paymentReceiver_
+    ) public view virtual returns (uint amount_) {
+        amount_ =
+            _unclaimableAmountsForRecipient[client_][token_][paymentReceiver_];
+    }
+
+    /// @inheritdoc IPaymentProcessor_v2
+    function claimPreviouslyUnclaimable(
```
FHieser, line 403: Am I misreading things or is the accesscontrol here missing?
Shouldnt it be reveiver = _msgSender()?

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,1089 @@
...
+        virtual
+        returns (uint queueId_)
+    {
+        // Check if orderID flag is set (bit 0)
+        bool hasOrderId = uint(flags_) & (1 << FLAG_ORDER_ID) != 0;
+
+        // If flag is set and data is provided, use that ID
+        if (hasOrderId && data_.length > FLAG_ORDER_ID) {
+            queueId_ = uint(data_[FLAG_ORDER_ID]);
+        }
```
FHieser, line 828: If it doesnt have the correct flags, then it always returns queueId of 0, which can result in multiple payment orders having the same id. Does that pose a problem?
What happens when a queue Id is duplicated?

src/modules/paymentProcessor/PP_Queue_v1.sol
```diff
@@ -0,0 +1,1089 @@
...
+            return true;
+        } catch {
+            return false;
+        }
+    }
+
+    /// @notice Validates a payment order.
+    /// @param  order_ The order to validate.
+    /// @return valid_ True if the order is valid.
+    function _validPaymentOrder(
```
FHieser, line 898: For later reference, as its too late for that now, but:
For clean code purposes I think the _validPaymentOrder function and most of the sub functions in here could probably be moved into a abstract PaymentProcessor contract as they provide a general solution in most cases. Examples are:
- _validTotalamount
- _validChainId
- _validPaymentToken

You could also just call super._validPaymentOrder and && the paymentQueueId here into the bool operation.
@leeftk Dont know your work for the next project, but if you guys are working on another PP then this might be something that we could adapt in the future
@marvinkruse You think thats viable here? I think that could be worth to create an issue for that.

src/modules/paymentProcessor/PP_Queue_ManualExecution_v1.sol
```diff
@@ -0,0 +1,152 @@
...
+    function executePaymentQueue(IERC20PaymentClientBase_v2 client_)
+        external
+        virtual
+        onlyModule
+    {
+        _executePaymentQueue(address(client_));
+    }
+
+    /// @dev    Gap for possible future upgrades.
+    uint[50] private __gap;
```
FHieser: Gap

FHieser (changes requested):
See comments

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
FHieser: Yeah its a learning process, some things only make sense when there was some substantial development in the first place and then its hard to look at it from a broader viewpoint.
If you want some feedback on the upcoming project feel free to just reach out to me :D

src/modules/logicModule/LM_Oracle_Permissioned_v1.sol
Zitzak: @leeftk your message refers to the `settings.json` I believe?

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
Zitzak: I'm not sure if it makes sense to split the implementation up more then currently is, as the oracle functionality heavily uses the bonding curve base setup i.e. we are using the base and changing the formula wrapper functions to call an oracle rather then a formula contract. Practically speaking this means a setter & getter and the wrapper functions which creating an extra base contract might be overkill.

In general of course I agree with the statement of abstraction

.vscode/settings.json
Zitzak: Not sure how they sneak in everytime. Deleted

src/modules/paymentProcessor/PP_Queue_v1.sol
Zitzak: So we call `_validPaymentOrder()` before adding a payment order. Within the check we call `_validQueueId()` with the queueID that is returned from this function. It checks the order or queue IDs, ensuring no IDs can be duplicated

src/modules/paymentProcessor/PP_Queue_v1.sol
Zitzak: Agree, we should move towards building an abstract PP that all implementations of the PP can inherit from

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
Zitzak: So looked into it and indeed the Amount Paid value could have been used for it. Overlooked it. However we did add events that are emitted when the open redemption amount is added/deducted, which are picked up by the indexer and platform. This is a feature which the underlying Amount Paid did not have.

Due to that reason I would not remove the current setup as the emitted event are needed, and adding them to the base contract is not straight forward. Wdyt @FHieser ?

src/modules/paymentProcessor/PP_Queue_v1.sol
marvinkruse: I remember that you brought this up to me in a huddle. I am not sure whether there are any more info in terms of what is going to happen with cancelled payments, but overall I think I agree. As far as our system is concerned, the payment is done, and the reason of failure is most likely not on our side. What do you think @0xNuggan?

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
FHieser: There is a right and a fast answer here:
The right one is to rework it, the fast one is you dont, as the only thing it does is "duplicate" code
Create a issue for it and this should be fine

src/modules/paymentProcessor/PP_Queue_v1.sol
0xNuggan: Discussed this with Ata: For now we are going with no fee on canceled or failed transfers. If we see projects exploiting this in the future, we can reassess.

FHieser (approved):

PR #711, opened 2025-01-14
Fix: Replace address in event

marvinkruse:
Updated the emission of the `ProtocolFeeMinted` event in `BondingCurveBase_v1` to correctly return the issuance token’s address instead of `address(this)`, which likely was a leftover from when the funding manager itself was the token.

Also updated the test, so this should be fine now.


0xNuggan (approved):

Zitzak:
Can this be merged? @marvinkruse

PR #712, opened 2025-01-15
 Fixes for [Audit] FM_PC_ExternalPrice_Redeeming, LM_ManualExternalPriceSetter_v1 

Zitzak:


src/modules/fundingManager/oracle/FM_PC_ExternalPrice_Redeeming_v1.sol
```diff
@@ -494,14 +492,18 @@ contract FM_PC_ExternalPrice_Redeeming_v1 is
...
         _addPaymentOrder(order);
 
         // Process payments through the payment processor.
         __Module_orchestrator.paymentProcessor().processPayments(
             IERC20PaymentClientBase_v1(address(this))
         );
 
-        // Emit event with all order details.
+        // Emit event with contract address that created the order and with the order ID.
+        emit PaymentOrderCreated(address(this), _orderId);
```
Zitzak: This should be included in the redemptionOrderCreated event. I don't think we need 2 events

PR #717, opened 2025-02-10
[Post-Audit] Revised Payment Order Struct

0xNuggan:
PR to merge the new PaymentOrder struct. 

Still missing:
- [x] Rename all contracts with interface changes to V2, + their tests


marvinkruse:
@0xNuggan Resolved the versioning issues. There are a few things missing in general that we do not solve here (i.e. Orchestrator technically now only supports the v2, not v1 anymore, but we will address that in another PR anyways) - all not relevant for the current development and more relevant for deployments etc. which I'll handle.

To get general compatibility I added the `PaymentProcessor_v1` and `PaymentProcessor_v2` distinction as well as the same of the ERC20 Payment Client and specifically allowed both in the `Module_v1`. 

Also merged the latest `dev` changes into this. Does that look good to you?

(Pinging @Zitzak for visibility to see this progress - can be merged if @0xNuggan is happy)

marvinkruse (approved):

Zitzak:
@marvinkruse thank you for the update. If you haven't already, could you make sure that we have a Linear issue tracking the changes you describe which will be addressed through a different PR. Thank you ser

PR #722, opened 2025-02-20
docs: setup module/contract docs folder

Zitzak:
## What has been done?
- folder directory which mirros src is created
- template directory got added with a module documentation template
- .gitkeep's are added to commit the empty folders

## Reason for setting this up
- Needed now as the external devs want to commit their docs
- Decided in the [code of conduct](https://docs.google.com/document/d/10jYXETE6ySx1e1yLOSY7Ccfy52Mfee1OJUEqGv-_wSA/edit?tab=t.0#heading=h.11u2ti36ttzp)


FHieser (approved):
Works for me
Dont know if we need the full src folder structure from the beginning.
If we do that we might as well create Placeholder contracts for each contract in src, so that we can reference them properly.
We might wanna do that as one of the next tasks actually.

PR #723, opened 2025-02-25
Chore: Revert forge-std upgrade

marvinkruse:
The last commit on `dev` falsely updated the `forge-std` dependency - rolling this back as that was not intentional.

Commit of `forge-std` should be `fc560fa`.


Zitzak (approved):
LGTM

PR #727, opened 2025-03-10
Feat: Funding Pot V1 -  Initial Review

zzzuhaibmohd:
### Overview
This pull request is for the first review of the `Funding_Pot_v1` contracts after cloning from the template and adding Funding Pot Admin Authorization implementation.

### Key Changes
- **Admin Authorization:** Added logic to allow only the orchestrator admin to grant/revoke the `FUNDING_POT_ADMIN_ROLE`.
- **Compilation Fixes:** Resolved various compilation issues that arose during the cloning of the template.
- **Unit Tests:**
  - Added unit fuzz tests for granting and revoking the `FUNDING_POT_ADMIN_ROLE`.
  - Added unit test for `exposed_` function to validate internal function behavior.
- **Code Cleanup:** Removed unnecessary state variables and outdated templates to streamline the codebase.

### Notable Commits
- **Mar 6, 2025:** Funding pot admin authorization.
- **Mar 7, 2025:** Added tests related to grant/revoke role functionality.

### Focus Areas
- Provide feedback on implementation, unit tests and identifying any blind spots or areas where we can improve.

### DIFF
--- a/src/modules/logicModule/LM_PC_FundingPot_v1.sol
+++ b/src/modules/logicModule/LM_PC_FundingPot_v1.sol
@@ -23,29 +23,6 @@ import {ERC165Upgradeable} from
 import {ILM_PC_FundingPot_v1} from
     "src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol";
 
-/**
- * @title   Inverter Template Logic Module Payment Client
- *
- * @notice  A template logic module payment client that handles deposits and payment processing.
- *          Users can deposit tokens up to a maximum amount, and authorized admins can process
- *          these deposits into payment orders.
- *
- * @dev     This contract implements the following key functionality:
- *          - Deposit handling with maximum amount validation
- *          - Payment order creation and processing through the Orchestrator
- *          - Role-based access control for deposit processing
- *          - ERC20 token integration with SafeERC20
- *          - Interface compliance checks via ERC165
- *
- * @custom:security-contact security@inverter.network
- *                          In case of any concerns or findings, please refer
- *                          to our Security Policy at security.inverter.network
- *                          or email us directly!
- *
- * @custom:version 1.0.0
- *
- * @author  Inverter Network
- */
 contract LM_PC_FundingPot_v1 is
     ILM_PC_FundingPot_v1,
     ERC20PaymentClientBase_v2
@@ -73,33 +50,14 @@ contract LM_PC_FundingPot_v1 is
     //--------------------------------------------------------------------------
     // Constants
 
-    uint internal constant _maxDepositAmount = 100 ether;
-
     //--------------------------------------------------------------------------
     // State
-
-    /// @dev The role that allows processing deposits
-    bytes32 public constant DEPOSIT_ADMIN_ROLE = "DEPOSIT_ADMIN";
 
     /// @dev The role for the funding pot admin.
     bytes32 public constant FUNDING_POT_ADMIN_ROLE = "FUNDING_POT_ADMIN";
 
-    /// @notice    Mapping of user addresses to their deposited token amounts.
-    mapping(address user => uint amount) internal _depositedAmounts;
-
-    /// @notice    Payment token.
-    IERC20 internal _paymentToken;
-
     /// @notice    Storage gap for future upgrades.
     uint[50] private __gap;
-
-    // =========================================================================
-    // Modifiers
-
-    modifier onlyValidDepositAmount(uint amount_) {
-        _ensureValidDepositAmount(amount_);
-        _;
-    }
 
     // =========================================================================
     // Constructor & Init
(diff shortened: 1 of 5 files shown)


test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
```diff
@@ -0,0 +1,177 @@
...
+ */
+contract LM_PC_FundingPot_v1_Test is ModuleTest {
+    // =========================================================================
+    // State
+
+    // SuT
+    LM_PC_FundingPot_v1_Exposed paymentClient;
+
+    // Mocks
+    ERC20Mock paymentToken;
```
leeftk: I think we can just use the _token that is provided to us from ModuleTest, please refer to other tests for this flow but the idea is a token should be deployed and minted to the msg.sender once ModuleTest is initialized. Will ping @fabianschu here for a second opinion on this as I may incorrect.

test/modules/logicModule/PP_FundingPot_v1_Exposed.sol
```diff
@@ -0,0 +1,30 @@
...
+// Internal
+import {LM_PC_FundingPot_v1} from
+    "src/modules/logicModule/LM_PC_FundingPot_v1.sol";
+
+// Access Mock of the PP_FundingPot_v1 contract for Testing.
+contract LM_PC_FundingPot_v1_Exposed is LM_PC_FundingPot_v1 {
+// Use the `exposed_` prefix for functions to expose internal contract for
+// testing.
+
+// function exposed_setPayoutAmountMultiplier(uint newPayoutAmountMultiplier_)
```
leeftk: My guess here is that _ensureValidDepositAmount should be exposed and tested

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,133 @@
...
+    /// @return bool True if the address has the funding pot admin role, false otherwise.
+    function _checkForFundingPotAdminRole(address admin_)
+        internal
+        view
+        returns (bool)
+    {
+        return __Module_orchestrator.authorizer().checkForRole(
+            getFundingPotAdminRoleId(), admin_
+        );
+    }
```
fabianschu: All of this is handled by the authorizer, so you don't have to bother about it :)

To be more precise:
- the orchestrator admin can call `grantModuleRole` on the funding pot module because that one is inheriting from Module_v1 (like [here](https://github.com/InverterNetwork/contracts/blob/template/logic-module-payment-client/src/templates/tests/unit/LM_PC_Template_v1.t.sol#L155))
- `grantModuleRole` does the assigning of the role
- you only need to define the role as a constant (which you did)
- then, any function that needs to be gated to be accessible only by that role just needs the modifier `onlyModuleRole` (like [here](https://github.com/InverterNetwork/contracts/blob/template/logic-module-payment-client/src/templates/modules/LM_PC_Template_v1.sol#L185C9-L185C23))

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,38 @@
...
+    /// @param admin_ The address to grant the funding pot admin role to.
+    function grantFundingPotAdminRole(address admin_) external;
+
+    /// @notice Revokes the funding pot admin role from an address.
+    /// @param admin_ The address to revoke the funding pot admin role from.
+    function revokeFundingPotAdminRole(address admin_) external;
+
+    /// @notice Returns the funding pot admin role id.
+    /// @return roleId_ The funding pot admin role id.
+    function getFundingPotAdminRoleId() external view returns (bytes32);
```
fabianschu: Because this is being handled by the authorizer, these errors and functions are also redundant.

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
```diff
@@ -0,0 +1,221 @@
+// SPDX-License-Identifier: UNLICENSED
```
fabianschu, line 1: Tests also redundant for the same reason.

fabianschu (changes requested):

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
fabianschu: You're right, `ModuleTest` comes with a test token already that can be used 👍🏼

test/modules/logicModule/PP_FundingPot_v1_Exposed.sol
zzzuhaibmohd: @JeffreyJoel I have made the required changes and pushed the latest, Can you have a look at this and add the test?

JeffreyJoel: Done

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,126 @@
...
+    uint8 internal constant FLAG_CLIFF = 2;
+
+    /// @notice The payment processor flag for the end timestamp.
+    uint8 internal constant FLAG_END = 3;
+
+    // -------------------------------------------------------------------------
+    // State
+
+    /// @notice Payment token.
+    IERC20 internal _paymentToken;
```
fabianschu: Ultimately, the `_paymentToken` is determined by the token of the funding manager in the context of the workflow. Therefore this doesn't need to be stored here imo. This is the function that returns the token:

https://github.com/InverterNetwork/contracts/blob/main/src/modules/fundingManager/IFundingManager_v1.sol#L25

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,126 @@
...
+        __ERC20PaymentClientBase_v2_init(flags);
+    }
+
+    // -------------------------------------------------------------------------
+    // Public - Getters
+
+    /// @inheritdoc ILM_PC_FundingPot_v1
+    function getFundingPotAdminRole() external pure returns (bytes32 role_) {
+        return FUNDING_POT_ADMIN_ROLE;
+    }
```
fabianschu: There is an oversight in the template. You can remove the getter for constants and make the constant public as per our coding guidelines. Will adapt this in the template as well. Sorry for this 🙏🏼

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,28 @@
...
+    // -------------------------------------------------------------------------
+    // Public - Getters
+
+    /// @notice Returns the payment token.
+    /// @return token_ The address of the payment token.
+    function getPaymentToken() external view returns (address token_);
+
+    /// @notice Returns the funding pot admin role.
+    /// @return role_ The address of the funding pot admin role.
+    function getFundingPotAdminRole() external pure returns (bytes32 role_);
```
fabianschu: Can be removed as per my prev comments.

fabianschu (changes requested):

fabianschu (approved):
LGTM. 

There is still a console in the test, but that can be removed in a later PR.

Also right now the configData param is unused, but right now I don't have clarity myself if it will be needed or not, so let's leave it in for now.

PR #728, opened 2025-03-13
Fixes for [Audit] FM_PC_ExternalPrice_Redeeming, LM_ManualExternalPriceSetter_v1,

vendrell46:
- Removed redundant check on `_executePaymentQueue`
- Removed the redundant client validation in `executePaymentQueue`
- Moved `RedemptionOrderCreated` event emit before `_addPaymentOrder`
- Added the bytes data from the low level call in `executeRedemptionQueue` to the custom error


Zitzak (changes requested):
The comments have been implemented. However formatting and testing is failing, and also missing test adjustments as outlined in Slack @vendrell46

PR #729, opened 2025-03-20
Task: Refactor the test file structure

FHieser:
I refactored the folder structure of the test folder so that it is implementing the structure proposed in the inverter-standard document.
I renamed the AccessMock Contracts to contain the new standard of ending with _Exposed.
I updated the rest of the E2E tests that didnt end in .t.sol to do so now.

@marvinkruse Take a look if I missed anything. Im sure you will find something :stuck_out_tongue_winking_eye: 
@Zitzak I moved the Mocks of the BondingCurves to the new mocks folders. i hope the new structure is easy to understand for the external developers. Let me know if you think there is anything to improve the structure in that regard. Your review is optional here.


0xNuggan (approved):

remappings.txt
```diff
@@ -9,3 +9,6 @@
 @lm_pc/=src/modules/logicModule/paymentClient/
 @aut/=src/modules/authorizer/
 @modLib/=src/modules/lib/
+@mock/=test/mocks/
```
Zitzak: Would go with the whole name of the directory, i.e. `mocks` instead of `mock` to reduce confusion. Same for tool

Zitzak (changes requested):
General great push. Thank you for that. Few notes:
- The `rebasing` directory can be removed from `e2e/fundingManager/` as we don't have that contract anymore
- Maybe we can also remove the empty `bonding curve` folder, as those are only for the base contract, which we don't test in the e2e
- Do you know if the `ModuleTest_Template.txt` is up-to-date and/or used?
- Some mocks have not been remanded to `..._Exposed.sol`. Notably Bonding curve mocks, also Module base mock
- Curious why you named the one folder `tools`, as I would assume tools to be something else. These are more testing utilities

Lastly, in the code of conduct we decided that the exposed mocks are within the same folder as the tests themselves, so not in a separate mock folder (see image)
Screenshot 2025-04-10 at 10 49 40

FHieser:
- The rebasing directory can be removed from e2e/fundingManager/ as we don't have that contract anymore
  -  Will remove that
- Maybe we can also remove the empty bonding curve folder, as those are only for the base contract, which we don't test in the e2e
  - Will do 
- Do you know if the ModuleTest_Template.txt is up-to-date and/or used?
  - This is an old push on trying to make the testing structure more clear for outside sources and a more module development focus. @Zitzak you are currently the only one with the external development team that develops new modules, so if you dont use it, nobody does xD 
- Some mocks have not been remanded to ..._Exposed.sol. Notably Bonding curve mocks, also Module base mock
  - That was not in the scope of this PR, as some of the old mocks are not split properly in their "Access" and "Mock" functionality
- Curious why you named the one folder tools, as I would assume tools to be something else. These are more testing utilities
  - Yeah tools as a naming is the equivalent from "dont know how to name this random stuff properly" ^^ Will go with utilities instead
- Lastly, in the code of conduct we decided that the exposed mocks are within the same folder as the tests themselves, so not in a separate mock folder (see image)
  - Fuu -> Will think about that one, as I pointed out for the Mocks positioning -> Some of them are not properly structured so it isnt clear to me on first glance what the proper way of handling them here would be

remappings.txt
FHieser: Will do

FHieser:
@Zitzak and I agreed to push the merge until the PR #704 is done, as the changes here are not time critical and would have implications for the other PR

FHieser:
@Zitzak needs your approval

Zitzak (approved):
Thank you for the restructuring @FHieser.

During a sync with @FHieser around this PR we discussed the misalignment between the location of the Mocks in the PR in relation to what has been defined in the Code Of Conduct. 

Approving the PR in light of unblocking the Authorizer development. However I would like to request that updating the Mocks location will be done asap, as we're creating ambiguity which will effect the work we do with the external devs, as well as internal devs who haven't been part of this PR. This [Linear issue](https://linear.app/inverter/issue/SC-970/task-separation-of-accessmocks-and-plain-mocks) tracks the before mentioned change

Tagging the ppl who I think this message might be relevant for:
@0xNuggan @fabianschu @marvinkruse

PR #730, opened 2025-03-20
Task: Roleauthorizer_v1.1 upgrade

FHieser:
Based on #729


src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -4,10 +4,42 @@ pragma solidity ^0.8.0;
...
+    error Module__Authorizer__InvalidInitialAdmin();
+
+    /// @notice The provided role ID is the default admin role.
+    error Module__Authorizer__CannotAddDefaultAdminRole();
+
+    /// @notice The provided role ID is not existing.
+    error Module__Authorizer__RoleIdNotExisting();
+
+    /// @notice The admin of the provided role ID is already burned.
+    error Module__Authorizer__RoleAdminBurned();
```
0xNuggan: can be deleted

0xNuggan: (since it is not used)

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -21,99 +53,226 @@ interface IAuthorizer_v1 is IAccessControlEnumerable {
...
+        bytes32 respectiveAdminRole_,
+        address[] memory initialMembers_
+    ) external returns (bytes32 _newRoleId);
+
+    /// @notice Changes the name of a role.
+    /// @dev    Function access controlled by authorizer.
+    /// @dev    The role has to be created already.
+    /// @param  roleId_ The ID of the role to change the name of.
+    /// @param  newRoleName_ The new name of the role.
+    function labelRole(bytes32 roleId_, string memory newRoleName_) external;
```
0xNuggan, line 371: I'd add another tag either here or in the implementation stating that labels are not stored onchain, just emmitted

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -54,50 +58,52 @@ contract AUT_Roles_v1 is
...
+    // ========================================================================
+    // Storage
 
-    //--------------------------------------------------------------------------
+    /// @notice The role that is used as a placeholder for a burned admin role.
+    bytes32 public constant BURN_ADMIN_ROLE =
+        0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
+
+    /// @notice The role that is used as a placeholder for a public role.
+    bytes32 public constant PUBLIC_ROLE = bytes32(uint(1));
```
0xNuggan: Out of curiosity: Why a public role and not just a mapping of public functions that shortcuts the role check in a similar way to being admin?

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -135,164 +144,323 @@ contract AUT_Roles_v1 is
...
-        address[] calldata targets
-    ) external onlyModule(_msgSender()) {
-        bytes32 roleId = generateRoleId(_msgSender(), role);
-        for (uint i = 0; i < targets.length; i++) {
-            _revokeRole(roleId, targets[i]);
+    function addAccessPermission(
+        address target_,
+        bytes4 selector_,
+        bytes32 roleId_
+    ) public permissioned idNotDefaultAdmin(roleId_) idExisting(roleId_) {
```
0xNuggan: for completeness' sake, should we not also ensure that the role is not the Burn admin role? It shouldn't make a diffrence though

src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1.sol
```diff
@@ -431,70 +330,56 @@ contract FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1 is
...
-    {
+    ) public view override(BondingCurveBase_v1, IBondingCurveBase_v1) {
         revert
             FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1__InvalidFunctionality(
         );
     }
 
+    /// @inheritdoc IFM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1
+    function restrictBuyAndSell() external {
+        revert IModule_v1.Module__FunctionDeprecated();
```
0xNuggan: Since this seems like an important functionality, are we sure we just want to deprecate it like that? Maybe it would be better to change the implementation to calls that add/remove the public role from the selectors?

src/modules/fundingManager/bondingCurve/interfaces/IVirtualCollateralSupplyBase_v1.sol
```diff
@@ -31,10 +31,12 @@ interface IVirtualCollateralSupplyBase_v1 {
     //--------------------------------------------------------------------------
     // Functions
 
+    //@todo downstream contract access control? Where should I add this comment?
```
0xNuggan: As a general rule, I think we agreed to add them with the dev tag on the implemetation.

src/modules/fundingManager/bondingCurve/FM_BC_Bancor_Redeeming_VirtualSupply_v1.sol
```diff
@@ -239,9 +240,13 @@ contract FM_BC_Bancor_Redeeming_VirtualSupply_v1 is
         public
         virtual
         override(BondingCurveBase_v1, IBondingCurveBase_v1)
+        permissioned
         buyingIsEnabled
     {
-        buyFor(_msgSender(), _depositAmount, _minAmountOut);
+        (uint amountIssued, uint collateralFeeAmount) =
```
0xNuggan, line 246: I see why you did this. Part of the reason we were routing both sell orders (and buy too) through the other was to save on contract size. So if we run into issues on that front again, we may need to refactor the content into a common private function.

0xNuggan (commented):
This is only for the [FM updates](https://linear.app/inverter/issue/SC-998/task-review-authorizer-update-fundingmanager-modules). Looks good in general. Although since we will do a major version upgrade, we are missing the removal of all the deprecated functions/variables

0xNuggan (changes requested):
This is only for the [FM updates](https://linear.app/inverter/issue/SC-998/task-review-authorizer-update-fundingmanager-modules). Looks good in general. Although since we will do a major version upgrade, we are missing the removal of all the deprecated functions/variables

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: The function is restricted via the IdExisting Role which prevents the usage of ids that are higher than the id counter.
The Burned Admin id is never reached and can therefor not be added

FHieser: Looking for this I found a statement in the id existing that was redundant so thanks for making me look :+1:  xD

src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1.sol
FHieser: The wording on this functionality in the original BondingSurface implementation was not that good (my bad)
This implemented a role access restriction in the background. The turn off buy and sell functionality is still implemented in the base contract and not affected by this change

src/modules/fundingManager/bondingCurve/interfaces/IVirtualCollateralSupplyBase_v1.sol
FHieser: Yeah I wanted to wait with the changes until I finished the Queued PP and Oracle stuff to do that
Should be changed now

src/modules/authorizer/role/interfaces/IAUT_EXT_VotingRoles_v1.sol
```diff
@@ -44,6 +44,9 @@ interface IAUT_EXT_VotingRoles_v1 {
     //--------------------------------------------------------------------------
     // Errors
 
+    /// @notice Theis function is only callable by this contract.
```
0xNuggan: Typo

src/modules/authorizer/role/interfaces/IAUT_EXT_VotingRoles_v1.sol
```diff
@@ -44,6 +44,9 @@ interface IAUT_EXT_VotingRoles_v1 {
...
+    error Module__VotingRoleManager__NoSelfCall();
```
0xNuggan: I would rename this. As it stands it's not explicitly clear if it means "reverts because it is not a self-call" or "reverts because self calls are not allowed". Even saying "NotSelfCall" would be clearer

src/modules/base/IModule_v1.sol
```diff
@@ -23,34 +23,11 @@ interface IModule_v1 {
...
     // Errors
 
     /// @notice Function is only callable by authorized caller.
-    /// @param  role The role that is required.
-    /// @param  caller The address that is required to have the role.
-    error Module__CallerNotAuthorized(bytes32 role, address caller);
-
-    /// @notice Function is only callable by the {Orchestrator_v1}.
-    error Module__OnlyCallableByOrchestrator();
+    error Module__NotPermissioned();
```
0xNuggan: Same as in VotingRoles, I think the name isn't explicit enough. I would go for something like "UnauthorizedCall", "PermissionDenied" or "PermissionMissing"

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -21,99 +50,181 @@ interface IAuthorizer_v1 is IAccessControlEnumerable {
...
-    ///         In modules, this function should be used instead of `hasRole`, as
-    ///         there are Authorizer-specific checks that need to be performed.
-    /// @param  role The identifier of the role we want to check
-    /// @param  who  The address on which to perform the check.
-    /// @return bool Returns if the address holds the role
-    function checkForRole(bytes32 role, address who)
+    // ========================================================================
+    // Events
+
+    /// @notice Emits when a function permission is added to a role.
```
0xNuggan: Kind of nitpicky but I would invert the order of the explanation to avoid confusion. What we are doing in the code is adding a role to a function, not a function selector to a role. This way the order of parameters makes more sense also.

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -21,99 +50,181 @@ interface IAuthorizer_v1 is IAccessControlEnumerable {
...
+    /// @param  target The address of the target contract.
+    /// @param  functionSelector The selector of the function.
+    /// @param  roleId The ID of the role.
+    event AccessPermissionAdded(
+        address target, bytes4 functionSelector, bytes32 roleId
+    );
+
+    /// @notice Emits when a function permission is removed from a role.
```
0xNuggan: Same as above

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -21,99 +50,181 @@ interface IAuthorizer_v1 is IAccessControlEnumerable {
...
+    /// @notice Returns the number of created role IDs.
+    /// @return roleIdCounter_ The number of created role IDs.
+    function getRoleIdCounter() external view returns (uint roleIdCounter_);
+
+    /// @notice Returns wether the given roleId has the permission to callthe given function in the target contract.
+    /// @param  target_ The address of the target contract.
+    /// @param  selector_ The selector of the function.
+    /// @param  roleId_ The roleId that we want to check.
+    /// @return isPermissioned_ Returns if the roleId is permissioned to call the function.
+    function isPermissioned(address target_, bytes4 selector_, bytes32 roleId_)
```
0xNuggan, line 110: why not "hasPermission(role, target, selector)"? I'd asssume this function will mostly be called from the perspectiva of a role holder, right?

0xNuggan: "isPermissioned"to me suggests a split between public/permissioned. I think it would be clearer to override hasPermission below, and have (roleId, target, selector) as parameters. They both serve the same purpose ("check whether X can call this function"). Or at least rename this function to be more clear that it's checking for a specific role.

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -21,99 +50,181 @@ interface IAuthorizer_v1 is IAccessControlEnumerable {
...
+    /// @param  initialMembers_ The addresses of the initial members.
+    /// @return _newRoleId The ID of the newly created role.
+    function createRole(
+        string memory roleName_,
+        bytes32 respectiveAdminRole_,
+        address[] memory initialMembers_
+    ) external returns (bytes32 _newRoleId);
+
+    /// @notice Changes the name of a role.
+    /// @dev    Labels are emitted as events and therefor are not accessible
```
0xNuggan: typo

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -21,99 +50,181 @@ interface IAuthorizer_v1 is IAccessControlEnumerable {
...
-    /// @param  targets The addresses to grant the role to.
-    /// @dev	Only the addresses with the Admin role should be able to call this function.
-    function grantGlobalRoleBatched(bytes32 role, address[] calldata targets)
-        external;
+    // ------------------------------------------------------------------------
+    // Mutating - Role Management
+
+    /// @notice Creates a new role and adds initial members to it.
+    /// @dev    Function access controlled by authorizer.
+    /// @dev    The role if of the admin has to be created already.
```
0xNuggan: typo

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -21,99 +50,181 @@ interface IAuthorizer_v1 is IAccessControlEnumerable {
...
+    function burnAdminFromRole(bytes32 roleId_) external;
+
+    // ------------------------------------------------------------------------
+    // Mutating - Mixed Utility
+
+    /// @notice Creates a new role, adds initial members to it and adds permission to call to the respective functions.
+    /// @dev    Function access controlled by authorizer.
+    /// @dev    The role of the admin has to be created already.
+    /// @dev    The array of targets corresponds with the two dimensional array of selectors.
+    ///         The first position of targets therefor is assigned to the first position of the selector array.
```
0xNuggan: Typo

0xNuggan: https://www.grammarly.com/blog/commonly-confused-words/therefore-vs-therefor/ 🤓

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -35,6 +35,10 @@ import {AccessControlEnumerableUpgradeable} from
  *                          In case of any concerns or findings, please refer to our Security Policy
  *                          at security.inverter.network or email us directly!
  *
+ * @custom:version  v1.1.0
```
0xNuggan, line 233: I guess this will change now :grin:

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -54,50 +58,52 @@ contract AUT_Roles_v1 is
...
+    /// @notice Mapping that stores the role IDs that can be used to call functions on a target contract.
+    /// @dev    target The address of the target contract.
+    /// @dev    selector The function selector of the function to call.
+    /// @dev    roleIds The role IDs that can be used to call the function.
+    mapping(address target => mapping(bytes4 selector => bytes32[] roleIds))
+        public _permissions;
+
+    /// @notice The counter for role IDs.
+    /// @dev	This is used to generate unique role IDs for each role.
+    /// @dev    Starts at 1, which symbolizes two roles: PUBLIC_ROLE and DEFAULT_ADMIN_ROLE,
```
0xNuggan: I guess this is pretty much semantics, but wouldn't DEFAULT_ADMIN_ROLE and PUBLIC_ROLE be two roles at the end of the day? So the counter should start at two, and then have an assign->increment flow instead of increment->assign as is now?

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -135,164 +144,246 @@ contract AUT_Roles_v1 is
...
-        bytes32 roleId = generateRoleId(_msgSender(), role);
-        for (uint i = 0; i < targets.length; i++) {
-            _grantRole(roleId, targets[i]);
+    function hasPermission(address caller_, address target_, bytes4 selector_)
+        public
+        view
+        virtual
+        returns (bool hasPermission_)
+    {
+        // If caller is the admin, they can call any function.
```
0xNuggan, line 407: Just to double check: did we agree on this in the final spec?

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -54,50 +58,52 @@ contract AUT_Roles_v1 is
...
-    /// @param  role The id number of the role.
-    modifier notLastAdmin(bytes32 role) {
-        if (
-            role == DEFAULT_ADMIN_ROLE
-                && getRoleMemberCount(DEFAULT_ADMIN_ROLE) <= 1
-        ) {
-            revert Module__Authorizer__AdminRoleCannotBeEmpty();
+    /// @dev     Verifies that the roleId is already existing.
+    /// @param  roleId_ The id of the role.
+    modifier idExisting(bytes32 roleId_) {
```
0xNuggan: would rename to "idExists"

test/e2e/logicModule/BountyManagerE2E.t.sol
```diff
@@ -107,10 +111,65 @@ contract BountyManagerE2E is E2ETest {
             }
         }
 
-        // we authorize the deployer of the orchestrator as the bounty admin
-        bountyManager.grantModuleRole(
-            bountyManager.BOUNTY_ISSUER_ROLE(), address(this)
+        // =========
+        // Setting up Roles
```
0xNuggan, line 115: I think we discussed this, but just to bring it up in writing: I think something like this should be part of the base Module in the form of an abstract function. So on initialization or right afterwards it gets executed to at least configure the role setup (register+setAdmin), even if the roles have no users yet.

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: Maybe the naming of the counter itself is chosen a bit poorly.
It essentially represents the last created Id instead of the exact amount of ids.
Because the DefaultAdmin Role has to start at 0 and the next role would be the already created public role it is set to 1 as the public role is technically the last created role.
The additional purpose is to compare wether given ids are valid, because they have to be either equal or lower to the value to be valid.
Would it help if we named it something like lastCreatedId ?

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: Yes
you can look it up in the google doc under "Role Auth V2 Living Documentation"

test/e2e/logicModule/BountyManagerE2E.t.sol
FHieser: I had a talk with Marvin about this and he wanted to get back to you on this, as I raised some concerns.
He didnt I presume

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: Yes I didnt want to do this yet, as a change in the major version would mean, that I have to adapt the file name itself, which would prevent you from actually checking the differences between the files themselves.
This is the real final step of the PR and I will do that last

src/modules/authorizer/IAuthorizer_v1.sol
FHieser: Yeah I see the point
I renamed it to isRolePermissioned, which should make it a bit more clear

src/modules/base/IModule_v1.sol
FHieser: Renamed it to Module__CallerNotPermissioned

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -4,10 +4,39 @@ pragma solidity ^0.8.0;
...
 interface IAuthorizer_v1 is IAccessControlEnumerable {
-    //--------------------------------------------------------------------------
+    // ========================================================================
     // Errors
 
+    /// @notice The provided initial admin address is invalid.
+    error Module__Authorizer__InvalidInitialAdmin();
+
+    /// @notice The provided role ID is the default admin role.
+    error Module__Authorizer__CannotAddDefaultAdminRole();
```
marvinkruse: I find the error a bit misleading. This error is thrown when trying to add function restrictions to the default admin role. The current error name implies that the default admin role can not be added "to something". Doesn't match for me.

A better name would be something like `Module__Authorizer__CannotModifyAdminRoleAccess` or something like that maybe?

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -21,99 +50,182 @@ interface IAuthorizer_v1 is IAccessControlEnumerable {
...
+    /// @param  roleId The ID of the role for which the admin was burned.
+    event RoleAdminBurned(bytes32 roleId);
+
+    // ========================================================================
+    // Public Getter Functions
+
+    // ------------------------------------------------------------------------
+    // Getter -  Authorization
+
+    /// @notice Returns the permissions of the given function in the target contract.
```
marvinkruse: Too long

marvinkruse: May not just be this one, pelase check the contract in general, I can see a few more.

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -21,99 +50,182 @@ interface IAuthorizer_v1 is IAccessControlEnumerable {
...
-    /// @param  role  The ID number of the role to generate the hash for.
-    /// @return bytes32 Returns the generated role hash.
-    function generateRoleId(address module, bytes32 role)
+        returns (bytes32[] memory permissions_);
+
+    /// @notice Returns the number of created role IDs.
+    /// @return roleIdCounter_ The number of created role IDs.
+    function getRoleIdCounter() external view returns (uint roleIdCounter_);
+
+    /// @notice Returns wether the given roleId has the permission to callthe given function in the target contract.
```
marvinkruse:
```suggestion
    /// @notice Returns whether the given roleId has the permission to call the given function in the target contract.
```

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -21,99 +50,182 @@ interface IAuthorizer_v1 is IAccessControlEnumerable {
...
+    /// @param  target_ The address of the target contract.
+    /// @param  selector_ The selector of the function.
+    /// @param  roleId_ The roleId that we want to check.
+    /// @return isRolePermissioned_ Returns if the roleId is permissioned to call the function.
```
marvinkruse:
```suggestion
    /// @return isRolePermissioned_ Returns whether the roleId is permissioned to call the function.
```

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -21,99 +50,182 @@ interface IAuthorizer_v1 is IAccessControlEnumerable {
...
+    /// @dev    The role of the admin has to be created already.
+    /// @param  roleName_ The name of the role to create.
+    /// @param  respectiveAdminRole_ The role ID of the admin role.
+    /// @param  initialMembers_ The addresses of the initial members.
+    /// @return _newRoleId The ID of the newly created role.
+    function createRole(
+        string memory roleName_,
+        bytes32 respectiveAdminRole_,
+        address[] memory initialMembers_
+    ) external returns (bytes32 _newRoleId);
```
marvinkruse:
```suggestion
    ) external returns (bytes32 newRoleId_);
```

src/modules/authorizer/IAuthorizer_v1.sol
```diff
@@ -21,99 +50,182 @@ interface IAuthorizer_v1 is IAccessControlEnumerable {
...
+    // ------------------------------------------------------------------------
+    // Mutating - Role Management
+
+    /// @notice Creates a new role and adds initial members to it.
+    /// @dev    Function access controlled by authorizer.
+    /// @dev    The role of the admin has to be created already.
+    /// @param  roleName_ The name of the role to create.
+    /// @param  respectiveAdminRole_ The role ID of the admin role.
+    /// @param  initialMembers_ The addresses of the initial members.
+    /// @return _newRoleId The ID of the newly created role.
```
marvinkruse:
```suggestion
    /// @return newRoleId_ The ID of the newly created role.
```

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
```diff
@@ -45,11 +45,9 @@ contract AUT_EXT_VotingRoles_v1 is IAUT_EXT_VotingRoles_v1, Module_v1 {
     // Modifiers
 
     /// @dev	Reverts if caller is not the module itself.
-    modifier onlySelf() {
+    modifier onlyCallableByMotion() {
```
marvinkruse: This function doesn't check whether it was called by "the motion" but by itself. The current naming is not correct.

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -20,6 +20,7 @@ import {
 import {AccessControlEnumerableUpgradeable} from
     "@oz-up/access/extensions/AccessControlEnumerableUpgradeable.sol";
 
+//@todo adapt
```
marvinkruse: todo left

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -54,50 +59,52 @@ contract AUT_Roles_v1 is
...
-    //--------------------------------------------------------------------------
+    // ========================================================================
     // Modifiers
 
-    /// @dev	Verifies that the caller is an active module.
-    /// @param  module The address of the module.
-    modifier onlyModule(address module) {
-        if (!orchestrator().isModule(module)) {
-            revert Module__Authorizer__NotActiveModule(module);
+    modifier idNotDefaultAdmin(bytes32 roleId_) {
```
marvinkruse, line 261: Is just used once, so technically not necessary? But if you think there may be more instances in the future we can keep it.

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -54,50 +59,52 @@ contract AUT_Roles_v1 is
...
-        if (role == DEFAULT_ADMIN_ROLE && who == address(orchestrator())) {
-            revert Module__Authorizer__OrchestratorCannotHaveAdminRole();
-        }
-        _;
-    }
+    // ========================================================================
+    // Storage
 
-    //--------------------------------------------------------------------------
+    /// @notice The role that is used as a placeholder for a burned admin role.
```
marvinkruse: Why placeholder? What does it temporarily hold a place for? It is the burn role, or what am I missing?

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -54,50 +59,52 @@ contract AUT_Roles_v1 is
...
+    bytes32 public constant BURN_ADMIN_ROLE =
+        0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
+
+    /// @notice The role that is used as a placeholder for a public role.
```
marvinkruse: Why placeholder? What does it temporarily hold a place for? It is the public role, or what am I missing?

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -54,50 +59,52 @@ contract AUT_Roles_v1 is
...
+    /// @notice Mapping that stores the role IDs that can be used to call functions on a target contract.
+    /// @dev    target The address of the target contract.
+    /// @dev    selector The function selector of the function to call.
+    /// @dev    roleIds The role IDs that can be used to call the function.
+    mapping(address target => mapping(bytes4 selector => bytes32[] roleIds))
+        public _permissions;
+
+    /// @notice The counter for role IDs.
+    /// @dev	This is used to generate unique role IDs for each role.
+    /// @dev    Starts at 1, which symbolizes two roles: PUBLIC_ROLE and DEFAULT_ADMIN_ROLE,
```
marvinkruse: This description is very confusing, makes it seem like 1 is used for both the admin and public.

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -54,50 +59,52 @@ contract AUT_Roles_v1 is
...
+    ///         but is immediately incremented when a role is created.
+    uint internal _roleIdCounter;
+
+    /// @dev	Storage gap for future upgrades.
+    uint[47] private __gap;
```
marvinkruse: If this turns into v2, this can be reset to 50.

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -123,6 +130,9 @@ contract AUT_Roles_v1 is
             revert Module__Authorizer__InvalidInitialAdmin();
         }
 
+        // Start with 1 to symbolize two roles: DEFAULT_ADMIN_ROLE at 0 and PUBLIC_ROLE at 1.
```
marvinkruse: Description is a bit misleading. Maybe something like:
```suggestion
        // Starts at 1, as we already have two roles: DEFAULT_ADMIN_ROLE at 0 and PUBLIC_ROLE at 1.
```

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -123,6 +130,9 @@ contract AUT_Roles_v1 is
...
+        _roleIdCounter = 1;
+
         // Note about DEFAULT_ADMIN_ROLE: The Admin of the workflow holds the DEFAULT_ADMIN_ROLE, and has admin
```
marvinkruse: Probably line length.

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -135,164 +145,245 @@ contract AUT_Roles_v1 is
...
+        virtual
+        permissioned
+        idExists(respectiveAdminRole_)
+        returns (bytes32 newRoleId_)
     {
-        bytes32 roleId = generateRoleId(_msgSender(), role);
-        _setRoleAdmin(roleId, BURN_ADMIN_ROLE);
+        newRoleId_ = bytes32(++_roleIdCounter);
+
+        emit RoleCreated(newRoleId_, roleName_);
```
marvinkruse, line 455: We usually emit events at the end of the function, any reason why this is up here?

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -135,164 +145,245 @@ contract AUT_Roles_v1 is
...
-    function revokeGlobalRole(bytes32 role, address target)
+    function burnAdminFromRole(bytes32 roleId_)
         external
-        onlyRole(DEFAULT_ADMIN_ROLE)
+        onlyRole(getRoleAdmin(roleId_))
+        idExists(roleId_)
     {
-        bytes32 roleId = generateRoleId(address(orchestrator()), role);
-        _revokeRole(roleId, target);
+        // If Role Admin is Burned do nothing
```
marvinkruse:
```suggestion
        // If Role Admin is burned do nothing
```

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
```diff
@@ -25,6 +25,7 @@ interface TokenInterface {
     function balanceOf(address _owner) external view returns (uint balance);
 }
 
+//@todo update
```
marvinkruse: I think the interface of this contract wasn't adapted while the code was (according to the standard).

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
```diff
@@ -62,24 +63,33 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
...
         // Check that the role is empty
-        if (getRoleMemberCount(roleId) != 0) {
+        if (getRoleMemberCount(roleId_) != 0) {
             revert Module__AUT_TokenGated_Roles__RoleNotEmpty();
         }
 
         _;
     }
 
+    /// @dev     Modifier to guarantee that the role is not the public role.
```
marvinkruse: dev -> notice

marvinkruse: Unless you are just planning to do minor changes here and keep it intact otherwise, then ignore.

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
```diff
@@ -95,201 +105,206 @@ contract AUT_TokenGated_Roles_v1 is IAUT_TokenGated_Roles_v1, AUT_Roles_v1 {
...
-    function setTokenGated(bytes32 role, bool to)
+    function setTokenGated(bytes32 roleId_, bool to)
         public
-        onlyRole(getRoleAdmin(role))
-        onlyEmptyRole(role)
+        permissioned
+        idExists(roleId_)
+        onlyEmptyRole(roleId_)
+        notPublicRole(roleId_)
+    // @todo Do we prevent default admin here?
```
marvinkruse: I think it's fine to also allow the default admin to be token-gated. at least i can't think of any reason why not to?

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -135,164 +340,246 @@ contract AUT_Roles_v1 is
...
+        return DEFAULT_ADMIN_ROLE;
     }
 
+    // ------------------------------------------------------------------------
+    // Getter -  Authorization
+
     /// @inheritdoc IAuthorizer_v1
-    function generateRoleId(address module, bytes32 role)
+    function getPermissions(address target_, bytes4 selector_)
         public
```
marvinkruse: I can't find any place where this is used internally, so can be external? (May also be true for other functions, but maybe i also oversaw something where its used)

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -135,164 +340,246 @@ contract AUT_Roles_v1 is
...
+        for (uint i = 0; i < permissionLength; i++) {
+            if (
+                // if the role the public role
+                // or if the caller has the role
+                roleIds[i] == PUBLIC_ROLE || hasRole(roleIds[i], caller_)
+            ) {
+                return true;
+            }
+        }
+        // Caller does not have any of the roles, so they cannot call the function.
```
marvinkruse:
```suggestion
        // Caller does not have any of the roles, so they cannot call the
        // function.
```

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -135,164 +340,246 @@ contract AUT_Roles_v1 is
...
+        idExists(newAdminRoleId_)
     {
-        bytes32 roleId = generateRoleId(_msgSender(), role);
-        _setRoleAdmin(roleId, BURN_ADMIN_ROLE);
+        _setRoleAdmin(roleId_, newAdminRoleId_);
     }
 
     /// @inheritdoc IAuthorizer_v1
-    function grantGlobalRole(bytes32 role, address target)
+    function burnAdminFromRole(bytes32 roleId_)
```
marvinkruse: To me it feels this should be called `burnAdminOfRole` rather than `fromRole`. Maybe a bit too pedantic though, but you are the admin `of` something, not `from` something. Idk if that's just me though.

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -135,164 +340,246 @@ contract AUT_Roles_v1 is
...
-    /// @inheritdoc IAuthorizer_v1
-    function getAdminRole() public pure returns (bytes32) {
-        return DEFAULT_ADMIN_ROLE;
-    }
+        newRoleId_ =
+            createRole(roleName_, respectiveAdminRole_, initialMembers_);
 
-    //--------------------------------------------------------------------------
-    // Overloaded and overridden functions
+        // Run through all target and selector combinations and add permission to role id
```
marvinkruse:
```suggestion
        // Run through all target and selector combinations and add permission
        // to role id.
```

marvinkruse (changes requested):
Couldn't find any logical issues really, good work! Looks really good to me, just some minor things to be annoying.

src/modules/base/IModule_v1.sol
0xNuggan: *chef's kiss*

src/modules/authorizer/role/AUT_Roles_v1.sol
0xNuggan: yes, something like lastAssignedRoleId() would be more clear. Specially for the public getter function

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
0xNuggan: Yeah, I also think it can stay like that

src/modules/authorizer/IAuthorizer_v1.sol
FHieser, line 352:
```suggestion
    ) external returns (bytes32 newRoleId_);
```

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: Renamed it to lastAssignedRoleId

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
FHieser: I see your point here, that technically this is not restricted to motions, but to be only callable by the module itself, but the intention of the modifier is that it is actually only callable by the motions of the module.
Also Pablo already criticized the naming before ([here](https://github.com/InverterNetwork/contracts/pull/730#discussion_r2084104164) ), so im not quite sure how to name this properly.
If you have a better name please let me know :see_no_evil:

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: I personally like to put access or parameter restrictions into modifier, even if they are only used a single time.
Fits better in my personal workflow (how i work).
Is there a reason that I should not do it like that?

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: Yeah this was a leftover when I was not sure which number the Burnadmin role would get.
Just copy pasted the comment for the Public role

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: See comment above

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: For me the moment the counter is increased the role is technically created.
The setting of the admin and the granting of roles where different "functionality blocks".
It felt right to place the emit in the "Creation Block".
Do you want me to change it?

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: Hm
I would argue against that.
the Admin role is burned away **from** the role. It cant be set afterwards anymore.
"Admin Of" in this case would indicate that it is usable afterwards in my opinion

0xNuggan: maybe `burnRoleAdmin()`?

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
0xNuggan: Maybe my comment wasn't clear, apologies. I think that if the modifier controls for "self" to be calling, `onlySelf` is an appropriate name. My nitpick was more with the error name, which could be more explicit in saying "this is reverting because this function can only be called by the contract itself"

src/modules/authorizer/IAuthorizer_v1.sol
FHieser: Adapted Aut_Roles TokenGated and VotingRoles.

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: See comment above

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: burnRoleAdmin equals burnAdminOfRole for me,
So my reasoning still stands here

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
FHieser: okay then I will revert it to onlySelf and adapt the error to OnlySelfCallAllowed

src/modules/authorizer/role/AUT_Roles_v1.sol
marvinkruse: I still find the naming confusing, but I get where you are coming from. While I focus on the relationship being severed (remove admin OF role) you focus on the action of removal (remove admin FROM role). But I still find "burnAdminOfRole" too confusing, as it can also be understood as "burning the admin of a role" and not "burning the admin role of a role" which it technically is.
I'd also argue in favor of the idea from @0xNuggan to call it "burnRoleAdmin" (which is NOT the same as burnAdminOfRole, declaring them to be the same linguistically makes no sense to me, as admin and an admin role are not the same) - as that is 100% clear to me.

script/workflowDeploymentAndSetupScripts/DeployAndSetupNavBasedPimWorkflow.s.sol
```diff
@@ -1,4 +1,6 @@
-// SPDX-License-Identifier: UNLICENSED
+//@todo how to adapt this?
```
marvinkruse: Whats the issue here? Its not using the specific factory that was removed in the PR.

src/modules/authorizer/role/AUT_Roles_v1.sol
marvinkruse: No special reason, I just wouldn't have done it as it doesn't service any purpose in my brain, as we have loads of checks in functions that we also don't put into modifiers. But I'm not really annoyed by this one, so I'm perfectly fine to keep this if you want to :D

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -23,305 +23,612 @@ import {AccessControlEnumerableUpgradeable} from
...
-        if (
-            role == DEFAULT_ADMIN_ROLE
-                && getRoleMemberCount(DEFAULT_ADMIN_ROLE) <= 1
-        ) {
-            revert Module__Authorizer__AdminRoleCannotBeEmpty();
+    /// @notice Verifies that the roleId is already existing.
+    /// @param  roleId_ The id of the role.
+    modifier idExists(bytes32 roleId_) {
+        // If the given roleId is not equal or smaller than the last assigned
+        // roleId, then it is not existing.
```
marvinkruse: Its the other way around. As you outline the case where the error is thrown, this should be `If the given roleId is greater than the last assigned roleId, it does not exist.`.

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -23,305 +23,612 @@ import {AccessControlEnumerableUpgradeable} from
...
+        if (initialAdmin_ == address(0)) {
             revert Module__Authorizer__InvalidInitialAdmin();
         }
 
-        // Note about DEFAULT_ADMIN_ROLE: The Admin of the workflow holds the DEFAULT_ADMIN_ROLE, and has admin
-        // privileges on all Modules in the contract.
-        // It is defined in the AccessControl contract and identified with bytes32("0x00")
-        // Modules can opt out of this on a per-role basis by setting the admin role to "BURN_ADMIN_ROLE".
+        // Start with 1 to represent the two native roles:
+        // DEFAULT_ADMIN_ROLE at 0 and PUBLIC_ROLE at 1.
```
marvinkruse:
```suggestion
        // Start with 1 to account for the two native roles:
        // DEFAULT_ADMIN_ROLE at 0 and PUBLIC_ROLE at 1.
```

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -23,305 +23,612 @@ import {AccessControlEnumerableUpgradeable} from
...
+        _lastAssignedRoleId = 1;
+
+        // Note about DEFAULT_ADMIN_ROLE:
+        // The Admin of the workflow holds the DEFAULT_ADMIN_ROLE, and has
+        // admin privileges on all Modules in the contract.
```
marvinkruse:
```suggestion
        // The admin of the workflow holds the DEFAULT_ADMIN_ROLE, and has
        // admin privileges on all modules in the contract.
```

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -23,305 +23,612 @@ import {AccessControlEnumerableUpgradeable} from
...
-        returns (bool)
+        returns (bytes32[] memory permissions_)
     {
-        return hasRole(role, who);
+        permissions_ = _permissions[target_][selector_];
     }
 
     /// @inheritdoc IAuthorizer_v1
-    function generateRoleId(address module, bytes32 role)
+    function getLastAssignedRoleId()
```
marvinkruse, line 377: Is not used here, right? So can be external? There may be more functions that could be, I haven't checked all. This one just sticked out to me.

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -23,305 +23,612 @@ import {AccessControlEnumerableUpgradeable} from
...
+
+        // If there are no roles, the caller cannot call the function.
+        if (permissionLength == 0) {
+            return false;
+        }
+
+        // Go through each role and check if the caller has permission.
+        for (uint i = 0; i < permissionLength; i++) {
+            if (
+                // Return true if the role the public role
```
marvinkruse:
```suggestion
                // Return true if the role is the public role
```

src/modules/authorizer/role/AUT_Roles_v1.sol
marvinkruse: Yeah, alright, your argument makes sense indeed. I think it's fine then!

marvinkruse (changes requested):
Answered where I could find open questions and added a few tiny things more. 
Let me know if I missed anything you tagged me in or needed feedback on! @FHieser

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: Okay then I will adapt it :+1:

script/workflowDeploymentAndSetupScripts/DeployAndSetupNavBasedPimWorkflow.s.sol
FHieser: Yeah forgot to write you here.
The Deploymentscript itslef relies heavily on the old Authorizer for the predetermined Roles Setup.
I initially wasnt sure how to proceed on this, but I already talked to Marvin G about this.
As the script is specific to the Piku deployments and technically not needed anymore after the deployments are done,
I would scrap it for now, as we can reference it in the "Before Authorizer Update" Legacy Branch in the planned Repo

FHieser: Do you think that fits here?

marvinkruse: Resolved in huddle, we can remove this script, as it wont be needed when this is merged.

test/unit/modules/authorizer/role/AUT_Roles_v1.t.sol
```diff
@@ -1,1147 +1,1181 @@
...
+    ├── And: The given roleId is not existing
+    │   └── When: burnRoleAdmin is called
+    │       └── Then: The function should revert (modifier in position check)
+    ├── Given: Caller is the admin of the role for which the admin is being burned
+    └── And: The given roleId is existing
+        └── When: burnRoleAdmin is called
+            └── Then: The Admin should be burned
+    */
+
+    function testburnRoleAdmin_OnlyRoleAdmin(
```
marvinkruse:
```suggestion
    function testBurnRoleAdmin_OnlyRoleAdmin(
```

test/unit/modules/authorizer/role/AUT_Roles_v1.t.sol
```diff
@@ -1,1147 +1,1181 @@
...
-                _authorizer.generateRoleId(newModule, ROLE_0), ALBA
-            ),
-            false
-        );
-        assertEq(
-            _authorizer.hasRole(
-                _authorizer.generateRoleId(newModule, ROLE_0), BOB
-            ),
-            false
+    function testburnRoleAdmin_ModifierInPositionChecks() public {
```
marvinkruse:
```suggestion
    function testBurnRoleAdmin_ModifierInPositionChecks() public {
```

test/unit/modules/authorizer/role/AUT_Roles_v1.t.sol
```diff
@@ -1,1147 +1,1181 @@
...
 
-    // Test grant and revoke global roles
-
-    // Grant global roles
-    function testGrantGlobalRole() public {
-        bytes32 globalRole =
-            _authorizer.generateRoleId(address(_orchestrator), bytes32("0x03"));
-        vm.prank(ALBA);
-
+    function testburnRoleAdmin_idExists(bytes32 roleId_, bytes32 adminRoleId_)
```
marvinkruse:
```suggestion
    function testBurnRoleAdmin_idExists(bytes32 roleId_, bytes32 adminRoleId_)
```

marvinkruse (changes requested):
Just some capitalization issue after ctrl+f replacements

src/modules/authorizer/role/AUT_TokenGated_Roles_v1.sol
```diff
@@ -21,304 +21,444 @@ import {AccessControlUpgradeable} from
...
+ *                    role token gated. A token gated role behaves in all
+ *                    respects like a regular role, but handles the membership
+ *                    of that role differently. A member of a token gated role
+ *                    is only allowed to access the role functionalities if
+ *                    they hold a certain amount of a token.
+ *                    The implementation of this contract uses a few tricks to
+ *                    achieve this. Without going into too much detail, this
+ *                    is the main part that is needed to understand the basic
+ *                    mechanism:
+ *                    In the contract the token gating is done by instead of
```
0xNuggan: This explanation is a bit confusing to me. Suggestion: "In the contract, token gating is implemented by storing the token address in the role’s members property, instead of directly listing user addresses. This setup allows the contract to check the token balance of a user against a defined threshold when access is requested.
Example: We want to restrict a role to users who hold a certain amount of Token A. Therefore, we configure the role to be token-gated and set a required token amount the user needs to hold as threshold. When we then call 'grantRole' with the address of Token A, the role becomes accessible only to users whose wallet holds at least the specified amount of that token."

0xNuggan (approved):

marvinkruse (approved):

PR #731, opened 2025-03-24
Feat: Add fee collection to Queue based FM/PP

Zitzak:
https://linear.app/inverter/issue/SC-884/task-add-protocol-fee-to-pp-queue


src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
```diff
@@ -615,6 +620,55 @@ contract FM_PC_Oracle_Redeeming_v1 is
...
+            /* collateralTreasury */
+            ,
+            /* issuanceTreasury */
+            ,
+            /* collateralSellFeePercentage */
+            ,
+            uint issuanceSellFeePercentage
+        ) = _getFunctionFeesAndTreasuryAddresses(
+            bytes4(keccak256(bytes("_sellOrder(address,uint,uint)")))
+        );
```
fabianschu: The naming on these two functions makes it look like as if the collateralSellFeePercentage could have easier be gotten together with the issuanceSellFeePercentage in the `_getFunctionFeesAndTreasuryAddresses` call, which I assume is not the case. Is there a way to make clearer why two separate functions need to be called?

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
```diff
@@ -636,19 +690,26 @@ contract FM_PC_Oracle_Redeeming_v1 is
     ///         the proceeds.
     /// @param  receiver_ The address that will receive the redeemed tokens.
     /// @param  depositAmount_ The amount of tokens to be sold.
-    /// @param  collateralRedeemAmount_ The amount of collateral to redeem.
+    /// @param  netCollateralRedeemAmount_ The net amount of collateral to be
```
fabianschu, line 684: I cannot add a comment to the natspec in l. 688 cause it wasn't touched in this PR, but I find it confusing that it says "This function wraps the `_createAndEmitOrder` internal function" but seems to happen no function overloading here, so I don't understand how this function wraps `_createAndEmitOrder`.

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
```diff
@@ -717,11 +779,21 @@ contract FM_PC_Oracle_Redeeming_v1 is
...
+            /* collateralTreasury */
+        ) = _getCollateralSellFeePercentage();
+
+        // Get protocol issuance sell fee percentage
+        (
+            /* protocolCollateralTreasury */
+            ,
+            address protocolIssuanceTreasury,
+            /* protocolCollateralSellFeePercentage */
+            ,
```
fabianschu: Same comment as above: `protocolCollateralSellFeePercentage` appears in both functions which is confusing.

test/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.t.sol
```diff
@@ -241,6 +245,70 @@ contract FM_PC_ExternalPrice_Redeeming_v1_Test is ModuleTest {
...
+        );
+        feeManager.setIssuanceWorkflowFee(
+            address(_orchestrator),
+            address(fundingManager),
+            bytes4(keccak256(bytes("_sellOrder(address,uint,uint)"))),
+            true,
+            protocolIssuanceFee_
+        );
+        // Test
+
```
fabianschu:
```suggestion

        // Test
```

fabianschu (approved):
Some small comments, no deal breaker, do with the comments what you want

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
Zitzak: Good point, refactored for clarity

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
Zitzak: Agree

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
Zitzak: Addressed, good point ser

src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
Zitzak: Addressed

PR #732, opened 2025-03-26
Feat: Funding Pot V1 - Creating and Editing a Round

zzzuhaibmohd:
### Overview
This pull request is for the review of the `Funding_Pot_v1` contracts after adding implementation of `createRound()`, `editRound()` and `setAccessCriteriaForRound()`

### Key Changes
- **Funding Pot Admin:** Added logic to allow only the `FUNDING_POT_ADMIN_ROLE` to call `createRound()`, `editRound()` and `setAccessCriteriaForRound()`
- **Input Validation:** Added various custom errors to handle incorrect input validation and revert in case invalid input is passed
- **Unit Tests:**
  - Added unit and fuzz tests for `createRound()`, `editRound()` and `setAccessCriteriaForRound()` following the gherkin style method
- **Code Cleanup:** Removed unnecessary state variables and outdated templates to streamline the codebase.

### Notable Commits
- **Mar 17, 2025:** Create and Edit Round Implementation.
- **Mar 18, 2025:** Add tests related to create/edit round functionality.
- **Mar 24, 2025:** Implement the access control criteria 
- **Mar 25, 2025:** Add tests related to access control criteria.

### Focus Areas
- Provide feedback on implementation, unit tests and identifying any blind spots or areas where we can improve.
- If missing to handle any edge case or introduced a bug in our current implementation!

### DIFF
--- a/src/modules/logicModule/LM_PC_FundingPot_v1.sol
+++ b/src/modules/logicModule/LM_PC_FundingPot_v1.sol
@@ -152,10 +152,11 @@ contract LM_PC_FundingPot_v1 is
     }
 
     /// @inheritdoc ILM_PC_FundingPot_v1
-    function getRoundAccessCriteria(uint64 _roundId, uint64 _id)
+    function getRoundAccessCriteria(uint64 _roundId, uint8 _id)
         external
         view
         returns (
+            bool isOpen,
             address nftContract,
             bytes32 merkleRoot,
             address[] memory allowedAddresses
@@ -163,12 +164,22 @@ contract LM_PC_FundingPot_v1 is
     {
         Round storage round = rounds[_roundId];
         AccessCriteria storage accessCriteria = round.accessCriterias[_id];
-        //@todo : Waht if the round is open? add a bool? check what can be done !!!
-        return (
-            accessCriteria.nftContract,
-            accessCriteria.merkleRoot,
-            accessCriteria.allowedAddresses
-        );
+
+        if (accessCriteria.accessCriteriaId == AccessCriteriaId.OPEN) {
+            return (
+                true,
+                accessCriteria.nftContract,
+                accessCriteria.merkleRoot,
+                accessCriteria.allowedAddresses
+            );
+        } else {
+            return (
+                false,
+                accessCriteria.nftContract,
+                accessCriteria.merkleRoot,
+                accessCriteria.allowedAddresses
+            );
+        }
     }
 
     /// @inheritdoc ILM_PC_FundingPot_v1
--- a/src/modules/logicModule/LM_PC_FundingPot_v1.sol
+++ b/src/modules/logicModule/LM_PC_FundingPot_v1.sol
@@ -126,8 +126,6 @@ contract LM_PC_FundingPot_v1 is
     // Public - Getters
 
     /// @inheritdoc ILM_PC_FundingPot_v1
-    /// @dev    Returns the generic parameters of a round.
-
     function getRoundGenericParameters(uint64 _roundId)
         external
         view
@@ -153,6 +151,7 @@ contract LM_PC_FundingPot_v1 is
         );
     }
 
+    /// @inheritdoc ILM_PC_FundingPot_v1
     function getRoundAccessCriteria(uint64 _roundId, uint64 _id)
         external
         view
@@ -228,16 +227,10 @@ contract LM_PC_FundingPot_v1 is
         bytes memory _hookFunction,
         bool _closureMechanism,
         bool _globalAccumulativeCaps
-    ) external onlyModuleRole(FUNDING_POT_ADMIN_ROLE) returns (bool) {
+    ) external onlyModuleRole(FUNDING_POT_ADMIN_ROLE) {
         Round storage round = rounds[_roundId];
 
-        if (round.roundEnd == 0 && round.roundCap == 0) {
-            revert Module__LM_PC_FundingPot__RoundNotCreated();
-        }
-
-        if (block.timestamp > round.roundStart) {
-            revert Module__LM_PC_FundingPot__RoundAlreadyStarted();
-        }
+        _validateEditRoundParameters(round);
 
         round.roundStart = _roundStart;
         round.roundEnd = _roundEnd;
@@ -258,8 +251,6 @@ contract LM_PC_FundingPot_v1 is
             _closureMechanism,
             _globalAccumulativeCaps
         );
-
-        return true;
     }
 
     /// @inheritdoc ILM_PC_FundingPot_v1
@@ -270,13 +261,7 @@ contract LM_PC_FundingPot_v1 is
     ) external onlyModuleRole(FUNDING_POT_ADMIN_ROLE) {
         Round storage round = rounds[_roundId];
 
-        if (round.roundEnd == 0 && round.roundCap == 0) {
-            revert Module__LM_PC_FundingPot__RoundNotCreated();
-        }
-
-        if (block.timestamp > round.roundStart) {
-            revert Module__LM_PC_FundingPot__RoundAlreadyStarted();
-        }
+        _validateEditRoundParameters(round);
 
         if (
             (
@@ -331,6 +316,19 @@ contract LM_PC_FundingPot_v1 is
         if (round.hookContract == address(0) && round.hookFunction.length > 0) {
             revert
                 Module__LM_PC_FundingPot__HookContractRequiredWithHookFunction();
+        }
+    }
+
+    /// @notice Validates the round parameters.
+    /// @param  round The round to validate.
+    /// @dev    Reverts if the round parameters are invalid.
+    function _validateEditRoundParameters(Round storage round) internal view {
+        if (round.roundEnd == 0 && round.roundCap == 0) {
+            revert Module__LM_PC_FundingPot__RoundNotCreated();
+        }
+
+        if (block.timestamp > round.roundStart) {
+            revert Module__LM_PC_FundingPot__RoundAlreadyStarted();
         }
     }
 }
--- a/src/modules/logicModule/LM_PC_FundingPot_v1.sol
+++ b/src/modules/logicModule/LM_PC_FundingPot_v1.sol
@@ -168,8 +168,7 @@ contract LM_PC_FundingPot_v1 is
         AccessCriteria storage accessCriteria = round.accessCriterias[id_];
 
         isOpen = (
-            accessCriteria.accessCriteriaType == AccessCriteriaType.UNSET
-                || accessCriteria.accessCriteriaType == AccessCriteriaType.OPEN
+           accessCriteria.accessCriteriaType == AccessCriteriaType.OPEN
         );
         return (
             isOpen,
@@ -360,7 +359,7 @@ contract LM_PC_FundingPot_v1 is
         }
     }
 
-    /// @notice Validates the round parameters.
+    /// @notice Validates the round parameters before editing.
     /// @param  round_ The round to validate.
     /// @dev    Reverts if the round parameters are invalid.
     function _validateEditRoundParameters(Round storage round_) internal view {
--- a/src/modules/logicModule/LM_PC_FundingPot_v1.sol
+++ b/src/modules/logicModule/LM_PC_FundingPot_v1.sol
@@ -63,6 +63,16 @@ contract LM_PC_FundingPot_v1 is
     // -------------------------------------------------------------------------
 
     // State
+
+    /// @notice Payment token.
+    IERC20 internal _paymentToken;
+
+    /// @notice Stores all funding rounds by their unique ID.
+    mapping(uint64 => Round) public rounds;
+
+    /// @notice The next available round ID.
+    uint64 private nextRoundId;
+
     /// @notice Storage gap for future upgrades.
     uint[50] private __gap;
 
@@ -100,6 +110,108 @@ contract LM_PC_FundingPot_v1 is
     // -------------------------------------------------------------------------
     // Public - Mutating
 
+    /// @inheritdoc ILM_PC_FundingPot_v1
+    function createRound(
+        uint _roundStart,
+        uint _roundEnd,
+        uint _roundCap,
+        address _hookContract,
+        bytes memory _hookFunction,
+        bool _closureMechanism,
+        bool _globalAccumulativeCaps
+    ) external returns (uint64) {
+        if (_roundStart <= block.timestamp) {
+            revert Module__LM_PC_FundingPot__RoundStartMustBeInFuture();
+        }
+
+        if (_roundEnd <= _roundStart && _roundCap == 0) {
+            revert Module__LM_PC_FundingPot__RoundMustHaveEndTimeOrCap();
+        }
+
+        if (_roundEnd > 0 && _roundEnd <= _roundStart) {
+            revert Module__LM_PC_FundingPot__RoundEndMustBeAfterStart();
+        }
+
+        uint64 roundId = nextRoundId;
+        rounds[roundId] = Round({
+            roundStart: _roundStart,
+            roundEnd: _roundEnd,
+            roundCap: _roundCap,
+            hookContract: _hookContract,
+            hookFunction: _hookFunction,
+            closureMechanism: _closureMechanism,
+            globalAccumulativeCaps: _globalAccumulativeCaps,
+            isActive: true
+        });
+
+        nextRoundId++;
+
+        emit RoundCreated(
+            roundId,
+            _roundStart,
+            _roundEnd,
+            _roundCap,
+            _hookContract,
+            _closureMechanism,
+            _globalAccumulativeCaps
+        );
+
+        return roundId;
+    }
+
+    /// @inheritdoc ILM_PC_FundingPot_v1
+    function editRound(
+        uint64 _roundId,
+        uint _roundStart,
+        uint _roundEnd,
+        uint _roundCap,
+        address _hookContract,
+        bytes memory _hookFunction,
+        bool _closureMechanism,
+        bool _globalAccumulativeCaps
+    ) external returns (bool) {
+        Round storage round = rounds[_roundId];
+
+        if (round.roundStart == 0) {
+            revert Module__LM_PC_FundingPot__RoundDoesNotExist();
+        }
+
+        if (block.timestamp >= round.roundStart) {
+            revert Module__LM_PC_FundingPot__RoundAlreadyStarted();
+        }
+
+        if (_roundStart <= block.timestamp) {
+            revert Module__LM_PC_FundingPot__RoundStartMustBeInFuture();
+        }
+
+        if (_roundEnd <= _roundStart && _roundCap == 0) {
+            revert Module__LM_PC_FundingPot__RoundMustHaveEndTimeOrCap();
+        }
+
+        if (_roundEnd > 0 && _roundEnd <= _roundStart) {
+            revert Module__LM_PC_FundingPot__RoundEndMustBeAfterStart();
+        }
+
+        round.roundStart = _roundStart;
+        round.roundEnd = _roundEnd;
+        round.roundCap = _roundCap;
+        round.hookContract = _hookContract;
+        round.hookFunction = _hookFunction;
+        round.closureMechanism = _closureMechanism;
+        round.globalAccumulativeCaps = _globalAccumulativeCaps;
+
+        emit RoundEdited(
+            _roundId,
+            _roundStart,
+            _roundEnd,
+            _roundCap,
+            _hookContract,
+            _closureMechanism,
+            _globalAccumulativeCaps
+        );
+
+        return true;
+    }
     // -------------------------------------------------------------------------
     // Internal
 }
--- a/src/modules/logicModule/LM_PC_FundingPot_v1.sol
+++ b/src/modules/logicModule/LM_PC_FundingPot_v1.sol
@@ -126,12 +126,47 @@ contract LM_PC_FundingPot_v1 is
     // Public - Getters
 
     /// @inheritdoc ILM_PC_FundingPot_v1
-    function getRoundDetails(uint64 _roundId)
+    function getRoundGenericParameters(uint64 _roundId)
         external
         view
-        returns (Round memory)
+        returns (
+            uint roundStart,
+            uint roundEnd,
+            uint roundCap,
+            address hookContract,
+            bytes memory hookFunction,
+            bool closureMechanism,
+            bool globalAccumulativeCaps
+        )
     {
-        return rounds[_roundId];
+        Round storage round = rounds[_roundId];
+        return (
+            round.roundStart,
+            round.roundEnd,
+            round.roundCap,
+            round.hookContract,
+            round.hookFunction,
+            round.closureMechanism,
+            round.globalAccumulativeCaps
+        );
+    }
+
+    function getRoundAccessCriteria(uint64 _roundId, uint64 _id)
+        external
+        view
+        returns (
+            address nftContract,
+            bytes32 merkleRoot,
+            address[] memory allowedAddresses
+        )
+    {
+        Round storage round = rounds[_roundId];
+        AccessCriteria storage accessCriteria = round.accessCriterias[_id];
+        return (
+            accessCriteria.nftContract,
+            accessCriteria.merkleRoot,
+            accessCriteria.allowedAddresses
+        );
     }
 
     /// @inheritdoc ILM_PC_FundingPot_v1
@@ -155,17 +190,17 @@ contract LM_PC_FundingPot_v1 is
         nextRoundId++;
 
         uint64 roundId = nextRoundId;
-        rounds[roundId] = Round({
-            roundStart: _roundStart,
-            roundEnd: _roundEnd,
-            roundCap: _roundCap,
-            hookContract: _hookContract,
-            hookFunction: _hookFunction,
-            closureMechanism: _closureMechanism,
-            globalAccumulativeCaps: _globalAccumulativeCaps
-        });
 
-        _validateRoundParameters(rounds[roundId]);
+        Round storage round = rounds[roundId];
+        round.roundStart = _roundStart;
+        round.roundEnd = _roundEnd;
+        round.roundCap = _roundCap;
+        round.hookContract = _hookContract;
+        round.hookFunction = _hookFunction;
+        round.closureMechanism = _closureMechanism;
+        round.globalAccumulativeCaps = _globalAccumulativeCaps;
+
+        _validateRoundParameters(round);
 
         emit RoundCreated(
             roundId,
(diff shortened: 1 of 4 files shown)


src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -6,15 +6,235 @@ import {IERC20PaymentClientBase_v2} from
     "@lm/interfaces/IERC20PaymentClientBase_v2.sol";
 
 interface ILM_PC_FundingPot_v1 is IERC20PaymentClientBase_v2 {
-// -------------------------------------------------------------------------
-// Events
+    //--------------------------------------------------------------------------
```
fabianschu:
```suggestion
    // -------------------------------------------------------------------------
```

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -6,15 +6,235 @@ import {IERC20PaymentClientBase_v2} from
...
+        address[] allowedAddresses; // Explicit allowlist
+    }
 
-// -------------------------------------------------------------------------
-// Public - Mutating
+    // -------------------------------------------------------------------------
+    // Enums
+
+    /// @notice Enum used to identify the type of access criteria.
+    enum AccessCriteriaId {
```
fabianschu: Can we rename this to `AccessCriteriaType` cause the duplicate use of `id` in the ` struct Round` as well as in the `struct AccessCriteria` can be confusing?

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -6,15 +6,235 @@ import {IERC20PaymentClientBase_v2} from
...
+        OPEN, // 0
```
fabianschu: I feel a bit uncomfortable with the default value of the enum being `OPEN`. If the `AccessCriteria` is not set for an `id` in the `Round`, then I think a check for that `AccessCriteriaType` would make it look like as if the (nonexistent) AccessCriteria is set to OPEN. 

I'm sure, additional checks can/will prevent any undesired consequences, but putting OPEN in a different slot in the enum would play it safer, or putting `UNSET` explicitly in there at slot zero. 

Maybe I'm thinking too far ahead and this isn't really a problem, so please feel free to ignore, if you disagree :)

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -6,15 +6,235 @@ import {IERC20PaymentClientBase_v2} from
...
+    // Events
+
+    /// @notice Emitted when a new round is created.
+    /// @dev    This event signals the creation of a new round with specific parameters.
+    /// @param  roundId_ The unique identifier for the round.
+    /// @param  roundStart_ The timestamp when the round starts.
+    /// @param  roundEnd_ The timestamp when the round ends.
+    /// @param  roundCap_ The maximum allocation or cap for the round.
+    /// @param  hookContract_ The address of an optional hook contract for custom logic.
+    /// @param  closureMechanism_ A boolean indicating whether a specific closure mechanism is enabled.
```
fabianschu: I think the naming of this param could be more specific, like `autoClosure_` or similar.

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -6,15 +6,235 @@ import {IERC20PaymentClientBase_v2} from
...
+        bool globalAccumulativeCaps_
+    );
+
+    /// @notice Emitted when access criteria is set for a round.
+    /// @param  roundId_ The unique identifier of the round.
+    /// @param  accessId_ The identifier of the access criteria.
+    /// @param  accessCriteria_ The access criteria.
+    event AccessCriteriaSet(
+        uint64 indexed roundId_, uint8 accessId_, AccessCriteria accessCriteria_
+    );
```
fabianschu, line 108: For the Round, there are two separate events for creation and editing. Can we do the same for `AccessCriteria`?

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -6,15 +6,235 @@ import {IERC20PaymentClientBase_v2} from
...
+    error Module__LM_PC_FundingPot__HookFunctionRequiredWithContract();
+
+    /// @notice Thrown when a hook contract is specified without a hook function.
+    error Module__LM_PC_FundingPot__HookFunctionRequiredWithHookContract();
+
+    /// @notice Thrown when a hook function is specified without a hook contract.
+    error Module__LM_PC_FundingPot__HookContractRequiredWithHookFunction();
+
+    /// @notice Round does not exist
+    error Module__LM_PC_FundingPot__RoundNotCreated();
```
fabianschu: Is this a duplicate to Module__LM_PC_FundingPot__RoundDoesNotExist ?

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -6,15 +6,235 @@ import {IERC20PaymentClientBase_v2} from
...
+        bool globalAccumulativeCaps;
+        mapping(uint64 id => AccessCriteria) accessCriterias;
+    }
 
-// -------------------------------------------------------------------------
-// Public - Getters
+    /// @notice Struct used to store information about a funding round's access criteria.
+    /// @param  nftContract Address of the NFT contract.
+    /// @param  merkleRoot Merkle root for the access criteria.
+    /// @param  allowedAddresses Mapping of addresses to their access status.
```
fabianschu, line 36: Right now this is an array. I kind of preferred the struct because it shifts the gas costs to the admin and not to the user and globally requires less gas. Wdyt?

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -6,15 +6,235 @@ import {IERC20PaymentClientBase_v2} from
...
+    error Module__LM_PC_FundingPot__RoundEndMustBeAfterStart();
+
+    /// @notice Round has already started and cannot be modified
+    error Module__LM_PC_FundingPot__RoundAlreadyStarted();
+
+    /// @notice Hook function is required when a hook contract is provided
+    error Module__LM_PC_FundingPot__HookFunctionRequiredWithContract();
+
+    /// @notice Thrown when a hook contract is specified without a hook function.
+    error Module__LM_PC_FundingPot__HookFunctionRequiredWithHookContract();
```
fabianschu, line 140: This is a duplicate.

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -6,15 +6,235 @@ import {IERC20PaymentClientBase_v2} from
     "@lm/interfaces/IERC20PaymentClientBase_v2.sol";
 
 interface ILM_PC_FundingPot_v1 is IERC20PaymentClientBase_v2 {
```
fabianschu, line 8: Each NatSpec comment should end with a `.` :) 

Just adding this here once, but it is relevant in many places.

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -6,15 +6,235 @@ import {IERC20PaymentClientBase_v2} from
...
+            uint roundCap_,
+            address hookContract_,
+            bytes memory hookFunction_,
+            bool closureMechanism_,
+            bool globalAccumulativeCaps_
+        );
+
+    /// @notice Retrieves the access criteria for a specific funding round.
+    /// @param  roundId_ The unique identifier of the round to retrieve.
+    /// @param  id_ The identifier of the access criteria to retrieve.
```
fabianschu: For consistency, let's call this `accessCriteriaId_` and as commented elsewhere what is now `accessCriteriaId` to `accessCriteriaType`?

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -6,15 +6,235 @@ import {IERC20PaymentClientBase_v2} from
...
+    /// @param  globalAccumulativeCaps_ A boolean indicating whether global accumulative caps are enforced.
+    event RoundEdited(
+        uint indexed roundId_,
+        uint roundStart_,
+        uint roundEnd_,
+        uint roundCap_,
+        address hookContract_,
+        bool closureMechanism_,
+        bool globalAccumulativeCaps_
+    );
```
fabianschu: Should we add the hookFunction here?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -97,9 +125,223 @@ contract LM_PC_FundingPot_v1 is
...
+                accessCriteria.allowedAddresses
+            );
+        } else {
+            return (
+                false,
+                accessCriteria.nftContract,
+                accessCriteria.merkleRoot,
+                accessCriteria.allowedAddresses
+            );
+        }
```
fabianschu: Maybe simplify this?

```
isOpen = accessCriteria.accessCriteriaId == AccessCriteriaId.OPEN;
return (
    isOpen,
    accessCriteria.nftContract,
    accessCriteria.merkleRoot,
    accessCriteria.allowedAddresses
);
```

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -6,15 +6,235 @@ import {IERC20PaymentClientBase_v2} from
...
+        address hookContract_,
+        bytes memory hookFunction_,
+        bool closureMechanism_,
+        bool globalAccumulativeCaps_
+    ) external;
+
+    /// @notice Set Access Control Check
+    /// @dev    Only callable by funding pot admin and only before the round has started
+    /// @param  roundId_ ID of the round
+    /// @param  accessId_ ID of the access criteria
```
fabianschu: I think this should be accessCriteriaId_

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
fabianschu: Also, all return values should be named as per our coding standards (will help us to automate documentation generation). If we missed this in the template contract, please excuse that.

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -6,15 +6,235 @@ import {IERC20PaymentClientBase_v2} from
...
+    /// @param  accessCriteria_ Access criteria to set
+    function setAccessCriteriaForRound(
```
fabianschu, line 256: I think it would be more consistent to have a function for creation and one for editing, like we have for rounds. Wdyt?

fabianschu: related to this: https://github.com/InverterNetwork/contracts/pull/732#discussion_r2017269621

fabianschu: Should we then have an incrementing counter for accessCriteriaId as well?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -97,9 +125,223 @@ contract LM_PC_FundingPot_v1 is
...
+    /// @dev    Reverts if the round parameters are invalid.
+    function _validateRoundParameters(Round storage round_) internal view {
+        // Validate round start time is in the future
+        // @note: The below condition wont allow _roundStart == block.timestamp
+        if (round_.roundStart <= block.timestamp) {
+            revert Module__LM_PC_FundingPot__RoundStartMustBeInFuture();
+        }
+
+        // Validate that either end time or cap is set
+        if (round_.roundEnd == 0 || round_.roundCap == 0) {
```
fabianschu: Should this not be an `&&`?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -97,9 +125,223 @@ contract LM_PC_FundingPot_v1 is
...
+            revert Module__LM_PC_FundingPot__RoundMustHaveEndTimeOrCap();
+        }
+
+        // If end time is set, validate it's after start time
+        if (round_.roundEnd > 0 && round_.roundEnd <= round_.roundStart) {
```
fabianschu: Should we not have `<=` rather be `<`? I don't see the case where end == start such that the round is immediately closed. Maybe I'm missing something.

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
fabianschu, line 1: Could you remove the role constant in the beginning of the file?

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
```diff
@@ -95,16 +92,1004 @@ contract LM_PC_FundingPot_v1_Test is ModuleTest {
...
+    │   │       └── Then it should revert
+    ├── And hook contract is set but hook function is not set
+    │   └── When user attempts to create a round
+    │       └── Then it should revert
+    ├── And hook function is set but hook contract is not set
+    │   └── When user attempts to create a round
+    │       └── Then it should revert
+    └── Given all the valid parameters are provided
+        └── When user attempts to create a round
+            └── Then it should not be active and should return the round id
```
fabianschu:
```suggestion
├── Given user does not have FUNDING_POT_ADMIN_ROLE
│   └── When user attempts to create a round
│       └── Then it should revert
│
└── Given user has FUNDING_POT_ADMIN_ROLE
    ├── And round start < block.timestamp
    │   └── When user attempts to create a round
    │       └── Then it should revert
    │
    ├── And round end time == 0
    ├── And round cap == 0
    │   └── When user attempts to create a round
    │       └── Then it should revert
    │
    ├── And round end time is set
    ├── And round end != 0
    ├── And round end < round start
    │   └── When user attempts to create a round
    │       └── Then it should revert
    │
    ├── And hook contract is set but hook function is not set
    │   └── When user attempts to create a round
    │       └── Then it should revert
    │
    ├── And hook function is set but hook contract is not set
    │   └── When user attempts to create a round
    │       └── Then it should revert
    │
    └── And all the valid parameters are provided
        └── When user attempts to create a round
            └── Then it should not be active and should return the round id
```

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
```diff
@@ -95,16 +92,1004 @@ contract LM_PC_FundingPot_v1_Test is ModuleTest {
...
+        assertEq(nftContract, accessCriteria.nftContract);
+        assertEq(merkleRoot, accessCriteria.merkleRoot);
+        assertEq(allowedAddresses, accessCriteria.allowedAddresses);
+    }
 
     // -------------------------------------------------------------------------
     // Test: Internal Functions
+
+    // -------------------------------------------------------------------------
+    // Helper Functions
```
fabianschu: While our code of conduct isn't mentioning this specifically, I think the default values (`roundStart` ...) should be placed in the top of the file.

Also I think we shouldn't use wrappers just to call the SuT.

So I think we could remove:
- _helper_createDefaultFundingRound
- _helper_callCreateRound
- _helper_callEditRound
- _helper_createEditedRoundParams

_helper_createAccessCriteria could reuse the default values defined in the top then

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
fabianschu: There is a compilation error (stack too deep) for this file.

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
```diff
@@ -95,16 +92,1004 @@ contract LM_PC_FundingPot_v1_Test is ModuleTest {
...
+    │       └── Then it should revert
+    ├── Given hook contract is set
+    │   ├── And hook function is empty
+    │   └── When user attempts to edit the round
+    │       └── Then it should revert
+    └── Given hook function is set
+        ├── And hook contract is empty
+            └── When user attempts to edit the round
+                └── Then it should revert  
+    */
```
fabianschu:
```suggestion
    ├── Given user does not have FUNDING_POT_ADMIN_ROLE
    │   └── When user attempts to edit a round
    │       └── Then it should revert
    │
    ├── Given round does not exist
    │   └── When user attempts to edit the round
    │       └── Then it should revert
    │
    ├── Given round is active
    │   └── When user attempts to edit the round
    │       └── Then it should revert
    │
    ├── Given round start time is in the past
    │   └── When user attempts to edit a round with the above parameter
    │       └── Then it should revert
    │
    ├── Given round end time == 0
    ├── And round cap == 0
    │   └── When user attempts to edit a round with the above parameters
    │       └── Then it should revert
    │
    ├── Given round end time is set
    ├── And round end is before round start
    │   └── When user attempts to edit the round
    │       └── Then it should revert
    │
    ├── Given hook contract is set
    ├── And hook function is empty
    │   └── When user attempts to edit the round
    │       └── Then it should revert
    │
    ├── Given hook function is set
    └── And hook contract is empty
        └── When user attempts to edit the round
            └── Then it should revert
```

fabianschu (changes requested):
Great job :) I left a bunch of comments, nothing structural.

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
zzzuhaibmohd: @fabianschu we are planning to fuzz these values in the next PR review!

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
zzzuhaibmohd: @fabianschu can you explain the above in detail?

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
leeftk: We should def change this nice callout.

leeftk:
Realizing I never commented on this last night I just pinged Zuhaib internally. Anyway, fixes LGTM let us know about the extension y'all use for the gherkin so we can start formatting that properly as well! Thanks :)

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -97,9 +127,249 @@ contract LM_PC_FundingPot_v1 is
...
+            bytes32 merkleRoot,
+            address[] memory allowedAddresses
+        )
+    {
+        Round storage round = rounds[roundId_];
+        AccessCriteria storage accessCriteria = round.accessCriterias[id_];
+
+        isOpen = (
+            accessCriteria.accessCriteriaType == AccessCriteriaType.UNSET
+                || accessCriteria.accessCriteriaType == AccessCriteriaType.OPEN
```
fabianschu: This seems like unintended behaviour. Why is the access criteria considered "open" if no access criteria type is set?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -97,9 +127,249 @@ contract LM_PC_FundingPot_v1 is
...
+        }
+
+        if (round_.hookContract == address(0) && round_.hookFunction.length > 0)
+        {
+            revert
+                Module__LM_PC_FundingPot__HookContractRequiredWithHookFunction();
+        }
+    }
+
+    /// @notice Validates the round parameters.
```
fabianschu: The NatSpec is the same as in l. 328. I think it should tell the dev what the difference to the other validation function is.

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
fabianschu: Sorry, maybe I'm thinking to far ahead, but I would assume if the allowlist is stored as an array, then later on when someone that is on the allowlist wants to contribute, within the contribution function it will be necessary to iterate over the array. This means for each contributor this iteration will need to take place, and contributing is probably by far the most common operation that will take place on this contract.

If on the other hand the allowlist is stored as a mapping during configuration, then it is not necessary to iterate over an array for each contribution and since this is supposed to happen on ETH mainnet, it is maybe more efficient gas-wise. Wdyt?

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
fabianschu: I think this one was overseen :)

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
fabianschu: I think this might have been overlooked at this place :)

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
fabianschu: Noticing that in testSupportsInterface there is a duplicate assertion

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
```diff
@@ -95,16 +90,1197 @@ contract LM_PC_FundingPot_v1_Test is ModuleTest {
...
+        └── Given all the valid parameters are provided
+            └── When user attempts to set access criteria
+                └── Then it should not revert
+    */
+
+    function testFuzzSetAccessCriteria_revertsGivenUserDoesNotHaveFundingPotAdminRole(
+        uint8 accessCriteriaEnum_,
+        address user_
+    ) public {
+        vm.assume(accessCriteriaEnum_ >= 0 && accessCriteriaEnum_ <= 3);
```
fabianschu: I don't understand why this uses a different vm.assume than the other fuzz tests. Could you explain your intentions?

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
```diff
@@ -95,16 +90,1197 @@ contract LM_PC_FundingPot_v1_Test is ModuleTest {
...
+    }
 
     // -------------------------------------------------------------------------
     // Test: Internal Functions
+
+    // -------------------------------------------------------------------------
+    // Helper Functions
+
+    // @notice Creates a default funding round
+    function _helper_createDefaultFundingRound()
```
fabianschu, line 1106: Now that you have added fuzz tests, I still think this test would be easiert to read if the default values were stored on top of the file. Could we change this?

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
```diff
@@ -95,16 +90,1197 @@ contract LM_PC_FundingPot_v1_Test is ModuleTest {
...
+            roundCap,
+            hookContract,
+            hookFunction,
+            autoClosure,
+            globalAccumulativeCaps
+        );
+    }
+
+    // @notice calls the create round function
+    function _helper_callCreateRound(
```
fabianschu: I still don't understand the case for wrappers like this. If it is just calling the function on the funding pot, why not call it directly and remove that abstraction layer?

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
```diff
@@ -95,16 +90,1197 @@ contract LM_PC_FundingPot_v1_Test is ModuleTest {
...
+            roundCap,
+            hookContract,
+            hookFunction,
+            autoClosure,
+            globalAccumulativeCaps
+        );
+    }
+
+    // @notice Creates a predefined funding round with edited parameters for testing
+    function _helper_createEditedRoundParams()
```
fabianschu: I think this helper obscures what is happening in the tests. Instead of using it, could you store the default values on top of the file and when you actually want to edit round parameters in the test, make it clear in each test which parameter is being edited? 

Example:
In `testEditRound_revertsGivenUserIsNotFundingPotAdmin` you are calling this helper to test that only admin can call it. If you had the default values on top of the file you could call the editRound function directly with the default values and test the same thing. 

If you would like to test that editing reverts if you are trying to set the new start date after the end date, then you could just exchange the startDate with an invalid one etc.

Does that make sense?

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
```diff
@@ -95,16 +90,1197 @@ contract LM_PC_FundingPot_v1_Test is ModuleTest {
...
+            roundCap_,
+            hookContract_,
+            hookFunction_,
+            autoClosure_,
+            globalAccumulativeCaps_
+        );
+    }
+
+    // @notice calls the create round function
+    function _helper_callEditRound(
```
fabianschu: Let's remove this wrapper also?

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
```diff
@@ -95,16 +90,1197 @@ contract LM_PC_FundingPot_v1_Test is ModuleTest {
...
+            roundEnd,
+            roundCap,
+            hookContract,
+            hookFunction,
+            autoClosure,
+            globalAccumulativeCaps
+        );
+    }
+
+    function _helper_createAccessCriteria(uint8 accessCriteriaEnum)
```
fabianschu, line 1134: This helper makes sense to me :)

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
```diff
@@ -95,16 +90,1197 @@ contract LM_PC_FundingPot_v1_Test is ModuleTest {
...
+                    address(0x0),
+                    bytes32(uint(0x0)),
+                    allowedAddresses
+                );
+            }
+        }
+    }
+
+    // Helper function to set up a round with access criteria
+    function _helper_setupRoundWithAccessCriteria(uint8 accessCriteriaEnum)
```
fabianschu, line 1193: Also makes sense to me

fabianschu (changes requested):
Found some more things mostly related to the clarity of the tests.

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
fabianschu: I added more comments on this. I still think these helpers are redundant and removing them would improve the clarity of the tests.

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
leeftk: You're right these are unecessary. Removing these, nice catch.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
leeftk: yes this is right I removed this.

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
leeftk: Yup just added this to the top, I basically created a set of default values and then a set of edited values. Though I do think for some tests like creating a round these values should be fuzzed as we should be testing a bunch of different combinations of possible rounds to make sure there aren't an edge cases that break the creation of a round, however I agree that having these in the set up makes the most sense so Im changing that for now.

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
leeftk: Don't think this made sense, I changed all of them to be the same and the tests are still passing so I think this was just a mismatch and not intentional.

fabianschu (approved):
Looks good to me

PR #735, opened 2025-04-08
Funding Pot V1: Round Contribution

leeftk:
# Funding Pot v1: Round Contribution and Closure Implementation

## Overview
This pull request completes the Funding Pot v1 implementation by adding round contribution for users, auto round closure conditions, and capacity roll over. This builds upon the previous PR's foundation of round creation and access criteria.

### Round Contribution System
- Added contribution validation with time window checks
- Implemented personal and round-wide cap enforcement
- Added access criteria validation (NFT, Merkle, List)
- Implemented token transfer and accounting logic

### Round Closure Mechanism
- Implemented automatic closure conditions (cap/time-based)
- Added hook execution system for round closure
- Implemented unused capacity calculations

### Capacity Management
- Added cross-round capacity tracking
- Implemented global accumulative caps system
- Added personal cap accumulation logic
- Implemented round-to-round capacity transfer

## Notable Commits
- Apr 7, 2025: Fix variable shadowing warning
- Apr 7, 2025: Rename accessId to accessCriteriaId
- [Previous dates]: Implementation of contribution and closure mechanics

### DIFF
--- a/src/modules/logicModule/LM_PC_FundingPot_v1.sol
+++ b/src/modules/logicModule/LM_PC_FundingPot_v1.sol
@@ -186,7 +186,8 @@ contract LM_PC_FundingPot_v1 is
         Round storage round = rounds[roundId_];
         AccessCriteria storage accessCriteria = round.accessCriterias[id_];
 
-        isOpen = (accessCriteria.accessCriteriaType == AccessCriteriaType.OPEN);
+        bool isOpen =
+            (accessCriteria.accessCriteriaType == AccessCriteriaType.OPEN);
         return (
             isOpen,
             accessCriteria.nftContract,
@@ -211,7 +212,7 @@ contract LM_PC_FundingPot_v1 is
         Round storage round = rounds[roundId_];
         AccessCriteria storage accessCriteria = round.accessCriterias[accessId_];
 
-        if (accessCriteria.accessCriteriaId == AccessCriteriaId.OPEN) {
+        if (accessCriteria.accessCriteriaType == AccessCriteriaType.OPEN) {
             return (true, 0, false, 0, 0, 0);
         }
 
@@ -393,15 +394,15 @@ contract LM_PC_FundingPot_v1 is
         _validateEditRoundParameters(round);
 
         if (
-            round.accessCriterias[accessId_].accessCriteriaId
-                == AccessCriteriaId.OPEN
+            round.accessCriterias[accessId_].accessCriteriaType
+                == AccessCriteriaType.OPEN
         ) {
             highestCap = personalCap_;
         }
 
         if (
-            round.accessCriterias[accessId_].accessCriteriaId
-                == AccessCriteriaId.NFT && capByNFT_ > 0
+            round.accessCriterias[accessId_].accessCriteriaType
+                == AccessCriteriaType.NFT && capByNFT_ > 0
         ) {
             uint nftCap = personalCap_ + capByNFT_;
             if (nftCap > highestCap) {
@@ -410,8 +411,8 @@ contract LM_PC_FundingPot_v1 is
         }
 
         if (
-            round.accessCriterias[accessId_].accessCriteriaId
-                == AccessCriteriaId.MERKLE && capByMerkle_ > 0
+            round.accessCriterias[accessId_].accessCriteriaType
+                == AccessCriteriaType.MERKLE && capByMerkle_ > 0
         ) {
             uint merkleCap = personalCap_ + capByMerkle_;
             if (merkleCap > highestCap) {
@@ -420,8 +421,8 @@ contract LM_PC_FundingPot_v1 is
         }
 
         if (
-            round.accessCriterias[accessId_].accessCriteriaId
-                == AccessCriteriaId.LIST && capByList_ > 0
+            round.accessCriterias[accessId_].accessCriteriaType
+                == AccessCriteriaType.LIST && capByList_ > 0
         ) {
             uint listCap = personalCap_ + capByList_;
             if (listCap > highestCap) {
@@ -668,19 +669,22 @@ contract LM_PC_FundingPot_v1 is
         Round storage round = rounds[roundId_];
         AccessCriteria storage accessCriteria = round.accessCriterias[accessId_];
 
-        if (accessCriteria.accessCriteriaId == AccessCriteriaId.OPEN) {
+        if (accessCriteria.accessCriteriaType == AccessCriteriaType.OPEN) {
             return;
         }
 
         bool accessGranted = false;
-        if (accessCriteria.accessCriteriaId == AccessCriteriaId.NFT) {
+        if (accessCriteria.accessCriteriaType == AccessCriteriaType.NFT) {
             accessGranted =
                 _checkNftOwnership(accessCriteria.nftContract, msg.sender);
-        } else if (accessCriteria.accessCriteriaId == AccessCriteriaId.MERKLE) {
+        } else if (
+            accessCriteria.accessCriteriaType == AccessCriteriaType.MERKLE
+        ) {
             accessGranted = _validateMerkleProof(
                 accessCriteria.merkleRoot, merkleProof_, msg.sender, roundId_
             );
-        } else if (accessCriteria.accessCriteriaId == AccessCriteriaId.LIST) {
+        } else if (accessCriteria.accessCriteriaType == AccessCriteriaType.LIST)
+        {
             accessGranted = _checkAllowedAddressList(
                 accessCriteria.allowedAddresses, msg.sender
             );
--- a/src/modules/logicModule/LM_PC_FundingPot_v1.sol
+++ b/src/modules/logicModule/LM_PC_FundingPot_v1.sol
@@ -17,9 +17,12 @@ import {
 
 // External
 import {IERC20} from "@oz/token/ERC20/IERC20.sol";
+import {IERC721} from "@oz/token/ERC721/IERC721.sol";
 import {SafeERC20} from "@oz/token/ERC20/utils/SafeERC20.sol";
 import {ERC165Upgradeable} from
     "@oz-up/utils/introspection/ERC165Upgradeable.sol";
+
+import "@oz/utils/cryptography/MerkleProof.sol";
 
 /**
  * @title   Inverter Funding Pot Module
@@ -93,6 +96,12 @@ contract LM_PC_FundingPot_v1 is
     mapping(
         uint64 roundId => mapping(uint8 accessId => AccessCriteriaPrivilages)
     ) private accessCriteriaPrivilages;
+
+    /// @notice Maps round IDs to user addresses to contribution amounts
+    mapping(uint64 => mapping(address => uint)) private userContributions;
+
+    /// @notice Maps round IDs to total contributions
+    mapping(uint64 => uint) private roundTotalContributions;
 
     /// @notice The next available round ID.
     uint64 private nextRoundId;
@@ -398,6 +407,73 @@ contract LM_PC_FundingPot_v1 is
             _end
         );
     }
+
+    function contribute(
+        uint64 roundId_,
+        uint amount_,
+        uint8 accessId_,
+        address contributionToken_,
+        bytes32[] calldata merkleProof_
+    ) external {
+        // Validate input amount.
+        if (amount_ == 0) {
+            revert Module__LM_PC_FundingPot__InvalidDepositAmount();
+        }
+
+        Round storage round = rounds[roundId_];
+
+        // Validate round exists.
+        if (round.roundEnd == 0 && round.roundCap == 0) {
+            revert Module__LM_PC_FundingPot__RoundNotCreated();
+        }
+
+        // Validate round timing.
+        uint currentTime = block.timestamp;
+        if (currentTime < round.roundStart) {
+            revert Module__LM_PC_FundingPot__RoundHasNotStarted();
+        }
+        if (round.roundEnd > 0 && currentTime > round.roundEnd) {
+            revert Module__LM_PC_FundingPot__RoundHasEnded();
+        }
+
+        // Validate access criteria.
+        _validateAccessCriteria(roundId_, accessId_, merkleProof_);
+
+        // Retrieve user's previous contribution and calculate the personal cap.
+        address user = msg.sender;
+        uint userPreviousContribution = _getUserContribution(roundId_, user);
+        uint userPersonalCap = _getUserPersonalCap(roundId_, user);
+
+        if (userPreviousContribution >= userPersonalCap) {
+            revert Module__LM_PC_FundingPot__PersonalCapReached();
+        }
+
+        // Adjust contribution if it would exceed the personal cap.
+        uint userRemainingCap = userPersonalCap - userPreviousContribution;
+        uint actualContributionAmount = amount_;
+        if (amount_ > userRemainingCap) {
+            actualContributionAmount = userRemainingCap;
+        }
+
+        uint totalRoundContribution = _getTotalRoundContribution(roundId_);
+        if (round.roundCap > 0 && totalRoundContribution >= round.roundCap) {
+            revert Module__LM_PC_FundingPot__RoundCapReached();
+        }
+        uint roundRemainingCap = round.roundCap - totalRoundContribution;
+        if (actualContributionAmount > roundRemainingCap) {
+            actualContributionAmount = roundRemainingCap;
+        }
+
+        // Transfer funds.
+        IERC20(contributionToken_).safeTransferFrom(
+            user, address(this), actualContributionAmount
+        );
+
+        // Record the contribution.
+        _recordContribution(roundId_, user, actualContributionAmount);
+        emit ContributionMade(roundId_, user, actualContributionAmount);
+    }
+
     // -------------------------------------------------------------------------
     // Internal
 
@@ -462,5 +538,124 @@ contract LM_PC_FundingPot_v1 is
         // _start + _cliff should be less or equal to _end
         // this already implies that _start is not greater than _end
         return _start + _cliff <= _end;
+    }
+
+    function _validateAccessCriteria(
+        uint64 roundId_,
+        uint8 accessId_,
+        bytes32[] calldata merkleProof_
+    ) internal view {
+        Round storage round = rounds[roundId_];
+        AccessCriteria storage accessCriteria = round.accessCriterias[accessId_];
+
+        if (accessCriteria.accessCriteriaId == AccessCriteriaId.OPEN) {
+            return;
+        }
+
+        bool accessGranted = false;
+        if (accessCriteria.accessCriteriaId == AccessCriteriaId.NFT) {
+            accessGranted =
+                _checkNftOwnership(accessCriteria.nftContract, msg.sender);
+        } else if (accessCriteria.accessCriteriaId == AccessCriteriaId.MERKLE) {
+            //TODO: Should I move this into a helper function
+
+            bytes32 leaf = keccak256(abi.encodePacked(msg.sender, roundId_));
+            accessGranted = MerkleProof.verify(
+                merkleProof_, accessCriteria.merkleRoot, leaf
+            );
+        } else if (accessCriteria.accessCriteriaId == AccessCriteriaId.LIST) {
+            accessGranted = _checkAllowedAddressList(
+                accessCriteria.allowedAddresses, msg.sender
+            );
+        }
+
+        if (!accessGranted) {
+            revert Module__LM_PC_FundingPot__AccessNotPermitted();
+        }
+    }
+
+    function _getTotalRoundContribution(uint64 roundId_)
+        internal
+        view
+        returns (uint)
+    {
+        return roundTotalContributions[roundId_];
+    }
+
+    function _getUserContribution(uint64 roundId_, address user_)
+        internal
+        view
+        returns (uint)
+    {
+        return userContributions[roundId_][user_];
+    }
+
+    function _getUserPersonalCap(uint64 roundId_, address user_)
+        internal
+        view
+        returns (uint)
+    {
+        uint basePersonalCap = 1000 ether;
+        Round storage round = rounds[roundId_];
+
+        if (round.globalAccumulativeCaps) {
+            uint unusedCapacity =
+                _getUnusedCapacityFromPreviousRounds(user_, roundId_);
+            return basePersonalCap + unusedCapacity;
+        }
+        return basePersonalCap;
+    }
+
+    function _getUnusedCapacityFromPreviousRounds(
+        address user_,
+        uint64 currentRoundId_
+    ) internal view returns (uint) {
+        uint totalUnusedCapacity = 0;
+        for (uint64 i = 1; i < currentRoundId_; i++) {
+            Round storage prevRound = rounds[i];
+            if (!prevRound.globalAccumulativeCaps) {
+                continue;
+            }
+            uint personalCap = 1000 ether;
+            uint userContribution = _getUserContribution(i, user_);
+            if (userContribution < personalCap) {
+                totalUnusedCapacity += (personalCap - userContribution);
+            }
+        }
+        return totalUnusedCapacity;
+    }
+
+    function _recordContribution(uint64 roundId_, address user_, uint amount_)
+        internal
+    {
+        userContributions[roundId_][user_] += amount_;
+        roundTotalContributions[roundId_] += amount_;
+    }
+
+    function _checkAllowedAddressList(
+        address[] memory allowedAddresses,
+        address sender
+    ) internal pure returns (bool) {
+        for (uint i = 0; i < allowedAddresses.length; i++) {
+            if (allowedAddresses[i] == sender) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    function _checkNftOwnership(address nftContract_, address user_)
+        internal
+        view
+        returns (bool)
+    {
+        if (nftContract_ == address(0) || user_ == address(0)) {
+            return false;
+        }
+        try IERC721(nftContract_).balanceOf(user_) returns (uint balance) {
+            return balance > 0;
+        } catch {
+            return false;
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -101,18 +109,54 @@ interface ILM_PC_FundingPot_v1 is IERC20PaymentClientBase_v2 {
 
     /// @notice Emitted when access criteria is set for a round.
     /// @param  roundId_ The unique identifier of the round.
-    /// @param  accessId_ The identifier of the access criteria.
+    /// @param  AccessCriteriaId The identifier of the access criteria.
```
fabianschu:
```suggestion
    /// @param  accessCriteriaId The identifier of the access criteria.
```

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -101,18 +109,54 @@ interface ILM_PC_FundingPot_v1 is IERC20PaymentClientBase_v2 {
...
     /// @param  accessCriteria_ The access criteria.
     event AccessCriteriaEdited(
-        uint64 indexed roundId_, uint8 accessId_, AccessCriteria accessCriteria_
+        uint64 indexed roundId_,
+        uint8 AccessCriteriaId,
+        AccessCriteria accessCriteria_
+    );
+
+    /// @notice Emitted when access criteria Privileges are set for a round.
```
fabianschu:
```suggestion
    /// @notice Emitted when access criteria privileges are set for a round.
```

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -101,18 +109,54 @@ interface ILM_PC_FundingPot_v1 is IERC20PaymentClientBase_v2 {
...
+    /// @param  roundId_ The unique identifier of the round.
+    /// @param  AccessCriteriaId The identifier of the access criteria.
+    /// @param  personalCap_ The personal cap for the access criteria.
+    /// @param  overrideCap_ Whether to override the global cap.
+    /// @param  start_ The start timestamp for the access criteria.
```
fabianschu:
```suggestion
    /// @param  start_ The start timestamp for for when the linear vesting starts.
```

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -101,18 +109,54 @@ interface ILM_PC_FundingPot_v1 is IERC20PaymentClientBase_v2 {
...
+    /// @param  cliff_ The cliff timestamp for the access criteria.
+    /// @param  end_ The end timestamp for the access criteria.
```
fabianschu:
```suggestion
    /// @param  end_ The end timestamp for when the linear vesting ends.
```

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -101,18 +109,54 @@ interface ILM_PC_FundingPot_v1 is IERC20PaymentClientBase_v2 {
...
+        AccessCriteria accessCriteria_
+    );
+
+    /// @notice Emitted when access criteria Privileges are set for a round.
+    /// @param  roundId_ The unique identifier of the round.
+    /// @param  AccessCriteriaId The identifier of the access criteria.
+    /// @param  personalCap_ The personal cap for the access criteria.
+    /// @param  overrideCap_ Whether to override the global cap.
+    /// @param  start_ The start timestamp for the access criteria.
+    /// @param  cliff_ The cliff timestamp for the access criteria.
```
fabianschu:
```suggestion
    /// @param  cliff_ The time in seconds from start time at which the unlock starts.
```

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -179,31 +261,52 @@ interface ILM_PC_FundingPot_v1 is IERC20PaymentClientBase_v2 {
     /// @notice Retrieves the access criteria for a specific funding round.
     /// @param  roundId_ The unique identifier of the round to retrieve.
     /// @param  accessCriteriaId_ The identifier of the access criteria to retrieve.
-    /// @return isOpen_ Whether the access criteria is open.
+    /// @return isRoundOpen_ Whether the access criteria is open.
```
fabianschu:
```suggestion
    /// @return isRoundOpen_ Whether anyone can contribute as part of the access criteria.
```

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -179,31 +261,52 @@ interface ILM_PC_FundingPot_v1 is IERC20PaymentClientBase_v2 {
...
         view
         returns (
-            bool isOpen_,
+            bool isRoundOpen_,
             address nftContract_,
             bytes32 merkleRoot_,
             address[] memory allowedAddresses_
         );
 
+    /// @notice Retrieves the access criteria Privileges for a specific funding round.
```
fabianschu:
```suggestion
    /// @notice Retrieves the access criteria privileges for a specific funding round.
```

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
fabianschu: applies to many places in the code

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
fabianschu: applies to many places

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -101,18 +109,54 @@ interface ILM_PC_FundingPot_v1 is IERC20PaymentClientBase_v2 {
...
+        uint64 indexed roundId_,
+        uint8 AccessCriteriaId,
+        AccessCriteria accessCriteria_
+    );
+
+    /// @notice Emitted when access criteria Privileges are set for a round.
+    /// @param  roundId_ The unique identifier of the round.
+    /// @param  AccessCriteriaId The identifier of the access criteria.
+    /// @param  personalCap_ The personal cap for the access criteria.
+    /// @param  overrideCap_ Whether to override the global cap.
```
fabianschu: Could we have consistent naming for the params? Here it is overrideCap_, elsewhere it is overrideContributionSpan_?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -114,14 +159,17 @@ contract LM_PC_FundingPot_v1 is
         bytes memory configData_
     ) external override(Module_v1) initializer {
         __Module_init(orchestrator_, metadata_);
-
+        address fundingPotToken;
+        (fundingPotToken) = abi.decode(configData_, (address));
```
fabianschu: The "fundingPotToken" always needs to be the `token()` of the funding manager. It doesn't need to be passed in the init.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -272,53 +362,214 @@ contract LM_PC_FundingPot_v1 is
...
+            cliff_,
+            end_
+        );
+    }
+
+    /// @inheritdoc ILM_PC_FundingPot_v1
+    function contributeToRound(
+        uint64 roundId_,
+        uint amount_,
+        uint8 accessCriteriaId_,
```
fabianschu, line 558: Doesn't this revert already for the very first contribution that comes in if autoClosure is activated?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
zzzuhaibmohd: @fabianschu Can you help with the code snippet to use to fetch the token address from the funding manager during the init call?

fabianschu: It's actually just `token()` and the function can be found on IFundingManager.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -272,53 +362,214 @@ contract LM_PC_FundingPot_v1 is
...
+        bytes32[] calldata merkleProof_
+    ) external {
+        uint adjustedAmount = _validateRoundContribution(
+            roundId_, accessCriteriaId_, merkleProof_, amount_, msg.sender
```
fabianschu: Throughout the whole codebase, any module should be using _msgSender() inherited from ERC2771ContextUpgradeable via Module_v1.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -369,4 +620,366 @@ contract LM_PC_FundingPot_v1 is
...
+            accessCriteria.accessCriteriaType == AccessCriteriaType.MERKLE
+        ) {
+            return _validateMerkleProof(
+                accessCriteria.merkleRoot, merkleProof_, user_, roundId_
+            );
+        } else if (accessCriteria.accessCriteriaType == AccessCriteriaType.LIST)
+        {
+            if (!accessCriteria.allowedAddresses[user_]) {
+                revert Module__LM_PC_FundingPot__AccessCriteriaListFailed();
+                return false;
```
fabianschu:
```suggestion
```

fabianschu: return can be removed

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -369,4 +620,366 @@ contract LM_PC_FundingPot_v1 is
...
+            }
+
+            return true;
```
fabianschu: Will this return true if the Access Criteria is unset?

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -38,7 +38,15 @@ interface ILM_PC_FundingPot_v1 is IERC20PaymentClientBase_v2 {
         AccessCriteriaType accessCriteriaType;
         address nftContract; // NFT contract address (0x0 if unused)
         bytes32 merkleRoot; // Merkle root (0x0 if unused)
-        address[] allowedAddresses; // Explicit allowlist
+        mapping(address user => bool isAllowed) allowedAddresses; // Mapping of allowed addresses
+    }
+
+    struct AccessCriteriaPrivileges {
```
fabianschu, line 50: lacks natspec

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -369,4 +620,366 @@ contract LM_PC_FundingPot_v1 is
...
+        }
+        return personalCap;
+    }
+
+    /// @notice Calculates unused contribution capacity from previous rounds
+    /// @dev    Aggregates unused contribution caps from previous rounds with global accumulative caps
+    /// @param  user_ The address of the user
+    /// @param  currentRoundId_ The ID of the current round
+    /// @return Total unused contribution capacity from previous rounds
+    function _getUserUnusedCapacityFromPreviousRounds(
```
fabianschu: I think this function includes unintended behavior: If I understand correctly, if global accumulative caps (GAC) apply for the current round, we iterate through all previous rounds and check per round if it also has GAC enabled. In this case we iterate over the access criteria and search for the one with the highest personal cap configured. Then we check how much the current user contributed in that round and subtract it from the highes personal cap. This difference we add to the amount that the user can contribute in the current round.

This means 
- given we are in round 2 now 
- and GAC is enabled 
- and GAC had been in round 1
- and in round 1, NFT holders could contribute most with 1k each
- and I do not hold an NFT
- and I am eligible to contribute amount X in round 2
- then I will be able to contribute X + 1k in round 2

But this is not what we need. In the above example if I did not hold the NFT (and weren't qualified for any other access criteria), then I should just be able to contribute as much as defined by the round 2 access criteria that I qualify for.

In other words personal contribution capacity of a round access criteria should only carry over if I actually were eligible to contribute via said access criteria in that round. Or how my Ai friend would formulate it:

> Problem: Why should the user's unused capacity be based on the maximum possible personal cap in a previous round, rather than the cap associated with the specific criteria they actually used? This logic seems complex and potentially doesn't reflect the intended carry-over mechanism accurately. It could grant users unintended high capacities in later rounds based on criteria they never qualified for in earlier rounds.

fabianschu: The problem is a bit tricky, because iterating over previous rounds and access criteria to see if the user was eligible would introduce quite some complexity, if it is even possible (NFT, merkle proof..). 

I would probably approach it as follows, but feel free to suggest other approaches:

Use function overloading to have **two** external contributeToRound functions and **one** internal _contributeToRounds:

A) The **internal** one _contributeToRounds does what your current contributeToRound is doing, except it doesn't check for individual caps from previous rounds that carry over (basically skips `_getUserUnusedCapacityFromPreviousRounds`) and instead takes another param `unspentPersonalCap` which conceptually corresponds to your `totalUnusedCapacity` in here which it uses to adjust the personal capacity for the current round.

B) Then there is one **external** function contributeToRound that is identical to the one that you have:

```
contributeToRound(uint64 roundId_, uint amount_, uint8 accessCriteriaId_, bytes32[] calldata merkleProof) external
```

It *is intended to be used if the user has no previous capacities to carry over* and just calls the internal one and passes `unspentPersonalCap = 0`.

C) Lastly there is a version of contributeToRound to be called *if there are previous personal caps to be considered*
```
struct UnspentPersonalRoundCap {
  uint8 roundId_,
  uint8 accessCriteriaId_,
  bytes32[] calldata merkleProof_
}

contributeToRound(uint64 roundId_, uint amount_, uint8 accessCriteriaId_, bytes32[] calldata merkleProof, UnspentPersonalRoundCap[] unspentPersonalRoundCaps) external
```

Before this function calls the internal one, it would iterate over the `unspentPersonalRoundCaps`, check per array element if the user was eligible for the specified access criteria in the specified round, how much the cap was, how much they actually contributed and add the difference to `unspentPersonalCap`, then pass the `unspentPersonalCap` to the internal function.

fabianschu: This way we would have a slim contributeToRound function for cases where GAC don't apply, so in this case nobody has to bother about assembilng the additional params.

fabianschu (changes requested):
Found unintended behavior in for global accumulative personal caps

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
leeftk: Yes sorry this was an issue when rebasing as we had them named separate things on divergent branches, we implemented a process internally around rebasing that should help avoid this issue in the future.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
JeffreyJoel: @fabianschu can you please elaborate on this? I don't understand what you mean

src/modules/logicModule/LM_PC_FundingPot_v1.sol
JeffreyJoel: Yes, but it shouldn't. I have corrected it.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
fabianschu: has been clarified via slack https://inverternetwork.slack.com/archives/C08F45T4J8H/p1744622788167199

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -81,17 +108,32 @@ contract LM_PC_FundingPot_v1 is
...
+    /// @notice Maps round IDs to user addresses to contribution amounts
+    mapping(uint64 => mapping(address => uint)) private userContributions;
 
-    /// @notice The next available round ID.
-    uint64 private nextRoundId;
+    /// @notice Maps round IDs to total contributions
+    mapping(uint64 => uint) private roundTotalContributions;
+
+    /// @notice Maps round IDs to closed status
+    mapping(uint64 => bool) private roundClosed;
```
fabianschu: We have a naming convention for mappings, they should include what the key is and what the value is in the rounds like 
roundIdToRound
roundItToAccessCriteriaIdToPrivileges
roundIdToUserToContribution
etc

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -272,53 +355,243 @@ contract LM_PC_FundingPot_v1 is
...
+
+        AccessCriteriaPrivileges storage accessCriteriaPrivileges =
+            accessCriteriaPrivileges[roundId_][accessCriteriaId__];
+
+        accessCriteriaPrivileges.personalCap = personalCap_;
+        accessCriteriaPrivileges.overrideContributionSpan =
+            overrideContributionSpan_;
+        accessCriteriaPrivileges.start = start_;
+        accessCriteriaPrivileges.cliff = cliff_;
+        accessCriteriaPrivileges.end = end_;
```
fabianschu: This is shadowing an existing state var, but the state var should be renamed anyways to conform with our conventions :)

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -154,40 +195,82 @@ contract LM_PC_FundingPot_v1 is
     }
 
     /// @inheritdoc ILM_PC_FundingPot_v1
-    function getRoundAccessCriteria(uint64 roundId_, uint8 id_)
+    function getRoundAccessCriteria(uint64 roundId_, uint8 id_, address user_)
```
fabianschu: As far as I understand this function has been changed so that now it fulfills two purposes:
- getter for access criteria
- access control check for contributors

I would find it cleaner to separate the two functions: 1 .have getRoundAccessCriteria just return the access criteria and have a separate validation function to check if a user can contribute

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -154,40 +195,82 @@ contract LM_PC_FundingPot_v1 is
...
+
+    /// @inheritdoc ILM_PC_FundingPot_v1
+    function getRoundAccessCriteriaPrivileges(
+        uint64 roundId_,
+        uint8 accessCriteriaId__
+    )
+        external
+        view
+        returns (
+            bool isRoundOpen_,
```
fabianschu: This is related to access control (who can contribute?) and not to privileges (what are the contribution terms?). Why is it needed here?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -272,53 +355,243 @@ contract LM_PC_FundingPot_v1 is
...
+            AccessCriteriaType(accessCriteriaId_);
+        round.accessCriterias[accessCriteriaId_].accessCriteriaType =
+            accessCriteriaType;
+        round.accessCriterias[accessCriteriaId_].nftContract = nftContract_;
+        round.accessCriterias[accessCriteriaId_].merkleRoot = merkleRoot_;
+
+        for (uint i = 0; i < allowedAddresses_.length; i++) {
+            round.accessCriterias[accessCriteriaId_].allowedAddresses[allowedAddresses_[i]]
+            = true;
+        }
```
fabianschu, line 488: Right now we can only add to the list of allowed addresses, I think we should also be able to remove addresses. Maybe just add another function that takes a list of `permittedAddresses` and sets the bools to false in the mapping for these addresses?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -272,53 +355,243 @@ contract LM_PC_FundingPot_v1 is
...
+        uint capByMerkle_,
+        uint capByList_,
+        bool overrideContributionSpan_,
+        uint start_,
+        uint cliff_,
+        uint end_
+    ) external onlyModuleRole(FUNDING_POT_ADMIN_ROLE) {
         Round storage round = rounds[roundId_];
 
+        uint highestCap;
```
fabianschu: Conceptually, I don't understand what this highest cap var is necessary for.

fabianschu: Ah, I think I understand what you are trying to achieve here: are you somehow summing up what a user can contribute who is part of multiple access criteria?

If so, this is not necessary. I think most of the if (all?) of the if statements in here can be removed. I might be missing something though, so if you have doubts please voice them :)

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -154,40 +195,82 @@ contract LM_PC_FundingPot_v1 is
...
+        Round storage round = rounds[roundId_];
+        AccessCriteria storage accessCriteria =
+            round.accessCriterias[accessCriteriaId__];
+
+        if (accessCriteria.accessCriteriaType == AccessCriteriaType.OPEN) {
+            return (true, 0, false, 0, 0, 0);
+        }
+
+        // Store the privileges in a local variable to reduce stack usage.
+        AccessCriteriaPrivileges storage priviledges =
```
fabianschu: typo (:

fabianschu (changes requested):

src/modules/logicModule/LM_PC_FundingPot_v1.sol
JeffreyJoel: That is what is happening here, initially we did not need the user param, because we were returning an array of allowed addresses. Now we are making use of a mapping, so we need to return whether or not a user is part of the allowed addresses.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
JeffreyJoel: I don't understand this, can you please explain further. Is there a better way to return the highest cap a user can contribute?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
fabianschu: I understand the intention, but I think in this case the getter function should just return a bool that indicates that the access criteria is determined by a list. Wdyt?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -426,68 +430,48 @@ contract LM_PC_FundingPot_v1 is
         emit AccessCriteriaEdited(roundId_, accessCriteriaId_);
     }
 
+    function removeAccessCriteriaAddressesForRound(
```
fabianschu: Suggestion for name `removeAllowlistedAddresses`

src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -167,6 +167,14 @@ interface ILM_PC_FundingPot_v1 is IERC20PaymentClientBase_v2 {
         uint64 roundId_, uint timestamp_, uint totalContributions_
     );
 
+    /// @notice Emitted when addresses are removed from an access criteria's allowed list.
+    /// @param  roundId_ The ID of the round.
+    /// @param  accessCriteriaId_ The ID of the access criteria.
+    /// @param  addressesRemoved_ The addresses that were removed from the allowlist.
+    event AccessCriteriaAddressesRemoved(
```
fabianschu: Suggestion for name `AllowlistedAddressesRemoved`

src/modules/logicModule/LM_PC_FundingPot_v1.sol
fabianschu, line 1: I think we should also have an external function for _checkAccessCriteriaEligibility so that a UI can easily validate if a connected wallet is eligible, wdyt?

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
```diff
@@ -911,32 +993,52 @@ contract LM_PC_FundingPot_v1_Test is ModuleTest {
...
+            accessCriteriaEnum,
+            nftContract,
+            merkleRoot,
+            allowedAddresses
+        );
+
+        (
+            bool isOpen,
+            address retreivedNftContract,
+            bytes32 retreivedMerkleRoot,
```
fabianschu: Typo (:

test/modules/logicModule/LM_PC_FundingPot_v1.t.sol
fabianschu, line 1: The tests can and should be improved to fuzzing where possible and to have higher coverage generally:
image

fabianschu (changes requested):

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -154,40 +196,146 @@ contract LM_PC_FundingPot_v1 is
...
-        view
-        returns (uint8 accessCriteriaCount_)
-    {
-        return roundIdtoAccessId[roundId_];
+    function isRoundClosed(uint64 roundId_) external view returns (bool) {
+        return roundIdToClosedStatus[roundId_];
+    }
+
+    /// @inheritdoc ILM_PC_FundingPot_v1
+    function getUserEligibility(
```
fabianschu, line 285: Right now this function is a bit problematic. It returns this RoundUserEligibility struct which doesn't really help the UI to understand if a user can contribute to a given accessCriteriaId.

I think this function should let the user specify
- roundId
- accessCriteriaId
- user
- merkleProof

and return _how much the user can still contribute_ to the accessCriteriaId..

fabianschu (changes requested):

fabianschu (approved):
Great, looks good to me and can be merged. The function needs to be tested still but can be done together with the other tests.

PR #736, opened 2025-04-08
Feat: Script Improvements

marvinkruse:
This PR will contain the following:
* Script to easily deploy modules
* Script to easily update modules
* Script that deploys the v1 version of the protocol without needing to switch to the old tag
  * Includes a new deployment script that does a full deployment with that in mind

### **Script to easily deploy modules**
The idea is to have a simple command that does all of the necessary steps for you under the hood. Called via (for example):
```bash
forge script script/utils/DeployNewModule.s.sol --rpc-url $ETHEREUM_SEPOLIA_RPC_URL --sig "run(string,uint,uint,uint)" "FM_DepositVault_v1" 3 0 0
```
This would deploy the `FM_DepositVault_v1` as a new module in version 3.0.0 to Ethereum Sepolia and add it to the factory. Output:
Screenshot 2025-04-08 at 14 59 10

### **Script to easily update modules**
Once you have a deployed module it was very annoying to update it accordingly. This script handles that for you in one command. Called via (for example):
 ```bash
forge script script/utils/UpgradeModule.s.sol --rpc-url $ETHEREUM_SEPOLIA_RPC_URL --sig "run(string,uint,uint,uint)" "FM_DepositVault_v1" 1 2 8
```
This would update the `FM_DepositVault_v1` to version 1.2.8. You can also omit the minor and patch version, in which case the script will just increase the minor version and reset the patch version - i.e. if the current version is 1.2.8 it would deploy as 1.3.0.

Output:
Screenshot 2025-04-08 at 14 59 41

### DIFF
--- a/script/deploymentSuite/ProtocolConstants_v1.s.sol
+++ b/script/deploymentSuite/ProtocolConstants_v1.s.sol
@@ -213,4 +213,31 @@ contract ProtocolConstants_v1 is Script {
         // Set to 0 for Local Deployments
         return 0x0000000000000000000000000000000000000000;
     }
+
+    function currentNetworkIsMainnet() public view returns (bool) {
+        // Checks whether the current network is a mainnet
+        uint chainId = block.chainid;
+        for (uint i = 0; i < deployedMainnets.length; i++) {
+            if (chainId == deployedMainnets[i]) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    function currentNetworkIsTestnet() public view returns (bool) {
+        // Checks whether the current network is a testnet
+        uint chainId = block.chainid;
+        for (uint i = 0; i < deployedTestnets.length; i++) {
+            if (chainId == deployedTestnets[i]) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    function currentNetworkIsLocal() public view returns (bool) {
+        // Checks whether the current network is neither a mainnet nor a testnet
+        return !currentNetworkIsMainnet() && !currentNetworkIsTestnet();
+    }
 }
--- a/script/deploymentSuite/ProtocolConstants_v1.s.sol
+++ b/script/deploymentSuite/ProtocolConstants_v1.s.sol
@@ -57,7 +57,7 @@ contract ProtocolConstants_v1 is Script {
     // 137 = Polygon PoS
     // 1101 = Polygon zkEVM
     // 43114 = Avalanche C-Chain
-    uint[] public mainnets = [10, 137, 1101, 43_114];
+    uint[] public deployedMainnets = [10, 137, 1101, 43_114];
 
     // Testnets
     // 2442 = Polygon zkEVM Cardona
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: script/deploymentSuite/ProtocolConstants_v1.s.sol [hunk 1]
```diff
@@ -34,13 +36,38 @@ contract ProtocolConstants_v1 is Script {
...
+    address public deployedModuleFactory;
+    address public deployedGovernor;
+    address public deployedReverter;
+
+    // Chain IDs for Networks with Deployments
+    // If the current chainid is not part of any array, we assume that we are working locally.
+    uint[] public mainnets = [10, 137, 1101];
+    uint[] public testnets = [2442, 80_002, 84_532, 11_155_111, 11_155_420];
+
+    // Internal Storage for the deployment addresses (hardcoded as they don't change)
```
FHieser: Can you add a section here, that because of the salted deployments these stay the same over all chains. When I first looked at this I was pretty confused xD

script/utils/UpgradeModule.s.sol
```diff
@@ -0,0 +1,169 @@
...
+
+    function _getCurrentVersion(string memory moduleName, uint majorVersion)
+        internal
+        returns (IInverterBeacon_v1, uint, string memory)
+    {
+        ModuleFactory_v1 moduleFactory = ModuleFactory_v1(deployedModuleFactory);
+        (IInverterBeacon_v1 beacon,) = moduleFactory.getBeaconAndId(
+            IModule_v1.Metadata(
+                majorVersion,
+                0,
```
FHieser, line 88: Do we always start with 0 for the minor or Patch version?
If we deploy on a new chain this might change right?

script/utils/UpgradeModule.s.sol
```diff
@@ -0,0 +1,169 @@
...
+        );
+        vm.stopBroadcast();
+
+        console2.log("\t%s Implementation: %s", moduleName, implementation);
+        console2.log("----------------------------------------");
+        console2.log("Step (2) Set new Version in Beacon...");
+        // Register the metadata
+        Governor_v1 governor = Governor_v1(deployedGovernor);
+        vm.startBroadcast(deployerPrivateKey);
+        governor.forceUpgradeBeaconAndRestartImplementation(
```
FHieser: Directly forcing it goes against our policy doesnt it?

marvinkruse: This script was built for testnet only so far, I didn't use it for mainnet/have a different process there. But good point, this isn't clear by just looking at the script. I should probably just add an if-branch here to only does this if it's testnet or local.

script/deploymentSuite/ProtocolConstants_v1.s.sol
marvinkruse: Sure, I don't really get the confusion but I've added something :D

script/utils/UpgradeModule.s.sol
marvinkruse: No, `getBeaconAndId` does not care about minor or patch version, just about the major version. So we can set them to zero here for this read action.

script/utils/UpgradeModule.s.sol
marvinkruse: Added this to the script to handle mainnet differently 👍 good point!

script/deploymentSuite/ProtocolConstants_v1.s.sol
marvinkruse: added it 🫡

FHieser (approved):

PR #737, opened 2025-04-09
[AUDIT] Feature: CrossChain PP Base & Everclear CrossChain PP

Zitzak:
## Audit Scope
This PR introduces two new contracts to the `dev` branch for which an Audit is requested. Those two contracts are as following:
1. **PP_CrossChainBase_v1**: An abstract contract implementing the base functionalities for a crosschain payment processor.
2. **PP_Everclear_CrossChain_v1**: An implementation contract integrated with the [Everclear bridge](https://docs.everclear.org/), which adds implementation specific functionalities to the PP_CrossChainBase.

*Both contracts have unit tests*


src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
```diff
@@ -0,0 +1,429 @@
...
+        validClient(address(client_))
+    {
+        // Get the payment orders from the payment client.
+        IERC20PaymentClientBase_v2.PaymentOrder[] memory orders;
+        (orders,,) = client_.collectPaymentOrders();
+
+        // Process each payment order.
+        for (uint i = 0; i < orders.length; i++) {
+            if (!_validPaymentOrder(orders[i])) {
+                revert
```
omega-audits, line 163: It could be safer to skip invalid orders instead of reverting, as revet will get the system essentially stuck, which would be a risk in case users are allowed to add payment orders in a way which is not fully validated

src/modules/paymentProcessor/abstracts/PP_CrossChainBase_v1.sol
```diff
@@ -0,0 +1,243 @@
...
+    function _claimPreviouslyUnclaimable(
+        address client_,
+        address token_,
+        address paymentReceiver_
+    ) internal virtual {
+        address sender = _msgSender();
+        uint amount = _unclaimableAmountsForRecipient[client_][token_][sender];
+        delete _unclaimableAmountsForRecipient[client_][token_][sender];
+
+        IERC20(token_).transfer(paymentReceiver_, amount);
```
omega-audits: Should use safeTransfer

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
```diff
@@ -0,0 +1,429 @@
...
+        if (unclaimableAmount == 0 || order_.amount > unclaimableAmount) {
+            revert Module__PP_CrossChain__InvalidUnclaimableAmount();
+        }
+
+        // Subtract the order amount from the unclaimable amount.
+        _unclaimableAmountsForRecipient[client_][order_.paymentToken][recipient_]
+        -= order_.amount;
+
+        // Delete the unclaimable amount mapping entry if the amount is 0.
+        if (
```
omega-audits: The entry is just an amount so if the amount is 0 delete will do nothing. Just remove this check

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
```diff
@@ -0,0 +1,429 @@
...
+        address client_,
+        address recipient_,
+        IERC20PaymentClientBase_v2.PaymentOrder memory order_
+    ) external virtual {
+        uint unclaimableAmount =
+            unclaimable(client_, order_.paymentToken, recipient_);
+
+        // Validate that receipient has an unclaimable amount and the order
+        // amount provided is not greater than the unclaimable amount.
+        if (unclaimableAmount == 0 || order_.amount > unclaimableAmount) {
```
omega-audits: It's not enough to check the order amount is correct here.

With the current logic, anyone can call here on behalf of anyone else with any order data. Ie. anyone can steal anyone else's unclaimed amounts. You should probably save the actual failed orders so you can use them instead of accepting here any order.

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
```diff
@@ -0,0 +1,429 @@
...
+    /// @dev    The tokens must be transferred to the payment processor first because
+    ///         the bridge contract will later call this contract to execute the
+    ///         cross-chain transfer.
+    /// @param  order_ The payment order details.
+    /// @param  client_ The payment client address.
+    function _transferTokenAndApproveToBridge(
+        IERC20PaymentClientBase_v2.PaymentOrder memory order_,
+        address client_
+    ) internal virtual {
+        IERC20(order_.paymentToken).transferFrom(
```
omega-audits: Should use safeTransferFrom

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
omega-audits: In general the required dependency that all orders must be successful is somewhat risky to get payments stuck for everyone

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
```diff
@@ -0,0 +1,429 @@
...
+        // Process each payment order.
+        for (uint i = 0; i < orders.length; i++) {
+            if (!_validPaymentOrder(orders[i])) {
+                revert
+                    IERC20PaymentClientBase_v2
+                    .Module__ERC20PaymentClientBase__InvalidPaymentOrder();
+            }
+            // Transfer the token for the order from the payment client into
+            // the payment processor.
+            _transferTokenAndApproveToBridge(orders[i], address(client_));
```
omega-audits, line 169: It could make sense to do this action of transfer and approve of the tokens for all orders combined (unless you fully remove the potential atomic failures in the action)

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
omega-audits: Otherwise transfers may fail but users will get credited for it with unclaimable amounts

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
```diff
@@ -0,0 +1,429 @@
...
+            // Execute the bridge transfer.
+            _executeBridgeTransfer(orders[i]);
+        }
+    }
+
+    /// @inheritdoc IPP_Everclear_CrossChain_v1
+    function retryFailedBridgeTransfer(
+        address client_,
+        address recipient_,
+        IERC20PaymentClientBase_v2.PaymentOrder memory order_
```
FHieser: Creating a PaymentOrder from scratch might have implications for changed targets values that where not intended by the original PaymentClient.
I would for starters keep it simple and just allow for claiming funds that were unclaimable before here, as we dont really know yet how much need there will be for this functionality in the first place. I estimate that the Bridge Transfer Failed issue will be more of a CustomerSupport Problem, so I would focus more on proper documentation in the first place and communicate that this will need to be handled from that side. 

Also in case we want to keep it: the paymentOrder does not get the validPaymentOrder check

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
FHieser, line 1: Just for my understanding:
The goal of the PP Crosschain is more of a POC as we currently dont have any PaymentClients that allow for Crosschain Targets, right?

src/modules/paymentProcessor/abstracts/PP_CrossChainBase_v1.sol
FHieser, line 1: @Zitzak and I had a talk about this. This comment is more of FYI
This contract is essentially a PaymentProcessor Base contract (like the BondingCurveBase)
With some small additions (paymentIdToBridge etc)
I think it makes sense to finally abstract the Base into its own contract, which we will Hopefully tackle in the future.
Tagging @0xNuggan and @marvinkruse for visibility

src/modules/paymentProcessor/abstracts/PP_CrossChainBase_v1.sol
```diff
@@ -0,0 +1,243 @@
...
+    mapping(
+        address paymentClient
+            => mapping(
+                address token => mapping(address recipient => uint amount)
+            )
+    ) internal _unclaimableAmountsForRecipient;
+
+    /// @notice Mapping of payment IDs to bridge data.
+    mapping(uint paymentId => bytes bridgeData) internal _paymentIdToBridgeData;
+
```
FHieser, line 100: 9rk4ad

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
```diff
@@ -0,0 +1,429 @@
...
+                == 0
+        ) {
+            delete _unclaimableAmountsForRecipient[client_][order_.paymentToken][recipient_];
+        }
+        // Execute the bridge transfer.
+        _executeBridgeTransfer(order_);
+    }
+
+    /// @inheritdoc IPaymentProcessor_v2
+    function validPaymentOrder(
```
FHieser, line 176: View Function

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
```diff
@@ -0,0 +1,429 @@
...
+            }
+        }
+
+        return data_.length == requiredDataLength
+            && (flagsValue & (1 << FLAG_MAX_FEE)) != 0
+            && (flagsValue & (1 << FLAG_TTL)) != 0;
+    }
+
+    /// @dev    Gap for possible future upgrades.
+    uint[50] private __gap;
```
FHieser: 0.o

FHieser (commented):
Apart from the comments LGTM

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
```diff
@@ -0,0 +1,429 @@
...
+
+    // -------------------------------------------------------------------------
+    // Internal Functions
+
+    /// @notice Validates the payment order.
+    /// @param  order_ The payment order to validate.
+    /// @return valid_ True if the payment order is valid, false otherwise.
+    function _validPaymentOrder(
+        IERC20PaymentClientBase_v2.PaymentOrder memory order_
+    ) internal virtual returns (bool valid_) {
```
jellegerbrandy, line 190: mark as view

src/modules/paymentProcessor/abstracts/PP_CrossChainBase_v1.sol
```diff
@@ -0,0 +1,243 @@
...
+        address client_,
+        address token_,
+        address paymentReceiver_
+    ) internal virtual {
+        address sender = _msgSender();
+        uint amount = _unclaimableAmountsForRecipient[client_][token_][sender];
+        delete _unclaimableAmountsForRecipient[client_][token_][sender];
+
+        IERC20(token_).transfer(paymentReceiver_, amount);
+        emit TokensReleased(paymentReceiver_, address(token_), amount);
```
jellegerbrandy, line 207: may want to add the msg.sender in the event - it seems important to know who's tokens were claimed

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
Zitzak: I believe you brought this up with the other payment processor as well and it's a valid point. However when adding a payment order through a payment client, calling the function `_addPaymentOrder()`, it will always call the Payment processor's public `validPaymentOrder()` function, ensuring that no invalid order can be added. Because of that reason we have the revert, as it would point to the fact that more is going wrong then just failed order. Hope that made sense @omega-audits

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
Zitzak: It can't be marked view, as the function `_validPaymentToken()` in the `PP_CrossChainBase` is flagged a possible state change

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
Zitzak: It can't be marked view, as the function _validPaymentToken() in the PP_CrossChainBase is flagged a possible state change

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
Zitzak: I would opt for removing this function in this case, as I agree that there are a lot of assumptions here on how the bridge handles failed transfers.

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
Zitzak: See comment above from @FHieser. I'm opting here for removing the function all together.

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
Zitzak: @FHieser happy to have your input here as well

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
Zitzak: As my current choice would be to keep the revert for invalid order, I believe we indeed need to keep the single approval/transfer. Wdyt?

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
Zitzak: Agree. The inital idea was to also develop a PC which has crosschain capabilities. However with the change of strategic vertical this is not planned atm

src/modules/paymentProcessor/abstracts/PP_CrossChainBase_v1.sol
Zitzak: As this is a Payment Process event, the change here would effect quite a lot of code. I do agree however with your point. @FHieser any thoughts here?

FHieser: This is essentially an extension of the amountPaid doesnt have enough info/open redemption amount from Queued PP [here](https://github.com/InverterNetwork/contracts/pull/704#discussion_r2047447327).
As we still dont know how extensively the claimUnclaimableAmount functioanlity will be used and we decided already to tackle this (we tracked this internally via Issue) I think it should be fine if we leave it as is for now and go for it later. Just add this to the created issue as a reference point and we are good to go.

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
FHieser: In addition to what @Zitzak said:
Every processPayments() call is currently preceded by the according addPaymentOrder call in the same transaction, which ensures that the PaymentOrders must be valid. In that sense we actually do the check multiple times, which definitely isnt ideal. Reverting seems correct here, as it currently would prevent the orginal faulty PaymentClient function to run through (faulty in the sense that it produced a invalid paymentOrder in the first place).
For context: The validPaymentOrder check was a later addition to the payment system and could have probably be handled a bit better. As it currently stands this coding except for costing some gas is not harmful in any way, I would (assuming my statement form above stands, @Zitzak I currently dont know if the new Oracle Fundingmanager changes that) say we just leave it as is and bring this up internally, if we can do something about the double validation.

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
omega-audits: But if you wan to revert in case any of the orders fail, that is actually a better reason to do the combined approved/ transfer for all of them at the end instead of for each order no?

src/modules/paymentProcessor/PP_Everclear_CrossChain_v1.sol
```diff
@@ -0,0 +1,427 @@
...
+    /// @return ttl_ The TTL.
+    function _getEverclearMaxFeeAndTTL(bytes32 flags_, bytes32[] memory data_)
+        internal
+        pure
+        virtual
+        returns (uint24 maxFee_, uint48 ttl_)
+    {
+        uint relativeIndex = 0;
+
+        for (uint8 i = 0; i <= FLAG_TTL; ++i) {
```
Zitzak, line 379: Wouldn't the retrieval of the values in the data array be wrong if e.g. we have a payment order that has flags and an array of 5 long? So in this case, where the wrong payment order might have been sent to this payment processor, the PP will take the first and second array entry as values. I believe your version which you implemented in the streaming PP would prevent this?

0xNuggan (approved):

PR #739, opened 2025-04-17
feature: deployment script for Piku

Zitzak:
Deployment script which deploys and setup a queue based PIM. The script validates all the deployment and setup at each step


dev.env
```diff
@@ -62,4 +62,44 @@ export OP_ETHERSCAN_API_KEY=ABC123ABC123ABC123ABC123ABC123ABC1
 # In general, the command to run a deployment script will look like this:
 # forge script script/deploymentScript/DeploymentScript.s.sol --rpc-url $SEPOLIA_RPC_URL --chain-id 11155111 --private-key $WALLET_DEPLOYER_PK --etherscan-api-key $ETHERSCAN_API_KEY --verify --broadcast --legacy -vvv
 
+
```
marvinkruse: In general I would probably have put this into a new env file, since this is very much specific to this specific script and not very "generalized". maybe like `usp.env` or whatever. I'd probably not have this in our repo though and mainly just documented in the `workflowDeploymentAndSetupScripts/` folder. But I can be disagreed with here, especially if you (@Zitzak) say that we should just push this out like it is now and adjust later.

dev.env
```diff
@@ -62,4 +62,44 @@ export OP_ETHERSCAN_API_KEY=ABC123ABC123ABC123ABC123ABC123ABC1
...
+# ------------------------------------------------------------------------------
+# Workflow Deployment Parameters
+
+# USP Workflow Deployment Parameters @todo change name of Workflow
+# Issuance Token
+export TOKEN_NAME_STRING=Test USP Yield Optimized Stablecoin
```
marvinkruse: Are these verified to work? on some systems this is a bit funky with strings and spaces so this could turn into a "Test" instead of "Test USP Yield Optimized Stablecoin". So i'd probably wrap it in single quotes to be safe.

marvinkruse: At least any strings, especially if they have spaces or other special characters like `-` or so. The rest can stay like it is.

script/workflowDeploymentAndSetupScript/DeployAndSetupQueueBasedPIM.s.sol
```diff
@@ -0,0 +1,862 @@
...
+    "@fm/oracle/FM_PC_Oracle_Redeeming_v1.sol";
+import {LM_Oracle_Permissioned_v1} from "@lm/LM_Oracle_Permissioned_v1.sol";
+import {IOrchestrator_v1} from
+    "src/orchestrator/interfaces/IOrchestrator_v1.sol";
+import {IModule_v1} from "src/modules/base/IModule_v1.sol";
+import {DeploymentScript} from "script/deploymentScript/DeploymentScript.s.sol";
+
+// External
+import "forge-std/Script.sol";
+import {console2} from "forge-std/console2.sol";
```
marvinkruse: Should already be imported via the line above (`Script.sol`)

script/workflowDeploymentAndSetupScript/DeployAndSetupQueueBasedPIM.s.sol
```diff
@@ -0,0 +1,862 @@
...
+    address internal _independentUpdateAdmin;
+
+    // Orchestrator
+    IOrchestrator_v1 internal _orchestrator;
+
+    // Workflow Modules
+    FM_PC_Oracle_Redeeming_v1 _fundingManager;
+    AUT_Roles_v1 _authorizer;
+    PP_Queue_ManualExecution_v1 _paymentProcessor;
+    LM_Oracle_Permissioned_v1 _oracleModule;
```
marvinkruse: No functional difference, just for consistency.
```suggestion
    FM_PC_Oracle_Redeeming_v1 internal _fundingManager;
    AUT_Roles_v1 internal _authorizer;
    PP_Queue_ManualExecution_v1 internal _paymentProcessor;
    LM_Oracle_Permissioned_v1 internal _oracleModule;
```

script/workflowDeploymentAndSetupScript/DeployAndSetupQueueBasedPIM.s.sol
```diff
@@ -0,0 +1,862 @@
...
+    // -------------------------------------------------------------------------
+    // Setup functions
+
+    function _setupWorkflowAdmin() internal {
+        console2.log("Administrative Setup");
+        console2.log("--------------------");
+
+        console2.log("  Setting up workflow admin...");
+        vm.startBroadcast(deployerPrivateKey);
+        _authorizer.grantGlobalRole(_authorizer.getAdminRole(), _workflowAdmin);
```
marvinkruse: Its still a bit confusing when using the current authorizer, but here you should just use the OZ version `grantRole`, as we do not want any generation, since the admin role needs to be the zero (default admin) role. Otherwise this would change it.

script/workflowDeploymentAndSetupScript/DeployAndSetupQueueBasedPIM.s.sol
```diff
@@ -0,0 +1,862 @@
...
+        require(
+            _authorizer.checkForRole(
+                _authorizer.generateRoleId(
+                    address(_orchestrator), _authorizer.getAdminRole()
```
marvinkruse: After the change above, you'd also just use `_authorizer.getAdminRole()` here instead of generating.

script/workflowDeploymentAndSetupScript/DeployAndSetupQueueBasedPIM.s.sol
marvinkruse: The default/overall admin is basically an exception to the whole generation thing.

script/workflowDeploymentAndSetupScript/DeployAndSetupQueueBasedPIM.s.sol
```diff
@@ -0,0 +1,862 @@
...
+                _workflowAdmin
+            ) == true,
+            "Workflow admin not set correctly"
+        );
+        console2.log("  * Workflow admin: ", _workflowAdmin);
+        console2.log("  [OK] Workflow admin configured");
+
+        // Revoke admin role from deployer
+        console2.log("  Revoking admin role from deployer...");
+        _authorizer.revokeGlobalRole(_authorizer.getAdminRole(), deployer);
```
marvinkruse: Same change here, use OZ's `revokeRole` instead of our global version.

script/workflowDeploymentAndSetupScript/DeployAndSetupQueueBasedPIM.s.sol
```diff
@@ -0,0 +1,862 @@
...
+        require(
+            _authorizer.checkForRole(
+                _authorizer.generateRoleId(
+                    address(_orchestrator), _authorizer.getAdminRole()
```
marvinkruse: Generating admin role still, change needed here.

script/workflowDeploymentAndSetupScript/DeployAndSetupQueueBasedPIM.s.sol
```diff
@@ -0,0 +1,862 @@
...
+        _authorizer.transferAdminRole(
+            _authorizer.generateRoleId(
+                address(_fundingManager), _fundingManager.getWhitelistRole()
+            ),
+            _authorizer.generateRoleId(
+                address(_fundingManager),
+                _fundingManager.getWhitelistRoleAdmin()
+            )
+        );
+        // @todo how to check if admin role has been set correctly
```
marvinkruse: Todo leftover, you are checking below already.

script/workflowDeploymentAndSetupScript/DeployAndSetupQueueBasedPIM.s.sol
```diff
@@ -0,0 +1,862 @@
...
+        _authorizer.transferAdminRole(
+            _authorizer.generateRoleId(
+                address(_fundingManager), _fundingManager.getQueueExecutorRole()
+            ),
+            _authorizer.generateRoleId(
+                address(_fundingManager),
+                _fundingManager.getQueueExecutorRoleAdmin()
+            )
+        );
+        // @todo how to check if admin role has been set correctly
```
marvinkruse: probably also a leftover? Or are you still looking for an actual check?

script/workflowDeploymentAndSetupScript/DeployAndSetupQueueBasedPIM.s.sol
```diff
@@ -0,0 +1,862 @@
...
+        _collateralToken = vm.envAddress("COLLATERAL_TOKEN_ADDRESS");
+        require(_collateralToken != address(0), "Collateral token not set");
+
+        // Oracle Based Funding Manager
+        // ---------------------------------------------------------------------
+        _projectTreasury = vm.envAddress("PROJECT_TREASURY_ADDRESS");
+        require(_projectTreasury != address(0), "Project treasury not set");
+
+        _buyFeeInBps = vm.envUint("BUY_FEE_IN_BPS");
+        // require(_buyFeeInBps > 0, "Buy fee not set");
```
marvinkruse: commented out atm, may be intentional but i can't tell.

script/workflowDeploymentAndSetupScript/DeployAndSetupQueueBasedPIM.s.sol
```diff
@@ -0,0 +1,862 @@
...
+
+        _maxBuyFeeInBps = vm.envUint("MAX_BUY_FEE_IN_BPS");
+        require(_maxBuyFeeInBps > 0, "Max buy fee not set");
+
+        _sellFeeInBps = vm.envUint("SELL_FEE_IN_BPS");
+        // require(_sellFeeInBps > 0, "Sell fee not set");
```
marvinkruse: commented out atm, may be intentional but i can't tell.

script/workflowDeploymentAndSetupScript/DeployAndSetupQueueBasedPIM.s.sol
```diff
@@ -0,0 +1,862 @@
...
+        require(
+            _queueOperatorRoleAdmin != address(0),
+            "Queue operator role admin not set"
+        );
+
+        // Workflow config
+        // ---------------------------------------------------------------------
+        _independentUpdateBool = vm.envBool("INDEPENDENT_UPDATE_BOOL");
+        if (_independentUpdateBool) {
+            _independentUpdateAdmin =
```
marvinkruse: there are definitely cases where someone might want to have this set to 0 while having the update bool set to `true` (un-updatable workflows), but you may still (for this usecase) want to check that the admin (if set, like in this branch) is not the zero address, as in their case they probably do not want that (i assume).

marvinkruse (changes requested):
Good work overall 🙌 Will check out the failing tests etc. now to see what is going on there.

dev.env
Zitzak: Can do this. If you say documented, does that mean just to have an example version of the env I would assume? I do think that the script and env needs to be available for other deployers as well right? @marvinkruse

dev.env
Zitzak: Good catch! Fixed it

script/workflowDeploymentAndSetupScript/DeployAndSetupQueueBasedPIM.s.sol
Zitzak: They don't want independent updates afaik. I will remove the if statement ser

dev.env
marvinkruse: yeah i probably would have added a Readme into the folder that contains a copy-pastable example for the required env file with example values. You think that makes sense? like in my mind we would have

* A `README.md` in the `workflowDeploymentAndSetupScript` folder listing the different scripts we have (currently just this one)
  * Like "hey we have these scripts for workflow deployments, see this one for queue based PIMs (link to the specific readme and script)"
* A Readme specifically for this script that is referenced in the main readme, that lists how to use it and what env is required for this, so here `QueueBasedPIM.md` or so

maybe that's overkill though, what do you think?

Zitzak: As we don't know which setup we will go with moving forward, I.e. a general script with flags, individual scripts for each workflow (might be a lot of options), I would go with a straight forward solution for now.

This could look like a readme and a env file in the directory for now. Wdyt? @marvinkruse

marvinkruse: yeah thats perfect fine for me! @Zitzak

script/workflowDeploymentAndSetupScripts/DeployAndSetupYieldBearingStableWorkflow.s.sol
```diff
@@ -0,0 +1,885 @@
...
+        _collateralToken = vm.envAddress("COLLATERAL_TOKEN_ADDRESS");
+        require(_collateralToken != address(0), "Collateral token not set");
+
+        // Oracle Based Funding Manager
+        // ---------------------------------------------------------------------
+        _projectTreasury = vm.envAddress("PROJECT_TREASURY_ADDRESS");
+        require(_projectTreasury != address(0), "Project treasury not set");
+
+        _buyFeeInBps = vm.envUint("BUY_FEE_IN_BPS");
+        require(_buyFeeInBps >= 0, "Buy fee not set");
```
marvinkruse: What does this check aim to do? (also the following ones)
Per definition a uint is either 0 or >0, so this check doesn't really tell us anything, unless I'm miss something ofc!

marvinkruse (changes requested):
lgtm overall, just that 1 check that i found. if I'm wrong or you consciously did it like this feel free to ignore this ofc.

script/workflowDeploymentAndSetupScripts/DeployAndSetupYieldBearingStableWorkflow.s.sol
Zitzak: You're right, it does not. I wanted to validate that an amount has been set by the deployer but this could of course also be found out through the error when trying to load the env variable

marvinkruse (approved):

PR #740, opened 2025-04-22
Funding Pot V1: Round Closure

zzzuhaibmohd:
# Funding Pot v1: Round Closure Implementation

## Overview

This pull request completes the **Funding Pot v1** implementation by adding the round closure logic. It enables:

- Purchasing bonding curve tokens at the end of a round
- Creating payment orders for contributors based on access criteria
- Supporting the above feature for both manual and auto round closure

This PR builds on the foundation laid by previous contributions around round creation and user contributions.

### Round Closure System
- Implemented logic to **buy bonding curve tokens** once the round ends.
- Created **payment orders** to distribute tokens based on access criteria.
- Added support for **manual** and **auto-closure** of rounds.
- Write **unit tests** covering key round closure scenarios.

> **Note:** Currently using a mock bonding curve token address for testing purposes.

## Notable Commits

- **Apr 18, 2025**: Add bonding curve token purchase and payment order creation logic  
- **Apr 19, 2025**: Add unit tests for round closure functionality


src/modules/logicModule/LM_PC_FundingPot_v1.sol
fabianschu, line 1: Without --via-ir this file throws a compilation error and will not go through our CI/CD

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -1020,23 +1034,151 @@ contract LM_PC_FundingPot_v1 is
...
+        uint totalContributions = _getTotalRoundContribution(roundId_);
+
+        address issuanceToken = address(
+            IBondingCurveBase_v1(
+                address(__Module_orchestrator.fundingManager())
+            ).getIssuanceToken()
+        );
+
+        uint balanceBefore = IERC20(issuanceToken).balanceOf(address(this));
+        IBondingCurveBase_v1(issuanceToken).buyFor(
```
fabianschu: This is using the wrong address. The `buyFor` function lives on the funding manager. I'm curious why the tests didn't catch this.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -1020,23 +1034,151 @@ contract LM_PC_FundingPot_v1 is
...
+                    flags |= bytes32(uint(1) << 2); // Flag 2 for cliff
+                    data[flagCount] = bytes32(cliff);
+                    flagCount++;
+                }
+
+                if (end > 0) {
+                    flags |= bytes32(uint(1) << 3); // Flag 3 for end
+                    data[flagCount] = bytes32(end);
+                    flagCount++;
+                }
```
fabianschu:
```suggestion
                if (start > 0) {
                    flags |= bytes32(1 << FLAG_START);
                    data[flagCount] = bytes32(start);
                    flagCount++;
                }

                if (cliff > 0) {
                    flags |= bytes32(1 << FLAG_CLIFF);
                    data[flagCount] = bytes32(cliff);
                    flagCount++;
                }

                if (end > 0) {
                    flags |= bytes32(1 << FLAG_END);
                    data[flagCount] = bytes32(end);
                    flagCount++;
                }
```

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -1020,23 +1034,151 @@ contract LM_PC_FundingPot_v1 is
...
+        for (uint i = 0; i < contributors.length; i++) {
+            address contributor = contributors[i];
+            uint contributorTotal =
+                roundIdToUserToContribution[roundId_][contributor];
+
+            if (contributorTotal == 0) continue;
+
+            // Calculate tokens for this contributor proportionally
+            uint contributorTokens =
+                (contributorTotal * tokensBought) / totalContributions;
```
fabianschu: This is unused

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -133,6 +134,16 @@ contract LM_PC_FundingPot_v1 is
     /// @notice Maps round IDs to closed status
     mapping(uint64 => bool) private roundIdToClosedStatus;
 
+    /// @notice Maps round IDs to bonding curve tokens bought
+    mapping(uint64 => uint) private roundTokensBought;
+
+    /// @notice Maps round IDs to contributors recipients
+    mapping(uint64 => EnumerableSet.AddressSet) private contributorsByRound;
```
fabianschu: Love this idea of using OZ's EnumerableSet

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -1020,23 +1034,151 @@ contract LM_PC_FundingPot_v1 is
...
-        // Emit event for round closure
         emit RoundClosed(
             roundId_, block.timestamp, roundIdToTotalContributions[roundId_]
         );
     }
 
+    /// @notice Creates payment orders for all contributors in a round based on their access criteria
+    /// @dev    Loops through all contributors and creates payment orders with appropriate vesting schedules
+    /// @param  roundId_ The ID of the round to create payment orders for
+    function _createPaymentOrdersForContributors(uint64 roundId_) internal {
```
fabianschu: This function definitely needs some kind of pagination mechanism to handle long lists of contributors and gas limit implications. 

I would be also fine with with `_createPaymentOrdersForContributors` not being called within the closeRound function and instead having an external function that does this that allows for pagination (specifying the number of contributors that should be processed at a time). 

This function should be callable by anyone if autoClosure is true and callable only by the FP Admin if autoClosure is false.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -1020,23 +1034,151 @@ contract LM_PC_FundingPot_v1 is
...
+                    start,
+                    cliff,
+                    end
+                );
+            }
+        }
+    }
+
+    function _buyBondingCurveToken(uint64 roundId_) internal {
+        uint totalContributions = _getTotalRoundContribution(roundId_);
```
fabianschu: Maybe add validation against totalContributions == 0 ?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -1020,23 +1034,151 @@ contract LM_PC_FundingPot_v1 is
...
+            uint contributorTotal =
+                roundIdToUserToContribution[roundId_][contributor];
+
+            if (contributorTotal == 0) continue;
+
+            // Calculate tokens for this contributor proportionally
+            uint contributorTokens =
+                (contributorTotal * tokensBought) / totalContributions;
+
+            for (
```
fabianschu, line 1141: I'm not a fan of the nested loop in here, since most contributors are expected to just contribute to one access criteria. But from the options I see, your approach is the best, I guess. Just a comment, no action required :)

fabianschu (changes requested):

src/modules/logicModule/LM_PC_FundingPot_v1.sol
zzzuhaibmohd: @fabianschu Can you explain more on this! I think may be our understanding is not correct!

The tests may be passing because we are using Mock contracts!

src/modules/logicModule/LM_PC_FundingPot_v1.sol
zzzuhaibmohd: @JeffreyJoel has fixed this now

fabianschu, line 1: This contract exceeds the code size limit

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -1088,91 +1142,139 @@ contract LM_PC_FundingPot_v1 is
...
+        AccessCriteriaPrivileges storage privileges =
+            roundItToAccessCriteriaIdToPrivileges[roundId_][accessCriteriaId_];
+
+        uint start = privileges.overrideContributionSpan
+            ? privileges.start
+            : round.roundStart;
+        uint cliff = privileges.overrideContributionSpan ? privileges.cliff : 0;
+        uint end = privileges.overrideContributionSpan
+            ? privileges.end
+            : round.roundEnd;
```
fabianschu: Why is the overrideContributionSpan influencing the vesting params?

fabianschu: overrrideContributionSpan should just enable users to contribute even if the regular contribution span is over (but round is not yet closed), but shouldn't affect the vesting schedule that is assigned to the privileges.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -1088,91 +1142,139 @@ contract LM_PC_FundingPot_v1 is
...
-        uint balanceBefore = IERC20(issuanceToken).balanceOf(address(this));
-        IBondingCurveBase_v1(issuanceToken).buyFor(
-            address(this), totalContributions, 0
+        // approve the fundingManager to spend the contribution token
+        IERC20(__Module_orchestrator.fundingManager().token()).approve(
+            address(__Module_orchestrator.fundingManager()), totalContributions
         );
+        uint balanceBefore = IERC20(issuanceToken).balanceOf(address(this));
+        IBondingCurveBase_v1(address(__Module_orchestrator.fundingManager()))
+            .buyFor(address(this), totalContributions, 1);
```
fabianschu: Can we use calculatePurchaseReturn, it gives you the amount that will be returned during buyFor. It is on IBondingCurveBase. Then you don't need to get the balance before and after. Then you can also replace the `1`, with the actual amount.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -646,12 +647,55 @@ contract LM_PC_FundingPot_v1 is
 
             _buyBondingCurveToken(roundId_);
 
-            _createPaymentOrdersForContributors(roundId_);
+            // Payment orders will be created separately via processContributorBatch
```
fabianschu: Where is this function `processContributorBatch`?

fabianschu: comment seems to be outdated :)

fabianschu (changes requested):

src/modules/logicModule/LM_PC_FundingPot_v1.sol
JeffreyJoel: We had to mock this function like this 
` function calculatePurchaseReturn(uint amount) public view returns (uint) {
        return amount;
    }` 
    for tests to pass. Is this ideal?

simulation.py
fabianschu, line 1: who dis? :D

test/utils/mocks/modules/logicModules/LM_PC_FundingPot_v1ERC20Mock.sol
```diff
@@ -0,0 +1,20 @@
...
+        _mint(to, value);
+    }
+
+    function token() public view returns (address) {
+        return address(this);
+    }
+
+    function getIssuanceToken() public view returns (address) {
+        return address(this);
+    }
```
fabianschu: These shouldn't be on the token

fabianschu (changes requested):

test/utils/mocks/modules/logicModules/LM_PC_FundingPot_v1ERC20Mock.sol
zzzuhaibmohd: Removed the mock contract, not using it any more in the tests

simulation.py
zzzuhaibmohd: not sure, may be pushed my mistake. removed it now!

fabianschu (approved):

PR #741, opened 2025-04-30
Feat: Funding Pot

fabianschu:


src/modules/logicModule/interfaces/ILM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,540 @@
...
+    error Module__LM_PC_FundingPot__RoundMustHaveEndTimeOrCap();
+
+    /// @notice Round end time must be after round start time.
+    error Module__LM_PC_FundingPot__RoundEndMustBeAfterStart();
+
+    /// @notice Round has already started and cannot be modified.
+    error Module__LM_PC_FundingPot__RoundAlreadyStarted();
+
+    /// @notice Hook function is required when a hook contract is provided.
+    error Module__LM_PC_FundingPot__HookFunctionRequiredWithContract();
```
0xNuggan: Unused error, seems a duplicate from the one below

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1294 @@
...
+    /// @return Boolean indicating if the round has reached its cap or time limit.
+    function _checkRoundClosureConditions(uint32 roundId_)
+        internal
+        view
+        returns (bool)
+    {
+        Round storage round = rounds[roundId_];
+        uint totalContribution = roundIdToTotalContributions[roundId_];
+        bool capReached =
+            round.roundCap > 0 && totalContribution == round.roundCap;
```
0xNuggan: should be >=

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1294 @@
...
+        Round storage round = rounds[roundId_];
+        AccessCriteria storage accessCriteria =
+            round.accessCriterias[accessCriteriaId_];
+
+        if (accessCriteria.accessCriteriaType == AccessCriteriaType.OPEN) {
+            return (
+                true,
+                accessCriteria.nftContract,
+                accessCriteria.merkleRoot,
+                true
```
0xNuggan: Should be false

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1294 @@
...
+        )
+    {
+        Round storage round = rounds[roundId_];
+        AccessCriteria storage accessCriteria =
+            round.accessCriterias[accessCriteriaId_];
+
+        if (accessCriteria.accessCriteriaType == AccessCriteriaType.OPEN) {
+            return (
+                true,
+                accessCriteria.nftContract,
```
0xNuggan: Both nftContract and merkleRoot shoul be set to 0 in this case

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1294 @@
...
+    /// @notice Validates the round parameters before editing.
+    /// @param  round_ The round to validate.
+    /// @dev    Reverts if the round parameters are invalid.
+    function _validateEditRoundParameters(Round storage round_) internal view {
+        if (round_.roundEnd == 0 && round_.roundCap == 0) {
+            revert Module__LM_PC_FundingPot__RoundNotCreated();
+        }
+
+        if (block.timestamp > round_.roundStart) {
+            revert Module__LM_PC_FundingPot__RoundAlreadyStarted();
```
0xNuggan: A bit of a nitpick, but since this will also trigger if somebody tries to edit an already finished round, maybe we can change the name (or add an extra error if size allows for it?)

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1294 @@
...
+            roundCap_,
+            hookContract_,
+            hookFunction_,
+            autoClosure_,
+            globalAccumulativeCaps_
+        );
+    }
+
+    /// @inheritdoc ILM_PC_FundingPot_v1
+    function setAccessCriteriaForRound(
```
0xNuggan: Depending on the accessCriteria, the other values should be left at to zero in the AccessCriteria storage. Also, both this funcion and the one above can probably be refactored into a commons subfunction.

0xNuggan (changes requested):
In general looks good, there are a couple of things I pointed out. I also have still an open question about the caps/accesstypes, but I wanted to get this out first

src/modules/logicModule/LM_PC_FundingPot_v1.sol
fabianschu: I can't really come up with a more descriptive error since the idea is that the admin can edit as long as a round hasn't started. Imo the error exactly tells this story.

0xNuggan: Yeah, fair. My comment came from the thought of "if a round started and ended, it's weird to revert saying 'it started' ". But it makes sense, the condition is "rounds are editable before start". Let's leavi it like that.

0xNuggan (approved):

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1439 @@
...
+    {
+        return interfaceId_ == type(ILM_PC_FundingPot_v1).interfaceId
+            || super.supportsInterface(interfaceId_);
+    }
+
+    // --------------------------------------------------------------------------
+    // Constants
+
+    /// @notice The role that allows creating funding rounds.
+    bytes32 public constant FUNDING_POT_ADMIN_ROLE = "FUNDING_POT_ADMIN";
```
omega-audits, line 100: You will need to update it for the new authorization logic

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1439 @@
...
+            round.roundCap,
+            round.hookContract,
+            round.hookFunction,
+            round.autoClosure,
+            round.accumulationMode
+        );
+    }
+
+    /// @inheritdoc ILM_PC_FundingPot_v1
+    function getRoundAccessCriteria(uint32 roundId_, uint8 accessCriteriaId_)
```
omega-audits, line 225: Should revert if the roundId_ or its accessCriteriaId_ don't exist

omega-audits: (Unless returning all empty over revert is intended)

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1439 @@
...
+        uint32 roundId_,
+        uint8 accessCriteriaType_,
+        uint8 accessCriteriaId_, // Optional: 0 for new, non-zero for edit
+        address nftContract_,
+        bytes32 merkleRoot_,
+        address[] calldata allowedAddresses_
+    ) external onlyModuleRole(FUNDING_POT_ADMIN_ROLE) {
+        Round storage round = rounds[roundId_];
+
+        if (accessCriteriaType_ > MAX_ACCESS_CRITERIA_ID) {
```
omega-audits: Here it checks against the type not the id. According to the name and other uses of MAX_ACCESS_CRITERIA_ID, it should both check here against accessCriteriaId_ instead of accessCriteriaType_, and in if accessCriteriaId_ is 0 it should check roundIdToNextAccessCriteriaId won't be higher.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1439 @@
...
+            if (allowedAddresses_.length == 0) {
+                revert
+                    Module__LM_PC_FundingPot__MissingRequiredAccessCriteriaData();
+            }
+        }
+
+        // Clear all existing data to prevent stale data
+        round.accessCriterias[criteriaId].nftContract = address(0);
+        round.accessCriterias[criteriaId].merkleRoot = bytes32(0);
+        // @note: When changing allowlists, call removeAllowlistedAddresses first to clear previous entries
```
omega-audits, line 436: Shouldn't this be called here then to make sure it's removed?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1439 @@
...
+            round.accessCriterias[accessCriteriaId_].accessCriteriaType
+                == AccessCriteriaType.UNSET
+        ) {
+            revert Module__LM_PC_FundingPot__InvalidAccessCriteriaId();
+        }
+
+        _validateEditRoundParameters(round);
+
+        for (uint i = 0; i < addressesToRemove_.length; i++) {
+            unchecked {
```
omega-audits: What's this unchecked for? (there's no math operation inside here)

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1439 @@
...
+                unspentPersonalRoundCaps_[i];
+            uint32 currentProcessingRoundId = roundCapInfo.roundId;
+
+            // Enforcement: Round IDs in the array must be strictly increasing.
+            if (currentProcessingRoundId <= lastSeenRoundId) {
+                revert
+                    Module__LM_PC_FundingPot__UnspentCapsRoundIdsNotStrictlyIncreasing(
+                );
+            }
+
```
omega-audits: Could move lastSeenRoundId = currentProcessingRoundId; here and delete it from the 3 other places it's done below to avoid code repetition.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1439 @@
...
+    }
+
+    /// @inheritdoc ILM_PC_FundingPot_v1
+    function closeRound(uint32 roundId_)
+        external
+        onlyModuleRole(FUNDING_POT_ADMIN_ROLE)
+    {
+        Round storage round = rounds[roundId_];
+
+        if (round.roundEnd == 0 && round.roundCap == 0) {
```
omega-audits, line 565: Instead of this, you could just check if the round ID is lower than the current one in the counter (here and in other places this check is done)

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1439 @@
...
+        if (!roundIdToClosedStatus[roundId_]) {
+            revert Module__LM_PC_FundingPot__RoundNotClosed();
+        }
+
+        address[] memory contributors =
+            EnumerableSet.values(contributorsByRound[roundId_]);
+        uint contributorCount = contributors.length;
+
+        // Check batch size is not zero
+        if (batchSize_ == 0 || batchSize_ > contributorCount) {
```
omega-audits: It might make operations more convenient to set batchSize_ to contributorCount instead of reverting if it's bigger

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1439 @@
...
+                uint remainingRoundCap =
+                    effectiveRoundCap - totalRoundContribution;
+                if (adjustedAmount > remainingRoundCap) {
+                    adjustedAmount = remainingRoundCap;
+                }
+            }
+        }
+
+        // --- Personal Cap Check ---
+        {
```
omega-audits, line 937: Should have an if here to check if personal cap exists/ should be applied

src/modules/logicModule/LM_PC_FundingPot_v1.sol
omega-audits: Or check roundStart != 0 as it cannot be 0 when a round is added

src/modules/logicModule/LM_PC_FundingPot_v1.sol
omega-audits: Though it's a bit unclear why require it to be up to 4?
Was that meant to be limiting the type or the ID in fact?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
leeftk: I think we were trying to reduce contract size here and so adding the revert may actually push us over the limit, if it doesn't then I think changing it makes sense but if it does then we'll have to leave as is.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
leeftk: Nice find, we fixed this in the fixes PR

src/modules/logicModule/LM_PC_FundingPot_v1.sol
leeftk: We need to add a note here to be more explicit but we won't be fixing this for now.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
leeftk: hmmm now sure this is needed here's why

- userPersonalCap will be 0 (default value)
- If a user tries to contribute any amount (let's say 100 tokens)
- The code checks: userPreviousContribution >= userPersonalCap
- Since userPersonalCap is 0, and any contribution > 0, this will be true
- The transaction will revert with PersonalCapReached

src/modules/logicModule/LM_PC_FundingPot_v1.sol
jellegerbrandy: so this function si not sued internally, so there is no internal logic problem. 
To me, also returning `isList_ = true` when the AccessCriteriaType is equal to OPEN is unexpected, but it is impossible to say if this is problematic as we do not know the use case. What is the use case?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1438 @@
...
+        uint32 roundId_,
+        uint8 accessCriteriaId_,
+        bytes32[] memory merkleProof_,
+        address user_
+    )
+        external
+        view
+        returns (bool isEligible, uint remainingAmountAllowedToContribute)
+    {
+        if (accessCriteriaId_ > MAX_ACCESS_CRITERIA_TYPE) {
```
jellegerbrandy: so here you compare accessCriteriaId with MAX_ACCESS_CRITERIA_TYPE, while in setAccessCriteria, it is the accessCriteriaType_ argument (not the id) that is supposed to be smaller than MAX_ACCESS_CRITERIA_TYPE. 

Not sure what the original intention is here, but you should decide if the constnat MAX_ACCESS_CRITERIA_TYPE limits the ids or the type (I guess the latter)

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1438 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity 0.8.23;
```
jellegerbrandy, line 2: as a general remark, it is better (easier to review, easier to understand, easier to version control) to split the logic over diffferent classes/files rather than mixing all together in one huge file.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1438 @@
...
+    uint32 private roundCount;
+
+    /// @notice Stores all funding rounds by their unique ID.
+    mapping(uint32 => Round) private rounds;
+
+    /// @notice Stores all access criteria privilages by their unique ID.
+    mapping(
+        uint32 roundId
+            => mapping(uint8 accessCriteriaId_ => AccessCriteriaPrivileges)
+    ) private roundIdToAccessCriteriaIdToPrivileges;
```
jellegerbrandy, line 127: why put this in a separate mapping? is it not cleaner to have the priviliges part of the accessCriteria object instead of createing a separate mapping?

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1438 @@
...
+        uint8 accessCriteriaType_,
+        uint8 accessCriteriaId_, // Optional: 0 for new, non-zero for edit
+        address nftContract_,
+        bytes32 merkleRoot_,
+        address[] calldata allowedAddresses_
+    ) external onlyModuleRole(FUNDING_POT_ADMIN_ROLE) {
+        Round storage round = rounds[roundId_];
+
+        if (accessCriteriaType_ > MAX_ACCESS_CRITERIA_TYPE) {
+            revert Module__LM_PC_FundingPot__InvalidAccessCriteriaId();
```
jellegerbrandy: perhaps not the error you want to use here, as the id could be fine..

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1438 @@
...
+        _validateEditRoundParameters(round);
+
+        uint8 criteriaId;
+        bool isEdit = false;
+
+        // If accessCriteriaId_ is 0, create a new access criteria
+        // Otherwise, edit the existing one
+        if (accessCriteriaId_ == 0) {
+            unchecked {
+                criteriaId = ++roundIdToNextAccessCriteriaId[roundId_];
```
jellegerbrandy, line 399: if you have prblems witht he contract size, a small gain is to have a global counter for teh criteriaId instead of a spearate one for each round.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1438 @@
...
+            accessCriteriaType;
+
+        // Set only the relevant data based on the access criteria type
+        if (accessCriteriaType == AccessCriteriaType.NFT) {
+            round.accessCriterias[criteriaId].nftContract = nftContract_;
+        } else if (accessCriteriaType == AccessCriteriaType.MERKLE) {
+            round.accessCriterias[criteriaId].merkleRoot = merkleRoot_;
+        } else if (accessCriteriaType == AccessCriteriaType.LIST) {
+            // For LIST type, update the allowed addresses
+            for (uint i = 0; i < allowedAddresses_.length; i++) {
```
jellegerbrandy, line 455: it is important to note here that you are NOT overwriting the allowedlist here, but only adding to it. This is not obvious at all. I'd recomend to adda "addressesToRemove" parameter as well to the function signature to make this explicit.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
jellegerbrandy: this should not be problematic, as these data are read only if the relevant Type is set.
However, the logic is not consistent, and that can be confusing for someone who calls this function.
you could simplify this block of code (i.e. these reset lines plus the stuff on l 531f) by just these variables to the value of the parameters.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1438 @@
...
+        uint32 roundId_,
+        uint amount_,
+        uint8 accessCriteriaId_,
+        bytes32[] memory merkleProof_,
+        UnspentPersonalRoundCap[] calldata unspentPersonalRoundCaps_
+    ) external {
+        uint unspentPersonalCap = 0; // Initialize to 0
+        uint32 lastSeenRoundId = 0; // Tracks the last seen roundId to ensure strictly increasing order
+
+        // Process each previous round cap that the user wants to carry over
```
jellegerbrandy: I do not see where it is enforced taht the unspentPersonalRoundCaps are "previous" to the given roundId_. This is important, as in this way, they could include the roundId_ itself (and so the personal cap would be counted twice), or even after roundId_ (if roundId_ has overrideContributionSpan set to true)

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -0,0 +1,1438 @@
...
+                    roundCapInfo.merkleProof,
+                    user_
+                )
+            ) {
+                AccessCriteriaPrivileges storage privileges =
+                roundIdToAccessCriteriaIdToPrivileges[currentProcessingRoundId][roundCapInfo
+                    .accessCriteriaId];
+
+                uint userContribution =
+                    roundIdToUserToContribution[currentProcessingRoundId][user_];
```
jellegerbrandy: roundIdToUserToContribution does not seem to be updated for the currentPRocessingRoundId. That would mean that the personalCap of previous rounds can be used over and over again.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
zzzuhaibmohd: removed the if check, since it is not required for `accessCriteriaId`

src/modules/logicModule/LM_PC_FundingPot_v1.sol
zzzuhaibmohd: Changed the error string from `Module__LM_PC_FundingPot__InvalidAccessCriteriaId` to `Module__LM_PC_FundingPot__InvalidAccessCriteriaType`

src/modules/logicModule/LM_PC_FundingPot_v1.sol
zzzuhaibmohd: Do you have a proposed fix in mind as to how to fix this? some pseudo code would be helpful as well @jellegerbrandy

src/modules/logicModule/LM_PC_FundingPot_v1.sol
leeftk: thanks good to note but a little too late for us to change here, will keep noted for the next contract

src/modules/logicModule/LM_PC_FundingPot_v1.sol
leeftk: yes for sure good note but we will not be fixing this time around

src/modules/logicModule/LM_PC_FundingPot_v1.sol
JeffreyJoel: I have pushed a fix for this, and added a test as well

src/modules/logicModule/LM_PC_FundingPot_v1.sol
zzzuhaibmohd: Added the fix and supporting unit test as well!

src/modules/logicModule/LM_PC_FundingPot_v1.sol
zzzuhaibmohd: I have pushed a fix for this, and added a couple of tests.

PR #748, opened 2025-05-15
Feat: Dynamic cross round behavior & contract size

fabianschu:
**Changes**
* round contribution capacity for a user is limited by the access criteria with the highest possible personal cap
  * both for contribution to current round
  * and for the unspent cap calculation (accumulationMode == personal/all): here it is implemented such that it it expects the user to pass unspent capacities in ascending order by roundId, so that a user can only provide one unspent cap per round
* accumulative cross-round behavior can be configured to apply to only 1) personal caps, 2) total caps, 3) both, 4) none
* accumulative cross-round behavior can be restricted by admin retrospectively; admin can define from which round onwards accumulative behavior starts

### DIFF
--- a/src/modules/logicModule/LM_PC_FundingPot_v1.sol
+++ b/src/modules/logicModule/LM_PC_FundingPot_v1.sol
@@ -171,11 +171,10 @@ contract LM_PC_FundingPot_v1 is
     /// @dev    MUST call `__Module_init()`.
     /// @param  orchestrator_ The orchestrator contract.
     /// @param  metadata_ The metadata of the module.
-    /// @param  configData_ The config data of the module, comprised of:
     function init(
         IOrchestrator_v1 orchestrator_,
         Metadata memory metadata_,
-        bytes memory configData_
+        bytes memory
     ) external override(Module_v1) initializer {
         __Module_init(orchestrator_, metadata_);
         // Set the flags for the PaymentOrders (this module uses 3 flags).
--- a/src/modules/logicModule/LM_PC_FundingPot_v1.sol
+++ b/src/modules/logicModule/LM_PC_FundingPot_v1.sol
@@ -1036,8 +1036,7 @@ contract LM_PC_FundingPot_v1 is
         }
 
         // --- Personal Cap Check ---
-        // Skip personal cap check if adjustedAmount is already 0
-        if (adjustedAmount > 0) {
+        {
             uint userPreviousContribution =
                 roundIdToUserToContribution[roundId_][user_];
 
--- a/src/modules/logicModule/LM_PC_FundingPot_v1.sol
+++ b/src/modules/logicModule/LM_PC_FundingPot_v1.sol
@@ -184,6 +184,8 @@ contract LM_PC_FundingPot_v1 is
         flags |= bytes32(1 << FLAG_END);
 
         __ERC20PaymentClientBase_v2_init(flags);
+        // Explicitly initialize the global start round ID
+        globalAccumulationStartRoundId = 1;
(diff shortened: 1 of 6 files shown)


src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -221,33 +229,24 @@ contract LM_PC_FundingPot_v1 is
             bool isList_
         )
     {
-        Round storage round = rounds[roundId_];
         AccessCriteria storage accessCriteria =
-            round.accessCriterias[accessCriteriaId_];
+            rounds[roundId_].accessCriterias[accessCriteriaId_];
```
0xNuggan, line 232: I assume this refactor was done because of contract size restrictions, but I love it.

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -957,49 +1011,76 @@ contract LM_PC_FundingPot_v1 is
...
-                remainingRoundCap = effectiveRoundCap - totalRoundContribution;
-            }
-            if (adjustedAmount > remainingRoundCap) {
-                adjustedAmount = remainingRoundCap;
+                // If user tries to contribute a non-zero amount when cap is full, revert.
+                if (amount_ > 0) {
+                    revert Module__LM_PC_FundingPot__RoundCapReached();
+                }
+                // If user tries to contribute zero when cap is full, allow adjustedAmount = 0.
+                adjustedAmount = 0;
```
0xNuggan: we are controlling for amount_ != 0 in the parent function, so I think this step can be removed?

0xNuggan: I'm referring to '_contributeToRoundFor', line 888

fabianschu: Great find, thanks. I simplified the function in this commit:
https://github.com/InverterNetwork/contracts/pull/748/commits/d3fc1adc2b91e8dbeabe6f80004b7bb1f5489782

src/modules/logicModule/LM_PC_FundingPot_v1.sol
```diff
@@ -953,52 +1006,64 @@ contract LM_PC_FundingPot_v1 is
...
+                }
             }
         }
 
-        // Check and adjust for personal cap
-        uint userPreviousContribution =
-            roundIdToUserToContribution[roundId_][user_];
+        // --- Personal Cap Check ---
+        // Skip personal cap check if adjustedAmount is already 0
+        if (adjustedAmount > 0) {
```
0xNuggan: Thanks for the changes! One remaining question here is:
- On line 888 we control for empty deposits by checking that the user hasn't put "amount = 0". If it is so, we revert.
- On line 1057 we control for empty deposits by checking if the user's previous contribution is already greater/equal than their personal cap and if it is so, revert.

But, on this line (1040) if the adjustedAmount is already 0 we would forward that value back to the function, and allow for an empty deposit. Shouldn't this also revert?

fabianschu: Actually it is impossible that adjustedAmount == 0 at this point in the execution, so the check is redundant. The function would always revert if a user attempts to contribute zero. I remove the check here.

fabianschu: Changes in this commit: fd3f3e041c113d77054dc430742456001236cd00

0xNuggan (approved):

PR #752, opened 2025-05-20
[Audit] Feature/Authorizer-Update 

FHieser:
This is the audit feature branch for the Authorizer Update.
The following contracts are the main focus points of the Audit and contain internal refactors and central logic changes that impact other contracts. I recommend going from top to bottom through the contracts to understand the changes:

**Module:**
- src/modules/base/Module_v1.sol 

Removing all ModuleRole Modifiers and grantRole functions and replace them with permissioned modifier.

**Orchestrator**
- src/orchestrator/Orchestrator_v1

Adapt to also use permissioned functionality.

**AUT_Roles:**
- src/modules/authorizer/role/AUT_Roles_v1.sol

Removing the hashed Role id setup and replacing it with a straight up counter system
Adding a flexible function permission system, where function restrictions can be adapted on a individual basis

**AUT_TokenGated_Roles**
- src/modules/authorizer/role/AUT_TokenGated_Roles.sol

Adapt to new base Contract and refactor to keep up with the inverter-standard

----
The permission system change lead to adaptions in different modules especially the predefined roles. The following folders should be examined for their authorization logic and the introduced changes. They can be looked at in any order, as they should work independent from each other:

**PaymentProcessor**
- src/module/paymentProcessor

Contains the least amount of changed contracts.

**FundingManager**
- src/module/fundingManager

**LogicModules**
- src/module/logicModule


src/orchestrator/Orchestrator_v1.sol
```diff
@@ -306,45 +300,50 @@ contract Orchestrator_v1 is IOrchestrator_v1, ModuleManagerBase_v1 {
...
+        permissioned
+    {
         _enforceNonPrivilegedModuleInterfaceCheck(module_);
-        _cancelModuleUpdate(module_);
+        _initiateAddModuleWithTimelock(module_);
     }
 
     /// @inheritdoc IOrchestrator_v1
-    function initiateAddModuleWithTimelock(address module_) external {
+    function executeAddModule(address module_) external permissioned {
```
jellegerbrandy, line 318: note that this function, jus as initiateAddModelWithTimelock and many others, was not permssioned in the previous version

jellegerbrandy: it is because the permissions are removed from _executeAddModule, and moved to the external functions, which seems sane

FHieser: Yes that has been bugging me for a while now :sweat_smile: 
Should still be fine no?

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -23,305 +23,612 @@ import {AccessControlEnumerableUpgradeable} from
...
+
+    /// @notice The counter for role IDs.
+    /// @dev	This is used to generate unique role IDs for each role.
+    /// @dev    Starts at 1, which symbolizes two roles: PUBLIC_ROLE and
+    ///         DEFAULT_ADMIN_ROLE, but is immediately incremented when a role
+    ///         is created.
+    uint internal _lastAssignedRoleId;
+
+    /// @dev	Storage gap for future upgrades.
+    uint[50] private __gap;
```
omega-audits, line 305: You added a few new state variables but left the gap the same. This is fine only if the contract is not already deployed or if you only use it for new deployments. But if the goal is to upgrade existing system this is not correct use of the gap and you will have wrong storage calls for contracts inheriting this one.

omega-audits: AUT_TokenGated_Roles_v1 is inheriting it, which means if you upgrade it like this its storage layout will be wrong

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -23,305 +23,612 @@ import {AccessControlEnumerableUpgradeable} from
...
-            role == DEFAULT_ADMIN_ROLE
-                && getRoleMemberCount(DEFAULT_ADMIN_ROLE) <= 1
-        ) {
-            revert Module__Authorizer__AdminRoleCannotBeEmpty();
+    /// @notice Verifies that the roleId is already existing.
+    /// @param  roleId_ The id of the role.
+    modifier idExists(bytes32 roleId_) {
+        // If the given roleId is greater than the last assigned roleId, then
+        // it is not existing.
+        if (uint(roleId_) > _lastAssignedRoleId) {
```
omega-audits, line 273: This misses the BURN_ADMIN_ROLE which also exists. Which means in certain functions it won't be usable.

omega-audits: Seems like it's on purpose to separate burning from other functionality, but if you add it here you could remove the special function for burning and have it as an option in the regular flow.

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -23,305 +23,612 @@ import {AccessControlEnumerableUpgradeable} from
...
-        _revokeRole(roleId, target);
+    function removeAccessPermission(
+        address target_,
+        bytes4 selector_,
+        bytes32 roleId_
+    ) public permissioned {
+        bytes32[] memory permissions = _permissions[target_][selector_];
+        uint permissionsLength = permissions.length;
+
+        for (uint i = 0; i < permissionsLength; i++) {
```
omega-audits, line 522: Instead of loop here, it would be more efficient to move the search logic to the off chain code, so you add an index parameter to the function and just verify it's correct or revert.

src/modules/authorizer/role/AUT_Roles_v1.sol
```diff
@@ -23,305 +23,612 @@ import {AccessControlEnumerableUpgradeable} from
...
-        onlyRole(DEFAULT_ADMIN_ROLE)
+        onlyRole(getRoleAdmin(roleId_))
+        idExists(roleId_)
     {
-        bytes32 roleId = generateRoleId(address(orchestrator()), role);
-        _grantRole(roleId, target);
+        // If Role Admin is burned do nothing.
+        if (getRoleAdmin(roleId_) == BURN_ADMIN_ROLE) {
+            return;
+        }
```
omega-audits: This can never happen because onlyRole(getRoleAdmin(roleId_)) can never pass if the admin is BURN_ADMIN_ROLE (since BURN_ADMIN_ROLE is assigned to mamage itself in the init)

src/modules/logicModule/LM_PC_KPIRewarder_v2.sol
```diff
@@ -218,9 +218,9 @@ contract LM_PC_KPIRewarder_v2 is
 
     /// @inheritdoc ILM_PC_KPIRewarder_v2
     /// @dev    Top up funds to pay the optimistic oracle fee
+
     function depositFeeFunds(uint amount)
         external
-        onlyOrchestratorAdmin
```
omega-audits, line 223: No permissioned added here to replace onlyOrchestratorAdmin. Is that on purpose?

src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1.sol
```diff
@@ -118,21 +111,13 @@ contract FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1 is
     uint internal _lastSeizeTimestamp;
     /// @notice Address of the reserve pool.
     address internal _tokenVault;
-    /// @notice Restricts buying and selling functionalities to specific role.
-    bool internal _buyAndSellIsRestricted;
```
omega-audits, line 122: Since the contract is meant to be upgradeable, if it's already deployed you should not change the storage layout like this by deleting a variable, as it can mess the storage layout later or for contracts inheriting from it.

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
```diff
@@ -156,184 +156,233 @@ contract AUT_EXT_VotingRoles_v1 is IAUT_EXT_VotingRoles_v1, Module_v1 {
...
     //--------------------------------------------------------------------------
-    // Voter Management Functions
+    // Mutating - Voter Management Functions
 
     /// @inheritdoc IAUT_EXT_VotingRoles_v1
-    function addVoter(address who) public onlySelf isValidVoterAddress(who) {
-        if (!isVoter[who]) {
-            isVoter[who] = true;
+    function addVoter(address who_) public onlySelf isValidVoterAddress(who_) {
+        if (!_isVoter[who_]) {
```
jellegerbrandy, line 293: This is not validating the threshold. (.e.g.  threshold 1 and 2 voters ("1/2"), and you call addVoter(..) here, you would get an situation of 1/3, which does not pass the validateThreshold check)

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
```diff
@@ -156,184 +156,233 @@ contract AUT_EXT_VotingRoles_v1 is IAUT_EXT_VotingRoles_v1, Module_v1 {
...
-        motion_.target = target;
-        motion_.action = action;
+        motion_.target = target_;
+        motion_.action = action_;
 
         motion_.startTimestamp = block.timestamp;
-        motion_.endTimestamp = block.timestamp + voteDuration;
-        motion_.requiredThreshold = threshold;
+        motion_.endTimestamp = block.timestamp + _voteDuration;
+        motion_.requiredThreshold = _threshold;
```
jellegerbrandy, line 354: the current logic does not guarantee that the threshold conditions are met (i.e. that threshold <= voter, and that it should be a 1/1, 1/2 or t/x with t>= 2) . The reason is that the threshold is stored with the motion, but hte legit voters are stored on the conract level. So we could have a situation (for example) with a motion crated with a threshold of 1, and we then add some voters, and get a 1/4 situation, whcih would not be valid)

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
```diff
@@ -13,68 +13,68 @@ import {ERC165Upgradeable, Module_v1} from "src/modules/base/Module_v1.sol";
...
     }
 
-    //--------------------------------------------------------------------------
+    //==========================================================================
     // Modifiers
 
-    /// @dev	Reverts if caller is not the module itself.
+    /// @notice	Reverts if caller is not the module itself.
     modifier onlySelf() {
         if (_msgSender() != address(this)) {
```
jellegerbrandy, line 51: note to self: how does this work?

src/modules/authorizer/role/AUT_Roles_v1.sol
jellegerbrandy: So as far as I can see there is nothing that breaks, right?

src/modules/authorizer/role/AUT_Roles_v1.sol
jellegerbrandy: yeah but it would typically be a very short list

src/modules/logicModule/LM_PC_KPIRewarder_v2.sol
jellegerbrandy: there is no security risk afaics

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: Ah yes I could have explained this better.
The plan is to do a major version upgrade for all of the contracts after your review is done. Meaning all of the contracts that contain changes of authorization will be replaced by newer versions of them (incrementing the version in the title and replacing the old ones).
As the contracts are then considered "new" ones and are not backwards compatible, a reset of the gap should not be harmful anymore.
I pushed the update of the mayorversion to after the Audit, as a change in the name of the file in combination with a change in its contents makes git see it as a new file which would be hassle for readability of the reviews.

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: I dont see how to implement this with a offchain setup in a way to dynamically handle this onchain, without to much hassle.

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: The BURN_ADMIN_ROLE is just to restrict the Admin position
As it is the max byte value it is always bigger then _lastAssignedRoleId which would in a revert as a invalid Id which is as intended. That way it doesnt have to be included here in any way

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: yes you are correct here
will remove it

src/modules/logicModule/LM_PC_KPIRewarder_v2.sol
FHieser: this was a deliberate decision from our side

src/modules/fundingManager/bondingCurve/FM_BC_BondingSurface_Redeeming_Restricted_Repayer_Seizable_v1.sol
FHieser: See comment above about the majorversion and file naming

src/modules/authorizer/role/AUT_Roles_v1.sol
omega-audits: well, just to be sure we understand each other. So looping over list of arbirary length can be a problem in solidity because of (a) costs and (b) gas limits (if the list is VERY long you may not be able to remove permissions towards the end of the list because you run out o gas) .  Ben's suggestion would be very easy to implement - you'd add an `index` parameter to the `removeAccessPermission` function (or overload it) and instead of looping over the permissions to ind the roleId you want to remove, you just remove the one at the index that is given.. That said, I think for this case you are fine without it

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
FHieser: The point is that all the functions that are tagged with this modifier are only callable by motions that are created in the contract itself

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
FHieser: But the moment the call was made it was valid, no?
Just for my understanding: You are saying that the moment the voter is added the threshold is no longer working correctly?

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
FHieser: Yeah you are correct here.
I will remove the motion threshold use the contract threshold instead :+1:

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
FHieser: Will adapt it. See comment below

src/modules/authorizer/role/AUT_Roles_v1.sol
FHieser: I cant read apparently :see_no_evil: 
Yeah you are correct, it would be possible to do it with an explicit index, but I agree that the list of roleIds would not be very long in this case

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
jellegerbrandy: This was not fixed yet in 435d9ee70f92103fd112786a78f94822d0199329

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
jellegerbrandy: I believe this fix introduces a new problem. Consider this scenario:

1. Start with a config with  6 voters (a, b, c .. f) and a threshold of 4.
2. Now a,  b and c vote YAY - and the threshold is not reached.
3. You have reasons to suspect aht a and b are not so  trustworthy, so these two accounts are removed and the threshold is changed to 3 (i.e. you need 3 out of 4 votes from voters c, d, e and f to pass the motion).
However, now anyone can call executeMotion succesfully, even if neither the condition from before nor the conditions from after the update were ever met

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
FHieser: Yes you are correct here. I thought this would be corrected with the adaption of the threshold I did there.
Will add it accordingly

src/modules/authorizer/extensions/AUT_EXT_VotingRoles_v1.sol
FHieser: okay after much discussion internally we agreed that the initial implementation is the lesser evil of all of the options available to properly fix this and if properly communicated should not cause any mayor problems.
Therefor I will revert the changes I made and adapt the natspec of the add and Remove voter functions

Issue #744, opened 2025-05-11
Denial of Service via Unbounded `_queue` Growth in `executePaymentQueue()`


MehdiKarimi81:
### Summary

The `executePaymentQueue()` function in `PP_Queue_ManualExecution_v1` is vulnerable to a Denial-of-Service (DoS) attack due to unbounded growth of the `_queue` linked list.

A malicious actor can repeatedly submit redeem requests with small or dust amounts, resulting in numerous entries being added to `_queue`. Because `executePaymentQueue()` attempts to iterate over **every item** in `_queue` and execute the order, the function’s gas consumption grows linearly with the number of items in the `_queue`.

Once the _queue grows large enough, the function will consistently **run out of gas**, making it inaccessible thereby orders can't be executed.
### Recommendation 

1. **Bound the iteration** in `executePaymentQueue()`:

   * Process up to a maximum number of orders per call.

2. **Rate-limit or filter spam requests**:

   * Reject extremely small or dust-value orders.
   * Implement minimum thresholds per token or per user

Zitzak:
Thank you for your finding @MehdiKarimi81. As it currently stands, the creation of payment orders, which are converted into queue elements, is currently permissioned through whitelisting in the `FM_PC_Oracle_Redeeming_v1` contract. Because of that reason I don't see how someone can DoS attack the contract by adding payment orders to the queue.

If we were to bound the execution of the queue to a certain number of queue elements, what would be your suggestion ser?

MehdiKarimi81:
> Thank you for your finding [@MehdiKarimi81](https://github.com/MehdiKarimi81). As it currently stands, the creation of payment orders, which are converted into queue elements, is currently permissioned through whitelisting in the `FM_PC_Oracle_Redeeming_v1` contract. Because of that reason I don't see how someone can DoS attack the contract by adding payment orders to the queue.
> 
> If we were to bound the execution of the queue to a certain number of queue elements, what would be your suggestion ser?

1 - It can also happen through normal user interactions, when number of orders in the queue is large enough ( users create redeem requests normally ) this can happen without an attack. 
2 - In my opinion whitelisted users should not act maliciously since they're whitelisted to perform specified interactions, however it heavily depends on your trust level to whitelisted users.
 
You can add a parameter to executePaymentQueue to specify the number of orders to execute in the underlying call by the caller module.

leeftk:
[...]

PR #755, opened 2025-05-26
fix(PP_Queue_v1): Executing queue can run out of gas

Zitzak:
## Why
If enough orders are in the redemption queue, the execution of the function to process the queue could run out of gas. This has been fixed by introducing a maxOrderPerExecution value which will stop the queue execution loop when it is reached.

## What has been done?
- Add setters and getters for maxOrderPerExecution value
- Add value to while loop in _executePaymentQueue
- Update tests

## Additional Context
- https://inverternetwork.slack.com/archives/C07RXJVLAUS/p1747214034621999
- Closes issue #744


0xNuggan (approved):
Looks good!

PR #758, opened 2025-06-04
FIX: Enable protocol fee collection during buy operations in BondingCurveBase

Zitzak:
## What has been done?
The BondingCurveBase defined the interface _handleCollateralTokensBeforeBuy to be overriden by the implementation. This created the effect that the implementation could transfer all the incoming collateral without the contract having the chance to deduct the protocol fee.

A change is made whereby the collateral first gets transferred into the contract after which the implementation can decide where to transfer the incoming collateral excluding the protocol fee.

All implementations and tests have been adapted

**Please reference the Linear issue for more context**


0xNuggan (approved):
Looks good!

PR #759, opened 2025-06-09
Fix: Add project fee to payment order to calculate protocol fee in PP_Queue

Zitzak:
## Why has this been done?
- The protocol collateral fee for selling is calculated in the Payment Client within a calculation using also the project fee percentage. However it is not collected in the Payment Client due to the collateral not being in the contract at time of calculation
- This created a situation whereby when re-calculating it in the Payment Processor, the total amount from which the fee is calculated differs from the first time the calculation is done, resulting in a different net and protocol fee amount
- To accommodated for this, the project fee is added to the payment order so the correct protocol fee amount can be correctly be re-calculated

## What has been done?
- Project Fee parameter has been added to the payment order.
- The functions to `get` and validate the data from the payment orders have been refactored and adapted for the new field in the payment order
- A E2E test has been added that tests the whole flow with collateral project fee enabled for buy and sell & protocol fee enabled for issuance/collateral for both buy and sell

### DIFF
--- a/src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
+++ b/src/modules/fundingManager/oracle/FM_PC_Oracle_Redeeming_v1.sol
@@ -231,6 +231,9 @@ contract FM_PC_Oracle_Redeeming_v1 is
     /// @notice Flag used for the payment order.
     uint internal constant FLAG_ORDER_ID = 0;
 
+    /// @notice Flag used for the payment order.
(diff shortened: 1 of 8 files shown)


0xNuggan (approved):
Looks good to me. The only thing I see is there will be an overlap in the Flag with the everclear maxFee. Since both are fee related, I don't think it's bad conceptually, but maybe it can make sense to "push" the everclar IDs down to keep it fully separate when the time comes

PR #760, opened 2025-06-10
[Feature] Authorizer_v2

FHieser:
This is it:
The final PR to get the Authorizer fully rolling.
The PR itself is the followup PR to the Authorizer Audit that adjusts the Versioning of all Contracts that were affected by the Authorizer update itself and some minor Documentation changes.

This PR is enormous. In size as well as depth. Thats why I want as many eyes on this as possible:
@0xNuggan and @marvinkruse are the main reviewers on all the version changes that are happening.
There are a lot of them and I personally think that the implications of these changes is where the heart of this PR lies. 
Take a look at the changes that tackle the explicit Interface Checks.
Looking up -> ).supports <- and -> ERC165Checker <- references should work best on this
@0xNuggan I had some talks with @marvinkruse about the implications on the Factory/Governor/existing Workflow side already, but I think it is healthy for you to also look into this, so ask as many questions as you can, so we can build up some understandings on the topic itself. @fabianschu if you have the time this would be something where I would appreciate your input.

@fabianschu and @Zitzak it would be very cool if you could take a look at how the new Authorization system will work.
The RoleAuthroizer E2E tests should be the easiest to get a grip on this in my opinion, including the RoleAuthorizer itself of course.


src/external/governance/Governor_v1.sol
FHieser, line 1: @marvinkruse Do we need to update the governor with a function to register V1 Contracts with this?

src/factories/ModuleFactory_v1.sol
FHieser, line 1: Same here?

src/factories/OrchestratorFactory_v1.sol
FHieser, line 1: Here as well

src/modules/lib/LibMetadata.sol
FHieser, line 1: This is also an outlier

src/modules/authorizer/role/AUT_Roles_v2.sol
```diff
@@ -0,0 +1,449 @@
...
+    function hasPermission(address caller_, address target_, bytes4 selector_)
+        external
+        view
+        virtual
+        returns (bool hasPermission_)
+    {
+        // If caller is the admin, they can call any function.
+        if (hasRole(DEFAULT_ADMIN_ROLE, caller_)) {
+            return true;
+        }
```
Zitzak, line 234: Is there no way to restrict the default admin to calling a function? I could see usecases whereby a function should (from a legal perspective) be restricted to only a specific address which should for example could be held by an external party for regulation needs

src/modules/authorizer/role/AUT_Roles_v2.sol
```diff
@@ -0,0 +1,449 @@
...
+    function labelRole(bytes32 roleId_, string memory newRoleName_)
+        external
+        permissioned
+        idExists(roleId_)
+    {
+        emit RoleLabeled(roleId_, newRoleName_);
+    }
+
+    /// @inheritdoc IAuthorizer_v2
+    function transferAdminRole(bytes32 roleId_, bytes32 newAdminRoleId_)
```
Zitzak, line 295: This function, and multiple below, are not virtual

Zitzak (commented):
Did a review into the main role AUT and module, and adjacent contracts I've touched in the past so I know the change. When I started the review we didn't know the strategic change was coming.

src/modules/authorizer/role/AUT_Roles_v2.sol
FHieser: You are correct that there is no obvious/default way to do this. There is a workaround though:
You can give a role the rights to change the Function Permissions and then renounce the default admin.
THe new role then cannot call all functions automatically, but has the rights to adapt that (kinda simulating the rights of the default admin)

docs/src/modules/authorizer/role/AUT_Roles_v2.md
```diff
@@ -0,0 +1,263 @@
...
+
+#### Role Creation
+
+A role can be created by calling the createRole function. The function takes the following parameters:
+
+- The role name
+- The role id of the role that will become the admin of the new role.
+- The addresses of the initial members of the new role.
+
+The rolename in this context refers to the label of the role and is therefor not referenceable onchain. The function can only be called by a permissioned address (See [permissioned](#permissioned-modifier)).
```
0xNuggan: therefore*

script/deploymentSuite/MetadataCollection_v1.s.sol
```diff
@@ -60,88 +60,77 @@ contract MetadataCollection_v1 {
...
-        "Orchestrator_v1"
+        "Orchestrator_v2"
     );
 
     // ------------------------------------------------------------------------
     // Authorizer
 
     // RoleAuthorizer
-    IModule_v1.Metadata public roleAuthorizerMetadata = IModule_v1.Metadata(
-        1, 0, 0, "https://github.com/InverterNetwork/contracts", "AUT_Roles_v1"
```
0xNuggan, line 76: Should be 2

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
```diff
@@ -2,14 +2,14 @@
...
-import {IOrchestrator_v1} from
-    "src/orchestrator/interfaces/IOrchestrator_v1.sol";
+import {IOrchestrator_v2} from
+    "src/orchestrator/interfaces/IOrchestrator_v2.sol";
 import {IFundingManager_v1} from "@fm/IFundingManager_v1.sol";
 import {IFM_DepositVault_v1} from
     "@fm/depositVault/interfaces/IFM_DepositVault_v1.sol";
 
 // Internal Dependencies
-import {Module_v1, IModule_v1} from "src/modules/base/Module_v1.sol";
```
0xNuggan, line 12: Curious about your choice to not make deposit permissioned here. Not saying I disagree, I think it makes sense to have a "super basic" FM, but was there any particular reason apart from ressource allocation?

script/deploymentSuite/SingletonDeployer_v1.s.sol
```diff
@@ -213,81 +205,81 @@ contract SingletonDeployer_v1 is ProtocolConstants_v1 {
...
+        impl_mod_FM_EXT_TokenVault_v2 = deployAndLogWithCreate2(
+            "FM_EXT_TokenVault_v2",
+            vm.getCode("FM_EXT_TokenVault_v2.sol:FM_EXT_TokenVault_v2")
         );
 
         // Logic Modules
         console2.log("  -- Logic Modules");
 
         impl_mod_LM_PC_Oracle_Permissioned_v1 = deployAndLogWithCreate2(
-            "LM_Oracle_Permissioned_v1",
```
0xNuggan, line 235: var name should be renamed :|

src/external/governance/Governor_v1.sol
0xNuggan: Given latest developments I would say no, unless the effort is minimal. WDYT @marvinkruse ?

src/modules/base/IModule_v1.sol
```diff
@@ -104,9 +104,9 @@ interface IModule_v1 {
     /// @return The module's title.
     function title() external view returns (string memory);
 
-    /// @notice Returns the module's {Orchestrator_v1} interface, {IOrchestrator_v1}.
+    /// @notice Returns the module's {Orchestrator_v2} interface, {IOrchestrator_v2}.
     /// @return The module's {Orchestrator_1}.
-    function orchestrator() external view returns (IOrchestrator_v1);
+    function orchestrator() external view returns (IOrchestrator_v2); //@todo what to do with this? Should this be version 1?
```
0xNuggan, line 109: Why are we keeping the V1s at all?

src/modules/logicModule/interfaces/IERC20PaymentClientBase_v3.sol
```diff
@@ -0,0 +1,185 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.0;
+
+// Internal Interfaces
+import {IPaymentProcessor_v3} from
```
0xNuggan: Import can be removed, also in v2 and v1

README.md
```diff
@@ -1,7 +1,8 @@
 <a href="https://inverter.network" target="_blank"><img align="right" width="150" height="150" top="100" src="./assets/logo_circle.svg"></a>
 
 # Inverter Network Smart Contracts
-*Inverter is the pioneering web3 protocol for token economies, enabling conditional token issuance, dynamic utility management, and token distribution. Build, customize, and innovate with Inverter's modular logic and extensive web3 interoperability.*
+
```
marvinkruse, line 4: Why is this change being done? Or rather: was this an accident or on purpose?

marvinkruse: Looks like it may just be your formatter or something? Same as later in the same file?

docs/src/modules/authorizer/role/AUT_Roles_v2.md
```diff
@@ -0,0 +1,263 @@
...
+
+- A unique identifier (ID)
+- A label
+- A list of members
+- A associated admin role
+
+The id is a value assigned by the authorizer module and is used to reference the role in the different functions of the authorizer module.
+
+The label is a string that is emitted as an event when the role is created. It is used to make the role human readable in the frontend and has no practical use in the
+onchain live setup.
```
marvinkruse:
```suggestion
The label is a string that is emitted as an event when the role is created. It is used to make the role human-readable in the frontend and has no practical use on-chain.
```

docs/src/modules/authorizer/role/AUT_Roles_v2.md
```diff
@@ -0,0 +1,263 @@
...
+A role can be granted to a address by calling the grantRole function. The function takes the following parameters:
+
+- The role id of the role to grant
+- The address to grant the role to
+
+The grantRole function can only be called by according admin of the role.
+
+#### Role revoking
+
+A role can be revoked from an address by calling therevokeRole function. The function takes the following parameters:
```
marvinkruse:
```suggestion
A role can be revoked from an address by calling the revokeRole function. The function takes the following parameters:
```

docs/src/modules/authorizer/role/AUT_Roles_v2.md
```diff
@@ -0,0 +1,263 @@
...
+
+- The role id of the role to revoke
+- The address to revoke the role from
+
+The revokeRole function can only be called by according admin of the role.
```
marvinkruse: There is a general inconsistency in formatting function names as normal text vs `codeText`. I think it makes more sense to be strict and just have it all as `codeText`, but at least it should be consistent throughout.

src/external/governance/Governor_v1.sol
marvinkruse: The metadata didn't change, so it makes no difference alltogether I'd say. I think the Metadata shouldn't have been part of the Module in the first place probably, but we can leave it like it as and just use v2. Unless I'm missing something?

src/factories/ModuleFactory_v1.sol
marvinkruse: It should be fine like it is here. We need to adapt a few things anyways I believe.

src/modules/authorizer/IAuthorizer_v2.sol
```diff
@@ -0,0 +1,268 @@
...
+ *              - Role creation and management. This includes the ability to
+ *                create roles, revoke roles, assigning and revoking role
+ *                admins, which can add and remove role members.
+ *
+ *              - Role-based access control. This includes the ability to grant
+ *                roles access to functions that implement the permissioned
+ *                modifier. Functions can also be set to public access by
+ *                adding the public role to the function permissions.
+ *
+ * @custom:documentation See https://github.com/InverterNetwork/contracts/tree/dev/docs/src/modules/authorizer/role/AUT_Roles_v2.sol
```
marvinkruse: Its not a sol file

src/modules/authorizer/role/AUT_Roles_v2.sol
```diff
@@ -0,0 +1,449 @@
...
+ *          Key features:
+ *              - Role creation and management. This includes the ability to
+ *                create roles, revoke roles, assigning and revoking role
+ *                admins, which can add and remove role members.
+ *              - Role-based access control. This includes the ability to grant
+ *                roles access to functions that implement the permissioned
+ *                modifier. Functions can also be set to public access by
+ *                adding the public role to the function permissions. *
+ *
+ * @custom:documentation See https://github.com/InverterNetwork/contracts/tree/dev/docs/src/modules/authorizer/role/AUT_Roles_v2.sol
```
marvinkruse: Same thing as above, this file doesnt exist as the doc file is not a sol file

src/modules/authorizer/role/AUT_Roles_v2.sol
```diff
@@ -0,0 +1,449 @@
...
+            return true;
+        }
+
+        bytes32[] memory roleIds = _permissions[target_][selector_];
+        uint permissionLength = roleIds.length;
+
+        // If there are no roles, the caller cannot call the function.
+        if (permissionLength == 0) {
+            return false;
+        }
```
marvinkruse: What does this do in the end? The code behaves the same way if this didn't exist, as the for loop below just doesn't do anything and the return false is reached below.

src/modules/authorizer/role/AUT_TokenGated_Roles_v2.sol
```diff
@@ -0,0 +1,367 @@
...
+            revert Module__AUT_TokenGated_Roles__InvalidThreshold(threshold_);
+        }
+        _;
+    }
+
+    // ========================================================================
+    // Storage
+
+    /// @dev	Stores if a role is token gated.
+    mapping(bytes32 => bool) internal _isTokenGated;
```
marvinkruse: I think we wanted to have named parameters here for mapping as well.

src/modules/authorizer/role/AUT_TokenGated_Roles_v2.sol
```diff
@@ -0,0 +1,367 @@
...
+
+        return super._grantRole(roleId_, who_);
+    }
+
+    /// @notice Revokes a role from an address.
+    /// @dev	Overrides {_revokeRole} to clean up threshold data on revoking.
+    /// @param  roleId_ The id number of the role.
+    /// @param  who_ The user we want to check on.
+    /// @return success_ Returns if revoke has been succesful.
+
```
marvinkruse:
```suggestion
```

src/modules/authorizer/role/AUT_TokenGated_Roles_v2.sol
```diff
@@ -0,0 +1,367 @@
...
+    /// @notice Sets the minimum threshold for a token-gated role.
+    /// @dev	This function does not validate the threshold. It is
+    ///         technically possible to set a threshold above the total supply
+    ///         of the token.
+    /// @param  roleId_  The ID of the role to be modified.
+    /// @param  token_ The token for which to the threshold.
+    /// @param  threshold_ The user will need to have at least this number to
+    ///         qualify for the role.
+
```
marvinkruse:
```suggestion
```

src/modules/base/IModule_v1.sol
marvinkruse: because workflows and modules still exist that reference them, use them, etc. - so I think we have a few cases where we can't just remove this.
we could look into this though with our new strategy potentially, to just cut old modules out kinda.

marvinkruse (changes requested):

src/modules/fundingManager/depositVault/FM_DepositVault_v1.sol
FHieser: Nope, it was just to enable everybody to do deposits

README.md
FHieser: Yeah seems to be formatter issue
Doesnt seem to change the form of the output file though
So I would keep it

docs/src/modules/authorizer/role/AUT_Roles_v2.md
FHieser: I tried my best :sweat_smile:

src/modules/authorizer/role/AUT_Roles_v2.sol
FHieser: I removed it

src/modules/authorizer/role/AUT_Roles_v2.sol
```diff
@@ -0,0 +1,460 @@
...
+            }
+        }
+        // Do nothing if the value is not found.
+    }
+
+    // ------------------------------------------------------------------------
+    // Mutating - Mixed Utility
+
+    /// @inheritdoc IAuthorizer_v2
+    function createRoleAndAddAccessPermissions(
```
0xNuggan, line 370: Just a suggestion: I could imagine this running into "out of gas" issues if the role setup becomes too complex. Maybe add a upper cap in the amount of targets/selectors? To force the developer to "split the setup" in advance

FHieser: In theory yes, this could happen, but splitting up the process might not be ideal too, as the process of determining the Id of the role win addition to then adding the access permissions separately would be a very manual process.
If gas is a general problem one could instead use the createRole and addAccessPermission functions independently anyway

FHieser: So I would keep it as is

0xNuggan: I had in mind sth more like "giving the user a clear error msg instead of the tx just failing". But I agree it's secondary. Let's just keep it.

0xNuggan (approved):

PR #768, opened 2025-07-07
[DONT MERGE] Create BC Forwarder contract and tests for QAcc

0xNuggan:
Intermediary forwarder contract to make the qAcc Restricted Bonding Curves public. This basically forwards the buy / sell calls without requiring whitelisting. That way, the original contract just needs to grant CURVE_USER_ROLE to the forwarder and from then on everybody can use the curve


0xNuggan:
Closing since qAcc ran their own solution

PR #775, opened 2025-08-18
Lending Facility & DFC

leeftk:
## Overview
Implemented a lending facility that allows users to borrow collateral tokens against locked issuance tokens with dynamic fee calculation.

## Key Changes
- Added lending facility logic module with borrow/repay functionality
- Integrated dynamic fee calculation based on utilization rates
- Added role-based access control for facility management
- Implemented borrowing limits and quotas

## Technical Details
- Users lock issuance tokens as collateral to borrow collateral tokens
- Dynamic fees scale with system utilization
- Configurable individual and system-wide borrowing limits
- Automatic issuance token locking/unlocking during borrow/repay
- Fee collection and distribution to DBC FM

### Formulas & Pricing
- **Borrow Capacity**: `virtualIssuanceSupply × P_floor` (where P_floor is first segment's initial price)
- **Required Collateral**: `borrowAmount ÷ floorPrice` (issuance tokens needed)
- **Utilization Rate**: `currentlyBorrowedAmount ÷ borrowCapacity`
- **Dynamic Fee**: `requestedAmount × feeRate` (feeRate from DynamicFeeCalculator based on utilization)
- **Floor Liquidity Rate**: `(borrowableAmount - currentlyBorrowedAmount) ÷ borrowableAmount`

## Testing
- Borrow/repay flows verified with proper state management
- Fee calculation accuracy confirmed across utilization ranges
- Access control and limit enforcement tested

## Missing
- Updates from tech specs and implementation that need to be made
- Looping feature for allowing multiple loops
- More fuzz tests
- Testing all unhappy paths
- Inverter Standard review
- Invariant tests


src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,538 @@
...
+        uint floorPrice = _getFloorPrice();
+        return borrowAmount_ * 1e18 / floorPrice; // Adjust for decimals
+    }
+
+    /// @dev Get the current floor price from the DBC FM
+    /// @return The current floor price
+    function _getFloorPrice() internal view returns (uint) {
+        IFM_BC_Discrete_Redeeming_VirtualSupply_v1 dbcFm =
+            IFM_BC_Discrete_Redeeming_VirtualSupply_v1(_dbcFmAddress);
+        return dbcFm.getStaticPriceForBuying();
```
FHieser: getStaticPriceForBuying is not returning you the floor price
You would have to fetch the price of the first segment of the Discrete bonding Curve here instead

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,538 @@
...
+    function _calculateUserBorrowingPower(address user_)
+        internal
+        view
+        returns (uint)
+    {
+        // Use the DBC FM to get the actual floor price
+        // User borrowing power = locked issuance tokens * floor price
+        IFM_BC_Discrete_Redeeming_VirtualSupply_v1 dbcFm =
+            IFM_BC_Discrete_Redeeming_VirtualSupply_v1(_dbcFmAddress);
+        uint floorPrice = dbcFm.getStaticPriceForBuying();
```
FHieser: Same here

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,538 @@
...
+        uint repaymentProportion =
+            (repaymentAmount_ * 1e27) / _outstandingLoans[user_];
+        // Calculate the proportion of locked issuance tokens to unlock
+        return (_lockedIssuanceTokens[user_] * repaymentProportion) / 1e27;
+    }
+
+    /// @dev Calculate the required collateral amount for a given issuance token amount
+    /// @param issuanceTokenAmount_ The amount of issuance tokens
+    /// @return The required collateral amount
+    function _calculateCollateralAmount(uint issuanceTokenAmount_)
```
FHieser, line 547: This is not used anywhere

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,538 @@
...
+
+        if (_outstandingLoans[user] < repaymentAmount_) {
+            repaymentAmount_ = _outstandingLoans[user];
+        }
+
+        // Transfer collateral back to DBC FM
+        _collateralToken.safeTransferFrom(user, _dbcFmAddress, repaymentAmount_);
+
+        // Calculate and unlock issuance tokens
+        uint issuanceTokensToUnlock =
```
FHieser, line 263: There are some implications if the floor price has risen inbetween the borrowing and the repayment
Would this still work properly in case that happens?

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,538 @@
...
+        // Update state
+        _outstandingLoans[user] -= repaymentAmount_;
+        currentlyBorrowedAmount -= repaymentAmount_;
+
+        // Emit event
+        emit Repaid(user, repaymentAmount_, issuanceTokensToUnlock);
+    }
+
+    /// @inheritdoc ILM_PC_Lending_Facility_v1
+    function unlockIssuanceTokens(uint amount_) external virtual {
```
FHieser: With the repay you already transfer tokens back to the user
What is the purpose of this function?

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,538 @@
...
+
+    /// @dev The role that allows managing the lending facility parameters
+    bytes32 public constant LENDING_FACILITY_MANAGER_ROLE =
+        "LENDING_FACILITY_MANAGER";
+
+    /// @notice Borrowable Quota as percentage of Borrow Capacity (in basis points)
+    uint public borrowableQuota;
+
+    /// @notice Individual borrow limit per user
+    uint public individualBorrowLimit;
```
FHieser: The individualBorrowLimit was probably an old specification
You can remove this

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: Fixed the issue to use `PackedSegmentLib._initialPrice(segments[0])` to fetch the floorPrice

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: Fixed the issue to use `PackedSegmentLib._initialPrice(segments[0])` to fetch the floorPrice

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: This is later being used in the looping feature.

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: Issue Fixed - removed all references to `individualBorrowLimit`

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: I have removed the `unlockIssuanceTokens` as it is no longer required

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: No, there is no effect of floorPrice on the `issuanceTokensToUnlock` calculation.

FHieser (approved):

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,803 @@
...
+        borrowableQuota = borrowableQuota_;
+        maxLeverage = maxLeverage_;
+        nextLoanId = 1; // Start loan IDs from 1
+    }
+
+    // =========================================================================
+    // Public - Mutating
+
+    /// @inheritdoc ILM_PC_Lending_Facility_v1
+    function borrow(uint requestedLoanAmount_)
```
FHieser, line 225: This should return the loan Id

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,803 @@
...
+
+        // Transfer collateral back to DBC FM
+        _collateralToken.safeTransferFrom(user, _dbcFmAddress, repaymentAmount_);
+
+        // Emit events
+        emit LoanRepaid(loanId_, user, repaymentAmount_, issuanceTokensToUnlock);
+    }
+
+    /// @inheritdoc ILM_PC_Lending_Facility_v1
+    function buyAndBorrow(uint leverage_) external virtual {
```
FHieser: Same here

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,803 @@
...
+        // Emit event for the completed buyAndBorrow operation
+        emit BuyAndBorrowCompleted(
+            user,
+            leverage_,
+            totalIssuanceTokensReceived,
+            totalBorrowed,
+            totalCollateralUsed
+        );
+    }
+
```
FHieser, line 315: I am still missing the rebalance function, that can be used if the floor price increased

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
FHieser, line 225: You could add a borrowFor and buy and borrowFor function

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,803 @@
...
+        address user = _msgSender();
+
+        // Calculate how much issuance tokens need to be locked for this borrow amount
+        uint requiredIssuanceTokens =
+            _calculateRequiredIssuanceTokens(requestedLoanAmount_);
+
+        // Check if borrowing would exceed borrowable quota
+        if (
+            currentlyBorrowedAmount + requestedLoanAmount_
+                > _calculateBorrowCapacity() * borrowableQuota / 10_000 // @note: Optimize this to an internal function later
```
FHieser: Still todo

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,803 @@
...
+        address user = _msgSender();
+
+        if (leverage_ < 1 || leverage_ > maxLeverage) {
+            revert
+                ILM_PC_Lending_Facility_v1
+                .Module__LM_PC_Lending_Facility_InvalidLeverage();
+        }
+
+        // Get user's total collateral balance at the start
+        uint userCollateralBalance = _collateralToken.balanceOf(user);
```
FHieser: Fetching the total amount of tokens is a bit much, please add a deposit amount

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,803 @@
...
+    /// @return The amount of issuance tokens to unlock
+    function _calculateIssuanceTokensToUnlockForLoan(
+        Loan memory loan_,
+        uint repaymentAmount_
+    ) internal pure returns (uint) {
+        if (loan_.remainingPrincipal == 0) return 0;
+
+        // Calculate the proportion of the loan being repaid
+        uint repaymentProportion =
+            (repaymentAmount_ * 1e27) / loan_.remainingPrincipal;
```
FHieser, line 494: where is 10^(27) coming from?

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,803 @@
...
+    function _calculateBorrowCapacity() internal view returns (uint) {
+        // Get the issuance token's total supply (this represents the virtual issuance supply)
+        uint virtualIssuanceSupply = IERC20(
+            IBondingCurveBase_v1(_dbcFmAddress).getIssuanceToken()
+        ).totalSupply();
+
+        uint pFloor = _getFloorPrice();
+
+        // Borrow Capacity = virtualIssuanceSupply * P_floor
+        return virtualIssuanceSupply * pFloor / 1e18; // Adjust for decimals
```
FHieser, line 511: What if the token is not using 1e18?
I would just remove this

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,803 @@
...
+    /// @return The user's borrowing power
+    function _calculateUserBorrowingPower(address user_)
+        internal
+        view
+        returns (uint)
+    {
+        // Use the DBC FM to get the actual floor price from the first segment
+        // User borrowing power = locked issuance tokens * floor price
+        uint floorPrice = _getFloorPrice();
+        return _lockedIssuanceTokens[user_] * floorPrice / 1e18; // Adjust for decimals
```
FHieser, line 525: same here

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
FHieser: Can we add another function that only does the loop without the buy?

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: Fixed it by returning the loanId

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: Fixed it by returning the loanId

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: Can be ignored

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: Fixed it

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: We use `1e27` instead of the standard `1e18` to avoid values rounding down to zero. I ran into this issue while running fuzz tests, and found that using `1e27` was a reliable solution. 

Do you have any input on this?

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: The result would be 1e18 times larger we dont divide by the 1e18. 
Do you want me to use IERC20(token).decimal instead?
Can you clarify on this? @FHieser

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: The result would be 1e18 times larger we dont divide by the 1e18. 
Do you want me to use IERC20(token).decimal instead?
Can you clarify on this?

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: I have pushed the changes to include borrowFor() and buyAndBorrowFor()

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: Why do we need this? Can you help me understand the utility for this!

leeftk: This will be handled by Felix in a future PR.

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
leeftk: I think this article does a good job of explaining WAD versus RAY especially when it comes to precision scaling in Defi. The basic idea here when using WAD bug due to rounding errors. Using RAY allows us to get enough decimals precision so that we can better handle small amounts.

https://medium.com/@olatunjimayowa0396/fixed-point-arithmetic-in-solidity-19ef23e9c4a6

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
FHieser: We decided to implement this feature at a later point of time due to scope creep for the external developers

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
FHieser: So it is a precision helper then?

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
FHieser: yes. In an ideal case this is decimal agnostic
Why would you abstract the decimals here in general?
Every other function also works with Decimal precision when returning values no?

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
FHieser: see comment above

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
```diff
@@ -0,0 +1,832 @@
...
+        }
+
+        // Return the initial price of the first segment (floor price)
+        return PackedSegmentLib._initialPrice(segments[0]);
+    }
+
+    /// @dev Internal function that handles all borrowing logic
+    function _borrow(
+        uint requestedLoanAmount_,
+        address tokenReceiver_,
```
FHieser, line 593: Couldnt the borrower and the token receiver be the same in this function?
I do not see a reason why they would be different inputs.

FHieser: The if `(tokenReceiver_ != address(this)) {` check could just be: 
`_msgSender!=address(this)` instead and you'd still have nearly the same behaviour.

FHieser: The reason that I am saying this is because the borrowFor function uses
`borrow(requestedLoanAmount_, receiver_, receiver_);`
I initially thought with this getup that the receiver also has to provide the initial issuance tokens,
but the borrow always uses _msgSender as the provider anyway

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: With respect to the abstraction of decimal, I meant in general but since all the token values are always scaled to 1e18. 
I think going ahead with the current changes should not be an issue.

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: With respect to the abstraction of decimal, I meant in general but since all the token values are always scaled to 1e18.
I think going ahead with the current changes should not be an issue.

src/modules/logicModule/LM_PC_Lending_Facility_v1.sol
zzzuhaibmohd: Yes, I ran into issues when trying to test edge cases like repay 1 wei amount of tokens etc., 
so that is when I changed the precision to be 1e27

PR #777, opened 2025-09-26
[Audit] Discrete Bondingcurve

FHieser:
This is the audit feature branch for the Discrete Bonding Curve Update.
The following contracts are the main focus points of the Audit.

- src/modules/fundingManager/bondingCurve/libraries/PackedSegmentLib.sol
- src/modules/fundingManager/bondingCurve/formulas/DiscreteCurveMathLib_v1.sol
- src/modules/fundingManager/bondingCurve/FM_BC_Discrete_Redeeming_VirtualSupply_v1.sol

- Direct periphery contracts like:
  - src/modules/fundingManager/bondingCurve/abstracts/BondingCurveBase_v1.sol
  - src/modules/fundingManager/bondingCurve/abstracts/RedeemingBondingCurveBase_v1.sol
  - src/modules/fundingManager/bondingCurve/types/PackedSegment_v1.sol
  - src/modules/fundingManager/bondingCurve/interfaces/IDiscreteCurveMathLib_v1.sol
  - src/modules/fundingManager/bondingCurve/interfaces/IFM_BC_Discrete_Redeeming_VirtualSupply_v1.sol

A special care should be had for the precision and robustness of the Discrete Bonding Curve logic itself.
A recommend read on this is the following blog post:
https://www.carbondefi.xyz/blog/theory-vs-reality-the-precision-and-accuracy-of-carbon-defi-s-fixed-point-arithmetic

To get a better feel for the implementation and usage of the contracts I recommend to look at the following files:
- script/workflowDeploymentAndSetupScripts/DeployFullSetupAndDiscreteBondingCurveWorkflow.s.sol
- test/e2e/fundingManager/FM_BC_Discrete_Redeeming_VirtualSupply_v1_E2E.t.sol

Folders and Files to ignore for the audit:
- context folder
- docs folder
- memory-bank folder


src/modules/fundingManager/bondingCurve/abstracts/RedeemingBondingCurveBase_v1.sol
```diff
@@ -235,7 +236,7 @@ abstract contract RedeemingBondingCurveBase_v1 is
         totalCollateralTokenMovedOut = collateralRedeemAmount;
 
         // Cache Collateral Token
-        IERC20 collateralToken = __Module_orchestrator.fundingManager().token();
+        IERC20 collateralToken = IFundingManager_v1(address(this)).token();
```
jellegerbrandy: why this change? Is this guaranteed to exist? Elsewher, for collateral, the original `__Module_orchestrator.fundingManager().token()` is used

jellegerbrandy: For example, in the _buyOrder in BondingCurveBase_v1.sol

src/modules/fundingManager/bondingCurve/FM_BC_Discrete_Redeeming_VirtualSupply_v1.sol
```diff
@@ -0,0 +1,546 @@
...
+            _depositAmount, issuanceToken.totalSupply()
+        );
+        return tokensToMint;
+    }
+
+    // ------------------------------------------------------------------------
+    // Internal - Overrides - RedeemingBondingCurveBase_v1
+
+    /// @inheritdoc RedeemingBondingCurveBase_v1
+    function _getSellFee() internal view virtual override returns (uint) {
```
jellegerbrandy: either override _getSellFee as you do here, or use the sellFee state variable (as in line 393) but do not mix the approaches

src/modules/fundingManager/bondingCurve/formulas/DiscreteCurveMathLib_v1.sol
```diff
@@ -0,0 +1,696 @@
...
+            revert
+                IDiscreteCurveMathLib_v1
+                .DiscreteCurveMathLib__ZeroCollateralInput();
+        }
+
+        // Phase 1: Find which segment and step to start purchasing from.
+        uint segmentIndex_ = 0;
+        uint supplyCoveredByPreviousSegments_ = 0;
+
+        if (currentTotalIssuanceSupply_ > 0) {
```
jellegerbrandy, line 257: these are a lot of potentially very long loops, and running out of gas may be a real problem here if you use a lot of segments.
also, many of the values here are are static (i.e. the segments capacity and total issuance) and could plausibly be cached or caluclated only once on init. In that way, you can do a tree search instead of each time looping over all segments and subsegments.

src/modules/fundingManager/bondingCurve/FM_BC_Discrete_Redeeming_VirtualSupply_v1.sol
FHieser: Removed the override

src/modules/fundingManager/bondingCurve/abstracts/RedeemingBondingCurveBase_v1.sol
FHieser: Reverted the change

src/modules/fundingManager/bondingCurve/formulas/DiscreteCurveMathLib_v1.sol
FHieser: yes we are aware of that. For our initial curves we only plan to host 2 to max 5 segments.
Yeah there is probably room for improvement from a efficiency perspective, but we can jump on that ship in later iterations. They main concern for now was that there might be bugs or errors in the code, as it is quite complex.

omega-audits (approved):

